PR #1113, opened 2019-09-04
Fix optimizer_wrapper_test

### LINKED ISSUE #1112: optimizer_wrapper_test fail
When running following command:

```bash
pytest elasticdl/python/tests/optimizer_wrapper_test.py -k test_set_slot_values_to_variables
```

The test reports an error:

```
self = <tf.Tensor: id=51, shape=(3, 4), dtype=bool, numpy=
array([[ True,  True,  True,  True],
       [ True,  True,  True,  True],
       [ True,  True,  True,  True]])>

    def __bool__(self):
>     return bool(self._numpy())
E     ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

Here compares two ResourceVariable

```python
self.assertTrue(
                    slots_dict[slot]
                    == opt_wrapper._slot_variables[layer][slot]
                )
```

The results of `slots_dict[slot] == opt_wrapper._slot_variables[layer][slot]` is 
```
tf.Tensor(
[[ True  True  True  True]
 [ True  True  True  True]
 [ True  True  True  True]], shape=(3, 4), dtype=bool)
<class 'tensorflow.python.framework.ops.EagerTensor'>
```

The tf.tensor could not be asserted true directly.

QiJune:
Fix #1112

### DIFF
--- a/elasticdl/python/tests/optimizer_wrapper_test.py
+++ b/elasticdl/python/tests/optimizer_wrapper_test.py
@@ -216,7 +216,10 @@ class OptimizerWrapperTest(unittest.TestCase):
                     ).all()
                 )
                 self.assertTrue(
-                    (slots_dict[slot].numpy() == opt_wrapper._slot_variables[layer][slot].numpy()).all()
+                    (
+                        slots_dict[slot].numpy()
+                        == opt_wrapper._slot_variables[layer][slot].numpy()
+                    ).all()
                 )
 
                 slots_dict[slot].assign(tf.ones((10, 4)))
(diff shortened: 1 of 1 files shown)


mhaoli (dismissed):
LGTM. It seems that in python 3.6.8 we can compare variable directly but we can't do that in python 3.7.x. It's better to transform to `numpy.ndarray` before comparison.

elasticdl/python/tests/optimizer_wrapper_test.py
```diff
@@ -271,7 +273,10 @@ def test_set_embedding_values_to_variables(self):
                 ).all()
             )
             self.assertTrue(
-                grads_and_vars[i][1] == opt_wrapper._embed_variables[layer]
+                (
+                    grads_and_vars[i][1].numpy()
+                    == opt_wrapper._embed_variables[layer].numpy()
+                ).any()
```
mhaoli: any -> all?

terrytangyuan (approved):

PR #1118, opened 2019-09-04
Move TensorBoard starting logic out of WorkerManager

terrytangyuan:
Previously `WorkerManager` starts TensorBoard service via `worker_manager.start_tensorboard_service()`. However, TensorBoard is only related to master pod and has nothing to do with other workers. This PR removes the dependency so the logic is clearer in `master.main`.

### DIFF
--- a/elasticdl/python/common/k8s_tensorboard_client.py
+++ b/elasticdl/python/common/k8s_tensorboard_client.py
@@ -7,19 +7,20 @@ from elasticdl.python.common.log_util import default_logger as logger
 
 
 class TensorBoardClient(object):
-    def __init__(self, k8s_client):
+    def __init__(self, **kwargs):
         """
         ElasticDL k8s TensorBoard client.
 
         Args:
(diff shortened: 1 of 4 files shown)


elasticdl/python/common/k8s_tensorboard_client.py
```diff
@@ -75,7 +85,7 @@ def _get_tensorboard_service(self):
             )
             return None
 
-    def get_tensorboard_url(self, check_interval=5, wait_timeout=120):
+    def _get_tensorboard_url(self, check_interval=5, wait_timeout=120):
```
ywskycn, line 88: Have we tested internally how much time it takes to get the address? I'm little worried here that the master pod will be blocked here for 120 secs, before it can start the training.

terrytangyuan: Good catch. Fixed.

ywskycn (approved):

PR #1120, opened 2019-09-04
Correct Travis CI badge

wangkuiyi:


wangkuiyi:
#1114 already fixed this problem.

PR #1125, opened 2019-09-05
Async SGD design Doc 

skydoorkai:


elasticdl/doc/async_sgd_design.md
```diff
@@ -0,0 +1,115 @@
...
+
+1. In `GetModel`, during the variable assign loop, there may be `ReportGradient` GRPC service running and updating the variables. Thus, variables in `local_model` in workers may contain values from model versions. `model_version` from `get_model_from_ps` is just a proximate model version.
+2. There may be multiple `ReportGradient` running concurrently. Different model variables may apply these gradients in different orders. 
+
+Also, the concurrent updates to variables in `ReportGradient` may cause some gradients are not applied, as the updates can overwritten by other concurrent running updates. TensorFlow optimizers have an argument [`use_locking`](https://github.com/tensorflow/tensorflow/blob/ff441191277b7e758deb48e45249fee9e880f2c8/tensorflow/python/training/optimizer.py#L319). If [`use_locking`](https://github.com/tensorflow/tensorflow/blob/ff441191277b7e758deb48e45249fee9e880f2c8/tensorflow/python/training/optimizer.py#L319) is `True`, TensorFlow will use a [lock](https://github.com/tensorflow/tensorflow/blob/11e22c01eb801ff24200afcdce8a03a7cdd2ed3f/tensorflow/core/kernels/training_ops.cc#L528) to prevent concurrent updates to variables.
```
chunyang-wen: can be overwritten

elasticdl/doc/async_sgd_design.md
```diff
@@ -0,0 +1,115 @@
+# Design Doc: Asynchronous SGD
+
+## Motivation
+Parameter server (PS) based distributed training uses data parallelism to speed up training. 
+
+We have implemented synchronous SGD in ElasticDL. When PS accumulates `grads_to_wait` gradients from workers, PS averages these gradients and updates the model with the averaged gradients. PS also maintains `model_version` which equals to the number of model updates. Each worker has a local copy of the model. Before a minibatch training step, if there is a new `model_version` on PS, the worker will get the new model from PS to replace the local model. After computing the gradients with a minibatch, the worker reports the gradients to PS together with the local model version. When PS receives gradients from a worker, it only accepts the gradients with a model version same as the current PS `model_version`.
```
chunyang-wen: `Before a minibatch training step` is not a complete clause? Change it to `Before a minibatch training step starts`?

elasticdl/doc/async_sgd_design.md
```diff
@@ -0,0 +1,115 @@
...
+
+This Synchronous SGD ensures model consistency in the price of wasted and blocked computation. 
+
+* Wasted computation: when a worker reports gradients with an outdated model version to PS, PS will reject these gradients. The worker have to get the current model from PS, reuse the minibatch data to train the model again.
+* Blocked computation: PS has to use a lock for model update with gradients and model read by workers to ensure model consistency.
+
+Asynchronous SGD can avoid the wasted and blocked computation mentioned above with a relaxed model consistency.
+
+* PS will accept all gradients from workers.
+* PS does not locks and supports concurrent model reads and updates.
```
chunyang-wen: does not lock

elasticdl/doc/async_sgd_design.md
```diff
@@ -0,0 +1,115 @@
...
+def ReportGradient(gradients, version):
+    grad_var = zip(gradients, model_variables)
+    optimizer.apply_gradient(grad_var)
+    PS_model_version.atomic_add(1)
+```
+
+### Relaxed Model Consistency
+PS can processes multiple GRPC calls `GetModel` and `ReportGradients` concurrently. Thus, there are two kinds of relaxed model consistency.
+
+1. In `GetModel`, during the variable assign loop, there may be `ReportGradient` GRPC service running and updating the variables. Thus, variables in `local_model` in workers may contain values from model versions. `model_version` from `get_model_from_ps` is just a proximate model version.
```
chunyang-wen: from different model versions.

elasticdl/doc/async_sgd_design.md
```diff
@@ -0,0 +1,115 @@
...
+2. There may be multiple `ReportGradient` running concurrently. Different model variables may apply these gradients in different orders. 
+
+Also, the concurrent updates to variables in `ReportGradient` may cause some gradients are not applied, as the updates can overwritten by other concurrent running updates. TensorFlow optimizers have an argument [`use_locking`](https://github.com/tensorflow/tensorflow/blob/ff441191277b7e758deb48e45249fee9e880f2c8/tensorflow/python/training/optimizer.py#L319). If [`use_locking`](https://github.com/tensorflow/tensorflow/blob/ff441191277b7e758deb48e45249fee9e880f2c8/tensorflow/python/training/optimizer.py#L319) is `True`, TensorFlow will use a [lock](https://github.com/tensorflow/tensorflow/blob/11e22c01eb801ff24200afcdce8a03a7cdd2ed3f/tensorflow/core/kernels/training_ops.cc#L528) to prevent concurrent updates to variables.
+
+### Staleness in Asynchronous SGD
+In `ReportGradient`, the argument `version` may be smaller `PS_model_version`.
```
chunyang-wen: smaller than

elasticdl/doc/async_sgd_design.md
chunyang-wen: I think if users set `use_locking` to `True`, the updating process will be serialized. If workers wait for PS's response until PS applies the update, the training speed can be slow.  Assume that workers in the cluster have a similar training speed, it will introduce a lot of conflicts and workers have to wait for a long time until PS applies their update.

elasticdl/doc/async_sgd_design.md
```diff
@@ -0,0 +1,115 @@
...
+```
+staleness_threshold = predefined_staleness_threshold
+local_model, model_version = get_model_from_ps()
+local_update_count = 0
+for minibatch in training_data:
+    gradients = compute_gradient(local_model, minibatch)
+    apply_gradient(local_model, gradients)
+    report_gradient_to_ps(gradients, model_version)
+    local_update_count += 1
+    if local_update_count >= staleness_threshold:
```
QiJune, line 95: It seems the logic here is not correct. 
Let's image a corner case, one server and two workers. worker0 is very slow, and work1 is very fast.
At first, model_version = 0. Then work1 begin to run, it keeps updating parameters. Work0 is so slow that it does not send any gradients to server.
So, the `local_update_count` in work1 is the same with `model_version`. When `local_update_count>= staleness_threshold`, it will reset again, and keep working.
At some point, work0 finishes its first computation, it will send gradient to server, and the server will accept it! This is not expected.

QiJune: The logic should be following:

There is a shared global clock table, each work has its own clock count. We init the clock table to zero first.
```
init_clock_table()
```

Then, for each worker:


```
for minibatch in training_data:
    min_clock = get_min_clock_from_table()
    local_clock = get_clock_from_table(worker_id)
    if (local_clock - min_clock < staleness_threshold):
        model = get_model_ps()
        gradients = compute_gradient(model, minibatch)
        report_gradient_to_ps(gradients)
        inc_clock_table(worker_id)
```

chunyang-wen: Parameter server should reject gradients from workers which lags behind it for a long time.

QiJune:
Here is another paper for reference: [REVISITING DISTRIBUTED SYNCHRONOUS SGD](https://static.googleusercontent.com/media/research.google.com/zh-CN//pubs/archive/45187.pdf).

This paper proposes three ways:

- asynchronous SGD
- synchronous SGD
- synchronous SGD with backups

elasticdl/doc/async_sgd_design.md
skydoorkai: This lock applies to the individual variable update, not for the whole `optimizer.apply_gradient`. 
For example, if there are two concurrent  `optimizer.apply_gradient(g1)` and `optimizer.apply_gradient(g2)` and there are n variables in the model.  Both `optimizer.apply_gradient(g1)` and `optimizer.apply_gradient(g2)` will aquire n locks in sequence, a lock for update a variable. The performance impact is small as they are likely to aquire different locks at a time.

elasticdl/doc/async_sgd_design.md
skydoorkai: Yes, in synchronous SGD, PS will reject gradients with a very old version. 
In asynchronous SGD, PS will accept all gradients. If staleness aware learning rate modulation is on, PS will modulate the learning rate with the staleness.

skydoorkai: > The logic should be following:
> 
> There is a shared global clock table, each work has its own clock count. We init the clock table to zero first.
> 
> ```
> init_clock_table()
> ```
> 
> Then, for each worker:
> 
> ```
> for minibatch in training_data:
>     min_clock = get_min_clock_from_table()
>     local_clock = get_clock_from_table(worker_id)
>     if (local_clock - min_clock < staleness_threshold):
>         model = get_model_ps()
>         gradients = compute_gradient(model, minibatch)
>         report_gradient_to_ps(gradients)
>         inc_clock_table(worker_id)
> ```

This is ESSP [(Eager Staleness Synchronous Parallel)](https://arxiv.org/pdf/1410.8043.pdf), which requires communication among workers (or indirectly through master or global storage). ESSP will reduce the staleness compared to SSP, but in the price of the worker clock communication. This clock communication will be very complexed as ElasticDL can add or remover workers during training. So we will not add ESSP feature in the first version of asynchronous SGD. 

ESSP/SSP is usually used in synchronous SGD. We can also use this to reduce PS/worker communication in asynchronous SGD. There are no experiments to show the effectiveness of  ESSP/SSP + async SGD training yet.

elasticdl/doc/async_sgd_design.md
```diff
@@ -0,0 +1,118 @@
...
+
+We have implemented synchronous SGD in ElasticDL. When PS accumulates `grads_to_wait` gradients from workers, PS averages these gradients and updates the model with the averaged gradients. PS also maintains `model_version` which equals to the number of model updates. Each worker has a local copy of the model. Before a minibatch training step starts, if there is a new `model_version` on PS, the worker will get the new model from PS to replace the local model. After computing the gradients with a minibatch, the worker reports the gradients to PS together with the local model version. When PS receives gradients from a worker, it only accepts the gradients with a model version same as the current PS `model_version`.
+
+
+
+This Synchronous SGD ensures model consistency in the price of wasted and blocked computation. 
+
+* Wasted computation: when a worker reports gradients with an outdated model version to PS, PS will reject these gradients. The worker have to get the current model from PS, reuse the minibatch data to train the model again.
+* Blocked computation: PS has to use a lock for model update with gradients and model read by workers to ensure model consistency.
+
```
QiJune, line 14: I think that the main problem of BSP is that a global barrier is needed to ensure PS get gradients from all workers. The fast worker has to wait for slow workers.

elasticdl/doc/async_sgd_design.md
```diff
@@ -0,0 +1,118 @@
...
+Asynchronous SGD can avoid the wasted and blocked computation mentioned above with a relaxed model consistency.
+
+* PS will accept all gradients from workers.
+* PS does not use locks and supports concurrent model reads and updates.
```
QiJune, line 18: Here, we still need a RWLock?

elasticdl/doc/async_sgd_design.md
```diff
@@ -0,0 +1,118 @@
...
+## Asynchronous SGD
+Let us recall how workers train the model in synchronous SGD. Below is the pseudocode:
+
+```
+for minibatch in training_data:
+    accepted = False
+    while not accepted:
+        local_model，model_version = get_model_from_ps()
+        gradients = compute_gradient(local_model, minibatch)
+        accepted = report_gradient_to_ps(gradients, model_version)
```
QiJune, line 30:
```
accepted = report_gradient_to_ps(gradients, model_version)
```
If worker report gradient gets success, will `accpeted` be true? And then, while-loop condition is not meeted, it will finish.

elasticdl/doc/async_sgd_design.md
skydoorkai: That is the BSP implemented by TensorFlow. In each iteration, if PS needs *n* gradients to accumulate, it places *n* tokens in a queue. When a worker takes a token from the queue, it starts minibatch training and reports gradients to PS. When PS gets *n* gradients, it updates the model, and places another *n* tokens in the queue. If there are no tokens in the queue, workers need to wait for new tokens.
In ElasticDL, we have implemented a task-based synchronous SGD. Workers do not wait after `ReportGradient`, but continue to train with the next minibatch. When the worker reports gradients to PS, PS may already have updated the model and rejects the gradients. This minibatch training is wasted.

elasticdl/doc/async_sgd_design.md
skydoorkai: No need for asynch.

elasticdl/doc/async_sgd_design.md
skydoorkai: If will get the next minibatch.

elasticdl/doc/async_sgd_design.md
```diff
@@ -0,0 +1,118 @@
...
+
+### Change in PS
+1. No need to use locks in `GetModel` and `_update_model` in [server.py](../python/master/servicer.py).
+2. No need to accumulate gradients in `ReportGradient` in [server.py](../python/master/servicer.py). `ReportGradient` calls `_update_model` directly.
+3. Users decide if disabling concurrent variable update by set `use_locking` argument in the optimizer.
+4. To support [Staleness-aware asychronous SGD](https://arxiv.org/abs/1511.05950), PS need to modulate the learning rate in the optimizer with the staleness value.
+
+### Change in Worker
+1. No need to retrain with the minibatch data.
+2. To support SSP strategy, the worker pulls the model from PS in every `staleness_threshold` minibatch steps. Also, the worker needs to update the local model with the computed gradients.
```
mhaoli: Here we don't need to update the embedding table when updating the local model. This is because we should pull embedding vectors from embedding service every minibatch.

skydoorkai: Yes, the local model only includes non-embedding variables.

mhaoli (approved):

skydoorkai:
Close #1125

PR #1130, opened 2019-09-06
Fixes #1129, add log for command line options

### LINKED ISSUE #1129: Print information on master and worker
Both master and worker have a lot of command-line options.  Print values of these options to help users trace problems if there exists any.

chunyang-wen:
Fixes #1129 

Print out command-line options to help users trace problem.

### DIFF
--- a/elasticdl/python/common/args.py
+++ b/elasticdl/python/common/args.py
@@ -1,3 +1,44 @@
+from elasticdl.python.common.log_util import default_logger as logger
+
+MODEL_SPEC_GROUP = [
+    "dataset_fn",
+    "eval_metrics_fn",
+    "model_def",
+    "model_params",
+    "optimizer",
+    "loss",
+    "output",
+    "minibatch_size",
+    "grads_to_wait",
+    "num_epochs",
+    "tensorboard_log_dir",
+    "training_data_dir",
+]
+
+EVALUATION_GROUP = [
+    "evaluation_steps",
+    "evaluation_data_dir",
+    "evaluation_start_delay_secs",
+    "evaluation_throttle_secs",
+]
+
+PREDICTION_GROUP = ["prediction_data_dir"]
+
+CHECKPOINT_GROUP = [
+    "checkpoint_filename_for_init",
+    "checkpoint_steps",
+    "keep_checkpoint_max",
+    "checkpoint_dir",
+]
+
+ALL_ARGS_GROUPS = [
+    MODEL_SPEC_GROUP,
+    EVALUATION_GROUP,
+    PREDICTION_GROUP,
+    CHECKPOINT_GROUP,
+]
+
+
 def pos_int(arg):
     res = int(arg)
     if res <= 0:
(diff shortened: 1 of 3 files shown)


terrytangyuan:
A couple questions:
1. What’s an example output look like? 
2. Why do we need this? Aren’t users already aware of the commands they are submitting? 
3. Does this print out optional args that have defaults?

chunyang-wen:
> A couple questions:
> 
> 1. What’s an example output look like?
> 2. Why do we need this? Aren’t users already aware of the commands they are submitting?
> 3. Does this print out optional args that have defaults?

1. 
image

2. Users can submit multiple jobs through shell scripts which can generate different arguments or they are investigating different combinations of parameters. I think it is helpful if they can view their configurations in the log. It is a little difficult to track those parameters themselves.

3. What do you mean by optional args? It will print all the values we define in the parser.

terrytangyuan:
@chunyang-wen For example, `parser.add_argument(
        "--worker_image", help="Docker image for workers", default=None
    )` is an optional arg which has a default. Will this default be printed out as well?

Regarding the output, would it be better to print them out in a single line just like what the user typed in the console?
```
elasticdl train --optimizer=...
```
This would make it easier for users to copy the command to reproduce their experiment results.

chunyang-wen:
> @chunyang-wen For example, `parser.add_argument( "--worker_image", help="Docker image for workers", default=None )` is an optional arg which has a default. Will this default be printed out as well?
> 
> Regarding the output, would it be better to print them out in a single line just like what the user typed in the console?
> 
> ```
> elasticdl train --optimizer=...
> ```
> 
> This would make it easier for users to copy the command to reproduce their experiment results.

Yes, the default value will be printed out. I can filter them out if it is not suitable to show them.

Would it be a little weird to print everything on a single line? Because the line width it occupies may be very long.

terrytangyuan:
Currently each printed line only contains one item will make the log very verbose. I personally think this kind of information/metadata should be tracked by some kind of experiment tracker such as Kubeflow Pipelines. cc @ywskycn What do you think?

chunyang-wen:
@terrytangyuan These logs will be printed only once. In a typical training loop, training logs will be the dominated logs. If there is a better way to track this information, the log is still a convenient way to check arguments. @ywskycn any suggestion is welcomed.

chunyang-wen:
@terrytangyuan These logs will be printed only once. In a typical training loop, training logs will be the dominated logs. If there is a better way to track this information, the log is still a convenient way to check arguments. @ywskycn any suggestion is welcomed.

ywskycn:
Sorry for late reply...  @chunyang-wen . 
Yes, this type of information is needed in production, especially when something goes wrong and we need more information to debug.

> I personally think this kind of information/metadata should be tracked by some kind of experiment > tracker such as Kubeflow Pipelines.
@terrytangyuan yes, this can be recorded by another service, and maybe used for reproducibility. But the job itself still can print it for own purpose. Should be fine here.

elasticdl/python/common/args.py
```diff
@@ -1,3 +1,44 @@
+from elasticdl.python.common.log_util import default_logger as logger
+
+MODEL_SPEC_GROUP = [
```
ywskycn, line 3: @chunyang-wen @terrytangyuan is there any way to auto-fetch the list of parameters in Python? Little worried when we add new parameters but forget add here, it will cause inconsistency.

terrytangyuan: Not on top of my head. Yes I am a little worried about this as well since this will easily get outdated if we forget to add them to the groups.

chunyang-wen: All the arguments will get printed. The groups only help to print related arguments together.

ywskycn (approved):

PR #1135, opened 2019-09-09
expose local docker host args

### LINKED ISSUE #1134: Expose local docker host args
Currently, we use docker python api to build docker images locally. The docker host is be set to "unix://var/run/docker.sock" by default. Please refer to 

https://github.com/sql-machine-learning/elasticdl/blob/1727e2368abffe5b8179c26c9aefa420e4bbc8cd/elasticdl/python/elasticdl/image_builder.py#L67-L69

However, sometimes, we may use another docker host. For example, we may use docker host of minikube. The docker host will be an another one. Or even we have to connect a docker host in remote.

So, the local docker host related args have to be exposed.

QiJune:
Fix #1134

### DIFF
--- a/elasticdl/python/common/args.py
+++ b/elasticdl/python/common/args.py
@@ -163,14 +163,14 @@ def add_common_params(parser):
     )
     parser.add_argument(
         "--docker_base_url",
-        help="The docker host url",
+        help="URL to the Docker server",
         default="unix://var/run/docker.sock",
     )
     parser.add_argument(
-        "--docker_tlscert", help="The tlscert of docker client", default=""
+        "--docker_tlscert", help="Path to Docker client cert", default=""
     )
     parser.add_argument(
-        "--docker_tlskey", help="The tlskey of docker client", default=""
+        "--docker_tlskey", help="path to Docker client key", default=""
     )
 
 
--- a/elasticdl/python/common/args.py
+++ b/elasticdl/python/common/args.py
@@ -161,6 +161,17 @@ def add_common_params(parser):
         help="The file that contains user-defined cluster specification",
         default="",
     )
+    parser.add_argument(
+        "--docker_base_url",
+        help="The docker host url",
+        default="unix://var/run/docker.sock",
+    )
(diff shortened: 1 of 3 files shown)


terrytangyuan:
Just curious - what's the correct docker related args that you used so you could get everything working successfully locally?

QiJune:
@terrytangyuan  I will put this to our document. 
Run `minikube docker-env` first, you will get following

```
export DOCKER_TLS_VERIFY="1"
export DOCKER_HOST="tcp://192.168.64.5:2376"
export DOCKER_CERT_PATH="/Users/qijun/.minikube/certs"
```

And then, you can pass these values to docker APIClient.

```python
import docker
tls_config = docker.tls.TLSConfig(
  client_cert=('/Users/qijun/.minikube/cert.pem', '/Users/qijun/.minikube/key.pem')
)
c = docker.APIClient(base_url="tcp://192.168.64.5:2376", tls=tls_config)
print(c.version())
```

It will connect minikube docker host successfully.

elasticdl/python/common/args.py
```diff
@@ -161,6 +161,17 @@ def add_common_params(parser):
...
+    parser.add_argument(
+        "--docker_base_url",
+        help="URL to the Docker server",
+        default="unix://var/run/docker.sock",
+    )
+    parser.add_argument(
+        "--docker_tlscert", help="Path to Docker client cert", default=""
+    )
+    parser.add_argument(
+        "--docker_tlskey", help="path to Docker client key", default=""
```
mhaoli: path -> Path

QiJune: Done

terrytangyuan (approved):

ywskycn:
@QiJune should be `eval $(minikube docker-env)`?

QiJune:
@ywskycn I run this command already. 
Since we use docker python API to build images, we should pass correct docker base url to docker.APIClient. 
Currently, the base_url is "unix://var/run/docker.sock" by default.

PR #1156, opened 2019-09-10
Delete lock in  `GetModel` and `_update_model`

chunyang-wen:
The lock in `GetModel` function can clearly removed. But there is also a lock in
+ `_update_model` from `ReportGradient`
+ `_update_model_version`

They depend on the lock from their parent's function.

### DIFF
--- a/elasticdl/python/master/servicer.py
+++ b/elasticdl/python/master/servicer.py
@@ -175,8 +175,10 @@ class MasterServicer(elasticdl_pb2_grpc.MasterServicer):
 
     def _update_model(self):
         if not self._use_async and not self._lock.locked():
+            # TODO (chengfu.wcy) `self._lock.locked` may be removed
+            # according to  changes in `ReportGradient` in async mode.
             raise RuntimeError(
-                "Lock must be acquired when updating " "the model in sync mode"
+                "Lock must be acquired when updating the model in sync mode"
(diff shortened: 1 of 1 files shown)


elasticdl/python/master/servicer.py
```diff
@@ -170,7 +174,8 @@ def _update_edl_embedding_table(self, name_var_list):
             )
 
     def _update_model(self):
-        assert self._lock.locked()
+        if not self._use_async and not self._lock.locked():
+            raise RuntimeError("In sync mode, lock is not acquired")
```
terrytangyuan: How about "Lock must be acquired when updating the model in sync mode"?

terrytangyuan:
@chunyang-wen Did you forget to update/complete the PR title? I am not sure what "Delete lock in and" means.

elasticdl/python/master/servicer.py
chunyang-wen: It seems better.

chunyang-wen:
> @chunyang-wen Did you forget to update/complete the PR title? I am not sure what "Delete lock in and" means.

I use \`GetModel\` and \`_update_model\` in git's message. It seems that Github fails to parse it.

terrytangyuan (approved):

PR #1157, opened 2019-09-10
Add missing TRAIN clause in SQL example

terrytangyuan:


index.html
```diff
@@ -954,7 +954,7 @@ <h3>Integration with SQLFlow</h3>
                   <p>ElasticDL will be integrated seamlessly with SQLFlow to connect SQL to distributed deep learning tasks with ElasticDL.</p>
                   <pre>
                     <code class="sql">
-SELECT * FROM employee LABEL income INTO my_elasticdl_model
+SELECT * FROM employee TRAIN DNNRegressor LABEL income INTO my_elasticdl_model
```
ywskycn, line 957: This is for auto-model selection, right?

terrytangyuan: Yes it was originally written for the auto model selection but since we updated the description  without any mentions of automl, we should probably add this TRAIN clause.

PR #1133, opened 2019-09-08
Test the correctness of optimizer_wrapper for momentum SGD

mhaoli:
Part of #1123, test the correctness of optimizer_wrapper for momentum SGD.

PR #1158, opened 2019-09-10
support all tf optimizers

### LINKED ISSUE #1123: Test the correctness of optimizer wrapper
Currently, we have unit test tests for functions in optimizer wrapper. 

Besides, we should test the correctness of optimizer wrapper for every TensorFlow optimizer.
For a specific TensorFlow optimizer, we can train two models and compare trainable parameters directly, one with Keras embedding, another one with ElasticDL embedding and optimizer wrapper.

mhaoli:
Fixes #1123. This is a follow-up PR of #1133 to support all optimizers in TensorFlow and test the correctness of them.

### DIFF
--- a/elasticdl/python/tests/optimizer_wrapper_test.py
+++ b/elasticdl/python/tests/optimizer_wrapper_test.py
@@ -6,7 +6,17 @@ import unittest
 import mock
 import numpy as np
 import tensorflow as tf
-from tensorflow.keras.optimizers import SGD, Adam
+from tensorflow.keras.optimizers import (
+    SGD,
+    Adadelta,
+    Adagrad,
+    Adam,
+    Adamax,
+    Ftrl,
+    Nadam,
+    RMSprop,
+)
+from tensorflow.python.ops import init_ops
 
 from elasticdl.python.common.model_helper import (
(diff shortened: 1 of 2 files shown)


elasticdl/python/tests/optimizer_wrapper_test.py
```diff
@@ -535,8 +569,189 @@ def _test_correctness(self, optimizer_class, **kwargs):
...
+            random_seed=seed,
+        )
+        self._test_correctness(SGD, X, Y, seed, learning_rate=0.1)
+        self._test_correctness(
+            SGD, X, Y, seed, learning_rate=0.1, momentum=0.5
+        )
+
+    def test_adadelta_correctness(self):
+        seed = 1
+        X, Y = _prepare_random_data(
```
terrytangyuan: Can we reuse this across tests and then maybe write a loop that iterates through different optimizers, finally check `self._test_correctness`? There seems to be a lot of duplicated code here.

mhaoli: Yes! Done.

terrytangyuan (approved):

Issue #1105, opened 2019-09-03
Support reading from custom data sources


terrytangyuan:
Currently ElasticDL only supports reading from RecordIO files, we should make this more flexible so users can read data from custom sources, such as ODPS.

PR #1159, opened 2019-09-10
[Design] Support Reading from Custom Data Source

terrytangyuan:
Part of https://github.com/sql-machine-learning/elasticdl/issues/1105.

### DIFF
--- a/elasticdl/doc/custom_data_source_support_design.md
+++ b/elasticdl/doc/custom_data_source_support_design.md
@@ -2,77 +2,6 @@
 
 This document describes the design for supporting custom data source in ElasticDL.
 
-### Example Reader Class - `ODPSReader`
-
-Reading data from ODPS data source is through the existing `ODPSReader` which has the following interface:
-
-````python
-class ODPSReader(object):
-    def __init__(
-        self,
-        project,
-        access_id,
-        access_key,
-        endpoint,
-        table,
-        partition=None,
-        num_processes=None,
-        options=None,
-    ):
-        """
-        Constructs a `ODPSReader` instance.
-
-        Args:
-            project: Name of the ODPS project.
-            access_id: ODPS user access ID.
-            access_key: ODPS user access key.
-            endpoint: ODPS cluster endpoint.
-            table: ODPS table name.
-            partition: ODPS table's partition.
-            options: Other options passed to ODPS context.
-            num_processes: Number of parallel processes on this worker.
-                If `None`, use the number of cores.
-        """
-        pass
-
-    def to_iterator(
-        self,
-        num_workers,
-        worker_index,
-        batch_size,
-        epochs=1,
-        shuffle=False,
-        columns=None,
-        cache_batch_count=None,
-        limit=-1,
-    ):
-        """
-        Load slices of ODPS table (partition of table if `partition`
-        was specified) data with Python Generator.
-
-        Args:
-            num_workers: Total number of worker in the cluster.
-            worker_index: Current index of the worker in the cluster.
-            batch_size: Size of a slice.
-            epochs: Repeat the data for this many times.
-            shuffle: Whether to shuffle the data or rows.
-            columns: The list of columns to load. If `None`,
-                use all schema names of ODPS table.
-            cache_batch_count: The cache batch count.
-            limit: The limit for the table size to load.
-        """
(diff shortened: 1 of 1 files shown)


elasticdl/doc/custom_data_source_support_design.md
```diff
@@ -0,0 +1,77 @@
...
+RecordIO files and reading the index for each file to obtain dictionaries of `{file_path: num_records}`.
+We then use this information to create tasks.
+* ``Worker`` starts a `TaskDataService` that generates a `tf.data.Dataset` instance that contains the actual data.
+The data is fetched in `TaskDataService._gen()` by opening a RecordIO reader to read the data offsets that each task
+we previously created.
+
+In order to support custom data sources in ElasticDL, we propose to make the following changes:
+
+* Rename ``shard_file_name`` in `Task` protobuf definition to `shard_name` so a task is independent with file names and `shard_name` will
+serve as an identifier for the shard.
```
ywskycn, line 36: So the new `shard` represents a file, or a table, or sth else?

elasticdl/doc/custom_data_source_support_design.md
```diff
@@ -0,0 +1,77 @@
...
+
+    @abstractmethod
+    def read_records(self, task):
+        """This method will be used in `TaskDataService` to read the records based on
+        the information provided for a given task into a Python generator/iterator.
+        """
+        pass
+
+    @abstractmethod
+    def create_task_dispatcher(self):
```
ywskycn: May need to change the func name. At the first glance, I thought users need to handle the creation of training/evaluation tasks. After seeing the following example, it looks like just a dict describing how many records inside each shard. In that case, why not just one AbstractDataReader maps to one dataset, and we don't mix training/evaluation/prediction together?

elasticdl/doc/custom_data_source_support_design.md
terrytangyuan: It represents a set that contains multiple records, which can be file, table, etc.

elasticdl/doc/custom_data_source_support_design.md
terrytangyuan: Good idea! The creation of `TaskDispatcher` object that's returned in this function should not be done by the user. 

Three `AbstractDataReader`s will increase the number of CLI args that we need to add to the client. How about creating three separate methods in this class, e.g. `create_training_shards`, `create_evaluation_shards`, and `create_prediction_shards`?

elasticdl/doc/custom_data_source_support_design.md
```diff
@@ -0,0 +1,77 @@
...
+There are a couple of places in the the current codebase (as of commit [77cc87a](https://github.com/sql-machine-learning/elasticdl/tree/77cc87a90eec54db565849f0ae07d271fd957190))
+that are coupled heavily with RecordIO, for example:
+
+* When we create and dispatch tasks via ``_make_task_dispatcher()``, we rely on scanning directories that contain
+RecordIO files and reading the index for each file to obtain dictionaries of `{file_path: num_records}`.
+We then use this information to create tasks.
+* ``Worker`` starts a `TaskDataService` that generates a `tf.data.Dataset` instance that contains the actual data.
+The data is fetched in `TaskDataService._gen()` by opening a RecordIO reader to read the data offsets that each task
+we previously created.
+
```
wangkuiyi, line 32: This is a logical gap between these two paragraphs. How do we connect the fact of being tightly-coupled with RecordIO and the requirement of custom data sources?

Is there any problem with being tightly-coupled with RecordIO? Is custom data sources a solution to the problem?

terrytangyuan: Thanks for the questions! I've added some brief background and motivation on this to the design doc. Please check it out and continue our discussion here.

elasticdl/doc/custom_data_source_support_design.md
chunyang-wen: `create_shards` with a mode defined in `constants.Mode`?

terrytangyuan: @chunyang-wen Good suggestion. Fixed.

wangkuiyi (approved):

PR #1162, opened 2019-09-11
Switch to OptimizerWrapper

### LINKED ISSUE #1161: Switch to OptimizerWrapper
After PR #1158, OptimizerWrapper is ready for synchronous SGD. We can switch to OptimizerWrapper.

mhaoli:
Fixes #1161.

OptimizerWrapper is not thread-safe. Thread-safe OptimizerWrapper will implement in a seprate PR.

### DIFF
--- a/elasticdl/python/master/main.py
+++ b/elasticdl/python/master/main.py
@@ -172,6 +172,7 @@ def main():
         task_d.set_evaluation_service(evaluation_service)
 
     embedding_service_endpoint = None
+    embedding_dims = {}
     # Search for embedding layers in the model,
     # if found, initialize embedding service
     layers = find_layer(model_inst, Embedding)
@@ -193,6 +194,9 @@ def main():
             "Embedding service start succeeded. The endpoint is %s."
             % str(embedding_service_endpoint)
         )
+        embedding_dims = dict(
+            [(layer.name, layer.output_dim) for layer in layers]
+        )
 
     # The master service
     logger.info("Starting master service")
@@ -212,6 +216,7 @@ def main():
         optimizer,
         task_d,
         init_var=model_inst.trainable_variables if model_inst.built else [],
+        embedding_dims=embedding_dims,
         checkpoint_filename_for_init=args.checkpoint_filename_for_init,
         checkpoint_service=checkpoint_service,
         evaluation_service=evaluation_service,
--- a/elasticdl/python/master/optimizer_wrapper.py
+++ b/elasticdl/python/master/optimizer_wrapper.py
@@ -121,8 +121,8 @@ class OptimizerWrapper(object):
             need_create_var: A python bool. If `need_create_var` is True,
                 OptimizerWrapper creates temporary variables for embedding
                 vectors and slots every iteration. If `need_create_var` is
-                False, OptimizerWrapper create persistent variables before
-                training and use these variables during the whole training
+                False, OptimizerWrapper creates persistent variables before
+                training and uses these variables during the whole training
                 process.
         """
         self._opt = opt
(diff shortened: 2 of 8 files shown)


elasticdl/python/master/optimizer_wrapper.py
```diff
@@ -100,14 +100,16 @@ def _var_key(var):
 class OptimizerWrapper(object):
     """ ElasticDL optimizer wrapper.
 
-    If model does not use ElasticDL embedding layer, `OptimizerWrapper`
+    If model does not use ElasticDL embedding layer, OptimizerWrapper
```
terrytangyuan: Why changing this line? You can keep them since you are referring a particular class

elasticdl/python/tests/mock_kv_store.py
```diff
@@ -23,11 +27,16 @@ def lookup(self, keys=[], *args, **kwargs):
                 unknown_key.append(i)
         return values, unknown_key
 
+    # keep same with EmbeddingService.update_embedding
```
terrytangyuan: What does this comment mean? You can make this a docstring instead of a comment

elasticdl/python/tests/mock_kv_store.py
```diff
@@ -12,7 +13,10 @@ def __init__(self, store={}):
                     )
         self._store = store
 
-    def lookup(self, keys=[], *args, **kwargs):
+    # keep same with EmbeddingService.lookup_embedding
```
terrytangyuan: same here

elasticdl/python/master/optimizer_wrapper.py
```diff
@@ -116,10 +118,17 @@ def __init__(self, opt, kv_store_endpoint, embedding_dims):
                 {layer name: `embedding_dim`} where layer name is the
                 name of ElasticDL embedding layer and `embedding_dim`
                 is the output dimension of corresponding embedding layer.
+            need_create_var: A python bool. If `need_create_var` is True,
+                OptimizerWrapper creates temporary variables for embedding
```
terrytangyuan: What does temporary variables mean here? Also maybe add comment on the relationship with `use_async`

elasticdl/python/master/main.py
```diff
@@ -193,6 +194,9 @@ def main():
             "Embedding service start succeeded. The endpoint is %s."
             % str(embedding_service_endpoint)
         )
+        embedding_dims = dict(
```
terrytangyuan, line 197: Maybe more this logic to inside `MasterServicer` and pass layers to keep `main.py`'s logic intuitive?

mhaoli: `MasterService` already has two arguments that are used by parameter server. They are `init_var` and `embedding_service_endpoint`. `embedding_dims` is also used by parameter server. We can put `embedding_dims` here. Moving these three arguments inside `MasterServicer` can be done in a separate PR.

elasticdl/python/master/optimizer_wrapper.py
mhaoli: Done.

elasticdl/python/master/optimizer_wrapper.py
mhaoli: 1. temporary variables means that `OptimizerWrapper` creates these variables every iteration. In one iteration, `OptimizerWrapper` creates these variables before calling optimizer's `apply_gradients` function and deletes these variables after calling `apply_gradients`.
2. Goot point. Putting `need_create_var` in the constructor will confuse users. Users only care about whether `OptimizerWrapper` is used in synchronous or asynchronous situation. I will put `use_async` argument here.

elasticdl/python/tests/mock_kv_store.py
mhaoli: Done.

elasticdl/python/tests/mock_kv_store.py
mhaoli: Done.

terrytangyuan (approved):

ywskycn (approved):

PR #1164, opened 2019-09-11
Skip gradient accumulation and update model directly when using async-SGD

### LINKED ISSUE #1138: When async-SGD, ReportGradient does not accumulate gradients, but apply_gradients directly
According to [design doc](https://github.com/sql-machine-learning/elasticdl/blob/develop/elasticdl/doc/async_sgd_design.md#change-in-ps).

No need to accumulate gradients in ReportGradient in server.py. ReportGradient calls _update_model directly.

terrytangyuan:
Fixes #1138. 

### DIFF
--- a/elasticdl/python/master/servicer.py
+++ b/elasticdl/python/master/servicer.py
@@ -185,7 +185,10 @@ class MasterServicer(elasticdl_pb2_grpc.MasterServicer):
         # (grad, var) pairs excluding keras Embedding layer and
         # ElasticDL Embedding layer
         for k in self._gradient_sum:
-            self._gradient_sum[k] = self._gradient_sum[k] / self._grad_to_wait
+            if not self._use_async:
+                self._gradient_sum[k] = (
+                    self._gradient_sum[k] / self._grad_to_wait
+                )
             grad_var.append((self._gradient_sum[k], self._model[k]))
 
         # (grad, var) pair of Keras Embedding layer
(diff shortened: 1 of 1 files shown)


elasticdl/python/master/servicer.py
```diff
@@ -185,7 +185,10 @@ def _update_model(self):
         # (grad, var) pairs excluding keras Embedding layer and
         # ElasticDL Embedding layer
         for k in self._gradient_sum:
-            self._gradient_sum[k] = self._gradient_sum[k] / self._grad_to_wait
+            if not self._use_async:
+                self._gradient_sum[k] = (
```
chunyang-wen, line 189: How about totally skipping the `for` loop and using `zip` to generate the `var` and `grad` tuple?

elasticdl/python/master/servicer.py
```diff
@@ -402,13 +405,13 @@ def ReportGradient(self, request, _):
 
             # other grads
             for k, v in tmp.items():
-                if k in self._gradient_sum:
+                if not self._use_async and k in self._gradient_sum:
                     self._gradient_sum[k] = self._gradient_sum[k] + v
```
chunyang-wen, line 409: We only need to check `self._use_async` once. Checking `self._use_async` inside the `for` loop seems that it will hurt the performance.

chunyang-wen (approved):

terrytangyuan:
We can improve the performance later. Merging this now to unblock @skydoorkai's work.

PR #1165, opened 2019-09-11
Remove unnecessary member variable and None event callback

terrytangyuan:
* The k8s client is only used locally within this function so no need to make it a member variable for the class.
* Removed unnecessary `event_callback=None` since `k8s.Client` has this as default.

### DIFF
--- a/elasticdl/python/master/embedding_service.py
+++ b/elasticdl/python/master/embedding_service.py
@@ -242,8 +242,8 @@ class EmbeddingService(object):
         **kargs,
     ):
         logger.info("Starting pod for embedding service ...")
-        self._k8s_client = k8s.Client(event_callback=None, **kargs)
-        pod = self._k8s_client.create_embedding_service(
+        k8s_client = k8s.Client(**kargs)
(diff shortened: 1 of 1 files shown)


elasticdl/python/master/embedding_service.py
```diff
@@ -259,9 +259,7 @@ def start_embedding_pod_and_redis(
         # and should not fix ports
         address_ip = pod.status.pod_ip
         while not address_ip:
-            pod = self._k8s_client.get_embedding_service_pod(
-                embedding_service_id
-            )
+            pod = k8s_client.get_embedding_service_pod(embedding_service_id)
```
chunyang-wen, line 262: should we wait for a while before the next check?

terrytangyuan: That's not relevant to this PR. Perhaps consider creating a new issue to discuss?

chunyang-wen: :smiley:

chunyang-wen (approved):

PR #1172, opened 2019-09-12
add staleness-aware learning rate modulation

### LINKED ISSUE #1139: Support staleness-aware async-SGD
According to [design doc](https://github.com/sql-machine-learning/elasticdl/blob/develop/elasticdl/doc/async_sgd_design.md#change-in-ps)

To support Staleness-aware asychronous SGD, PS need to modulate the learning rate in the optimizer with the staleness value. PS may have multiple threads running concurrently for model updates with the same optimizer instance. Thus, we cannot modify the learning rate in the optimizer instance. We may modify the learning rate as a callable method, and use a thread local storage threading.local() to store the staleness. The callable method uses the staleness value to modulate the learning rate. The optimizer will call this callable method when it reads the learning rate hyperparameter.

skydoorkai:
fix #1139


elasticdl/python/tests/staleness_aware_test.py
```diff
@@ -0,0 +1,87 @@
...
+from elasticdl.python.master.lr_modulation import (
+    add_lr_modulation_to_optimizer,
+)
+
+
+class LearningRateTest(unittest.TestCase):
+    @staticmethod
+    def get_lr(lr_modulation, opt, multiplier):
+        lr_modulation.set_multiplier(multiplier)
+        time.sleep(1)
```
mhaoli, line 17: Perhaps add a comment here explaining the reason for sleeping 1s?

elasticdl/python/tests/staleness_aware_test.py
```diff
@@ -0,0 +1,87 @@
...
+                for j in range(len(original_values))
+            ]
+            for j in range(len(original_values)):
+                # variable value change ratio equals the learning rate ratio
+                # for SGD without momentum
+                self.assertAlmostEqual(
+                    i_diff[j] / first_diff[j],
+                    multipliers[i] / multipliers[0],
+                    place,
+                )
```
mhaoli, line 79: Why don't we compare `i_diff[j] == lr * multipliers[i] * grads[j]` directly?

elasticdl/python/master/servicer.py
```diff
@@ -55,6 +59,7 @@ def __init__(
         self._version = 0
         self._embedding_service_endpoint = embedding_service_endpoint
         self._init_model(checkpoint_filename_for_init, init_var)
+        self._modulate_lr_if_needed()
```
mhaoli: Maybe we can call `self._modulate_lr_if_needed` in function `self._init_optimizer`? `self._init_optimizer` is also called after `self._init_model` in develop branch now. 

https://github.com/sql-machine-learning/elasticdl/blob/cecec52a9e856f00ff4d5e17dc782ec2602404de/elasticdl/python/master/servicer.py#L58-L61

mhaoli (approved):

PR #1181, opened 2019-09-15
Parameter server design

QiJune:
Currently, we have two concepts, master and worker concepts. We need to split pserver from master, and make the functionality of each concept clear. Following is an overview design doc.

[Here](https://github.com/QiJune/elasticdl/blob/ps_design/doc/parameter_server.md) is the link which for a better review.


doc/parameter_server.md
```diff
@@ -0,0 +1,98 @@
...
+Please refer to:
+
+![parameter_server](./images/parameter_server.png)
+
+## Key concepts
+
+
+### Parameter sharding
+
+We usually shard big parameters of a neural network model to some nodes under a rule, in order to achieve network load banlance. There are several kinds of parameter to be handled seperately:
```
chunyang-wen: Typo: seperately.

doc/parameter_server.md
```diff
@@ -0,0 +1,98 @@
...
+
+- Very big embedding table: Embedding table is a collection of <item id, embedding vector> pairs. There is also a mapping from item id to pserver id, so the corresonding embedding vector will be stored at the certain pserver
+
+- Big dense tensor: If the dense tensor parameter exceeds certain size, it will be sliced into several subtensors. The parameter name combines subtensor id will become a unique key, and the value is the subtensor(Please note that the slice tensor operation is zero-copy)
```
chunyang-wen: `The parameter name combines subtensor id will become a unique key`

doc/parameter_server.md
```diff
@@ -0,0 +1,98 @@
...
+
+- Small dense tensor: The small dense tensor parameter will be stored at certain pserver wholely
+
+### Parameter Initialization
+
+
+Parameter initialization of very big embedding table is lazy. For example, in online learing, there could be unkown item id in the training data. So, after worker send the unkown item id to pserver, pserver will initialize corresponding embedding vector and send back to worker. This is a get_or_create semantic.
```
chunyang-wen: **a** very big

chunyang-wen: online learing.

chunyang-wen: Multiple grammar errors. Please double check the sentences.

chunyang-wen:
This design is very similar to the design of [psplus of XDL](https://github.com/alibaba/x-deeplearning/tree/master/xdl/ps-plus). FYI.

doc/parameter_server.md
```diff
@@ -0,0 +1,98 @@
+# Parameter Server Design
+
+## Overview
+
+
+A typical parameter server architecture contains three roles:
+
+- master, responsible for creating/deleting/scheduling pservers and workers
```
wangkuiyi: The master has two main goals:

1. call Kubernetes API to start workers and optionally parameter servers, and
1. separate data into tasks and dispatch tasks to workers.

doc/parameter_server.md
```diff
@@ -0,0 +1,98 @@
+# Parameter Server Design
+
+## Overview
+
+
+A typical parameter server architecture contains three roles:
```
wangkuiyi: parameter server (pserver)

doc/parameter_server.md
```diff
@@ -0,0 +1,98 @@
...
+
+- master, responsible for creating/deleting/scheduling pservers and workers
+- pserver, responsible for provide parameter pull/push/optimize/checkpoint service
```
wangkuiyi: I think the primary goal of the parameter servers of ElasticDL is to maintain the **global** model. The word "parameter" refers to parameters of the global model.  This is in contrast with the fact that workers have their local models.

doc/parameter_server.md
```diff
@@ -0,0 +1,98 @@
...
+- worker, responsible for training minibatches of data
```
wangkuiyi: Workers compute gradients of model parameters using local data and collaboratively update the global model, either via Kubernetes-native fault-tolerable AllReduce or a set of parameter server processes.

doc/parameter_server.md
wangkuiyi: I think we need a few more sentences here explaining why the global model is sharded but the local models are not.

> We prefer to shard the global model for some reasons:
>
> 1.  A model could be too large to fit in the memory of a process.  For example, many recommending and ranking models take very high-dimensional (and sparse) inputs, thus require large embedding tables. In such cases, we'd have to have multiple parameter server instances to hold the global model.
> 1. Even when a single process can hold the global model, we might still shard it onto multiple parameter server instances, who share the optimization workload.

doc/parameter_server.md
```diff
@@ -0,0 +1,98 @@
...
+### Parameter sharding
+
+We usually shard big parameters of a neural network model to some nodes under a rule, in order to achieve network load banlance. There are several kinds of parameter to be handled seperately:
+
+- Very big embedding table: Embedding table is a collection of <item id, embedding vector> pairs. There is also a mapping from item id to pserver id, so the corresonding embedding vector will be stored at the certain pserver
+
+- Big dense tensor: If the dense tensor parameter exceeds certain size, it will be sliced into several subtensors. The parameter name combines subtensor id will become a unique key, and the value is the subtensor(Please note that the slice tensor operation is zero-copy)
+
+- Small dense tensor: The small dense tensor parameter will be stored at certain pserver wholely
+
```
wangkuiyi, line 33: To make a complete discussion about model sharding, we also need to talk about the local model.

Indeed, we assume the local model is small enough that workers don't need to shard them. We need to explain why.

> The local model on workers and the global model on the parameter server(s) have the same size. In case that large embedding tables make the model size too big to fit in a single process's memory space, an intuitive solution is to save the model copies on an external storage service like Redis or Memcached.
>
> However, such a solution creates a complex workflow which introduces too many cross-node communications.
>
> We propose an alternative solution that doesn't rely on Redis or Memcachd.  Instead, the global model, including the large embedding tables, is sharded across parameter servers, and the local model contains only part of the embedding table -- the small fraction that is required to compute recent minibatch(es).

doc/parameter_server.md
```diff
@@ -0,0 +1,98 @@
...
+Parameter initialization of very big embedding table is lazy. For example, in online learing, there could be unkown item id in the training data. So, after worker send the unkown item id to pserver, pserver will initialize corresponding embedding vector and send back to worker. This is a get_or_create semantic.
+
+Other parameters could be initialized before training.
+
+
+## Workflow
+
+### Assumption
+
+- only support asynchronous SGD
```
skydoorkai: synchronous SGD not supported?

doc/parameter_server.md
```diff
@@ -0,0 +1,98 @@
...
+- only support worker failover, do not support pserver
```
skydoorkai: do not support pserver -> do not support pserver failover

doc/parameter_server.md
QiJune: Yes, it is. The local model is pulled from pserver, and is usually a part of the global model.

doc/parameter_server.md
QiJune: I think we may take different solutions for different scenarios.


|  scenario   | hardware  |  update stragety |  communication  stragety |
| :----  |:----  |:---|:---|
| recommend  | CPU|  Async |  PS |
| CV | GPU | BSP| Allreduce|
| NLP | GPU | BSP| Allreduce |

This design addresses the first scenario.

doc/parameter_server.md
QiJune: Do we need a data loader role here? The master will also create some data loaders, and these data loaders will prefetch data from local/remote database, and send data to workers.
And it seems that, we can also add some data preprocessing logic to data loader.

doc/parameter_server.md
QiJune: Done.

doc/parameter_server.md
QiJune: Done

doc/parameter_server.md
QiJune: Done.

doc/parameter_server.md
QiJune: Done

doc/parameter_server.md
QiJune: Done

doc/parameter_server.md
```diff
@@ -0,0 +1,144 @@
...
+
+Following is the assumption under recommendation scenario:
+
+- only support asynchronous SGD
+- only support worker failover, do not support pserver failover
+
+
+**Note**
+
+Allreduce failover and pserver failover will be discussed on other design docs.
```
wangkuiyi: Allreduce => AllReduce

doc/parameter_server.md
```diff
@@ -0,0 +1,144 @@
...
+
+### Assumption
+
+We apply different solutions to different scenarios.
+
+
+| scenario| hardware | update stragety | communication  stragety |
+| :----  |:----  |:---|:---|
+| Recommendation  | CPU|  Async |  Parameter server |
+| CV | GPU | BSP| Allreduce|
```
wangkuiyi: Need an explanation of what BSP is short for.

doc/parameter_server.md
```diff
@@ -0,0 +1,144 @@
...
+## Workflow
+
+### Assumption
+
+We apply different solutions to different scenarios.
+
+
+| scenario| hardware | update stragety | communication  stragety |
+| :----  |:----  |:---|:---|
+| Recommendation  | CPU|  Async |  Parameter server |
```
wangkuiyi: Async => async SGD

doc/parameter_server.md
```diff
@@ -0,0 +1,144 @@
...
+
+
+Parameter initialization of a very big embedding table is lazy. For example, in online learning, there could be unkown item id in the training data. So, until worker send the unkown item id to pserver, will pserver initialize corresponding embedding vector and send back to worker. This is a `get_or_create` semantic.
+
+Other parameters could be initialized before training.
+
+
+## Workflow
+
+### Assumption
```
wangkuiyi: Assumption => Distribution Strategy

doc/parameter_server.md
```diff
@@ -0,0 +1,144 @@
...
+- master, calling Kubernetes API to start workers and optionally parameter servers, and scheduling tasks to workers
+- pserver, maintaining the global model, and providing parameter pull/push/optimize/checkpoint service
+- worker, computing gradients of model parameters using local data and pushing gradients to pservers
+
+
+Please refer to:
+
+![parameter_server](./images/parameter_server.png)
+
+## Key concepts
```
wangkuiyi: I don't think we need `## Key concepts`

doc/parameter_server.md
```diff
@@ -0,0 +1,144 @@
...
+
+
+### Parameter sharding
```
wangkuiyi: Parameter sharding => Parameter Sharding

doc/parameter_server.md
```diff
@@ -0,0 +1,144 @@
...
+
+### Parameter Initialization
+
+
+Parameter initialization of a very big embedding table is lazy. For example, in online learning, there could be unkown item id in the training data. So, until worker send the unkown item id to pserver, will pserver initialize corresponding embedding vector and send back to worker. This is a `get_or_create` semantic.
+
+Other parameters could be initialized before training.
+
+
+## Workflow
```
wangkuiyi: I don't think we need `## Workflow`.

doc/parameter_server.md
```diff
@@ -0,0 +1,144 @@
...
+- only support worker failover, do not support pserver failover
+
+
+**Note**
+
+Allreduce failover and pserver failover will be discussed on other design docs.
+
+### Master
+
+- responsible for creating/deleting/scheduling pserver and worker
```
wangkuiyi: These clauses don't have the subject. We need the subject, the predicate, and the object to form a complete sentence.

doc/parameter_server.md
```diff
@@ -0,0 +1,144 @@
...
+- define parameter sharding strategy, and generate unique key for each parameter(or sharded parameter)
+- define the hash strategy from parameter key to pserver id 
+
+Workflow:
+
+
```
wangkuiyi, line 84: Again, `Workflow:` here is not a valid subject. And these are not valid sentences.

doc/parameter_server.md
```diff
@@ -0,0 +1,144 @@
+# Parameter Server Design
+
+## Overview
+
+A typical parameter server(pserver) architecture contains three roles:
```
wangkuiyi: We need an extra space here: `parameter server(pserver)` => `parameter server (pserver)`.

wangkuiyi (commented):
I have some suggestions about English grammar and lexical corrections.

doc/parameter_server.md
QiJune: Done

doc/parameter_server.md
QiJune: Done

doc/parameter_server.md
QiJune: Done

doc/parameter_server.md
QiJune: Done

doc/parameter_server.md
QiJune: Done

doc/parameter_server.md
QiJune: Done

doc/parameter_server.md
QiJune: Done

doc/parameter_server.md
QiJune: Done

doc/parameter_server.md
QiJune: Done

doc/parameter_server.md
```diff
@@ -0,0 +1,198 @@
...
+- Master is responsible for creating/deleting/scheduling pserver and worker
+- Master defines a parameter sharding strategy, and generate unique key for each parameter(or sharded parameter)
+- Master defines a hash strategy from parameter key to pserver id 
+
+Workflow:
+
+
+1. Master creates several pservers and workers according to user's configuration
+2. Master generates parameter sharding strategy based on model definition provided by user
+3. Master initializes parameter, and send to pservers
```
skydoorkai, line 87: Master tells pservers how to initialize the parameter, and pservers do the initialization.

doc/parameter_server.md
```diff
@@ -0,0 +1,198 @@
...
+
+Since our target is to scheduling deep learning jobs elasticly, we must have the ability to increase/decrease CPU resources of worker, and increase/decrease network bandwidth resources of pserver.
+
+### Solution
+
+As we are using asynchronous SGD, increasing/decreasing CPU resources of worker is trivial. We can just create more workers or kill some workers under different cluster status.
+
+And for pserver, how can we adjust network bandwidth? One solution is to create or delete pserver nodes. Since each node has a network interface card, more pserver nodes means more network bandwidth.
+
+The second solution is to adjust network bandwidth of current pserver node. We can create many pserver nodes first, but set network bandwidth limit to certain medium value. If we want to increase/decrease network bandwith, we increase/decrease the network bandwidth limit.
```
skydoorkai: ElasticDL should not change network bandwidth. 
Network bandwidth is very complicated in a data center network environment. For different k8s pods, they may reside in a physical machine, and share PCIe and network cards bandwidth. Or in different machines connected by routers/switches. From pods, there is no way to tell what network resources are shared and if the corresponding network bandwidth already exceeds its limit or not. Decrease a pod's network bandwidth may have no effect on any other ElasticDL pods, but decrease performance of itself.

QiJune: Yes, network environment could be complicated in a data center.
In an EDL job, workers need CPU most, pservers need network bandwidth most. Assigning CPU resources to workers is trivial, assigning network bandwidth resources to pservers is hard.

So, here comes some questions:

- Do we need to schedule pservers elastically? Or we just schedule workers.
- How to define increasing/decreasing resources to pservers? Is increasing/decreasing pserver pod enough?
- If we have to schedule pservers, how could we allocate the network bandwidth resources to different EDL jobs? 
- If EDL jobs are running with other jobs, how could we distribute network bandwidth to EDL jobs and other jobs? We know assigning CPU resources is trivial.

doc/parameter_server.md
QiJune: This are two choices:
- A big tensor with gaussian initializer,
- A big tensor are split into 10 subtensors, and each subtensor with gaussian initializer

These two choices will generate different value. 

But it seems that it will not influence the training accuracy. I think both are fine.

doc/parameter_server.md
```diff
@@ -0,0 +1,229 @@
...
+### Analysis
+
+First, we analyse the hardware resources that pserver and worker consume.
+
+
+| role| CPU | memory | network bandwidth |  disk |
+| :----  |:----  |:---|:---|:---|
+| worker  | high| medium |  medium | low|
+| pserver | low | medium | high| medium |
+
```
skydoorkai, line 152: Different models may result in different hw resource requirements.
Worker memory depends on the model and the minibatch size, may be very high. For example, MaskRCNN model can only support `minibatch=2` for a GPU with 16G memory in TensorFlow with default setting.

doc/parameter_server.md
```diff
@@ -0,0 +1,229 @@
...
+Worker does the forward/backward computation and consumes CPU most, while pserver provides kvstore service and consumes network bandwidth most.
+
+Since our target is to scheduling deep learning jobs elasticly, we must have the ability to increase/decrease CPU resources of worker, and increase/decrease network bandwidth resources of pserver.
+
+### Solution
+
+As we are using asynchronous SGD, increasing/decreasing CPU resources of worker is trivial. We can just create more workers or kill some workers under different cluster status.
+
+For example, if job A requests 50 CPU cores, but current cluster only has 30 CPU cores. The cluster will lanuch the job immediately with 30 CPU cores. If the cluster has more CPU cores after a while, it will tell the master of job A to occupy the idle CPU cores and create more workers.
```
skydoorkai, line 160: Current ElasticDL lets user set an upper limit of the number of workers and tries to launch that number of worker pods by k8s APIs.

doc/parameter_server.md
```diff
@@ -0,0 +1,229 @@
...
+
+And for pserver, how can we adjust network bandwidth? One solution is to create or delete pserver pods. Since each node has a network interface card, more pserver pods means more network bandwidth.
```
skydoorkai: network card != network bandwidth.  Multiple network cards may share other network components. If other components are bottlenecks, there would be no more bandwidth.

doc/parameter_server.md
```diff
@@ -0,0 +1,229 @@
...
+
+The third solution consumes the same memory, a little more CPU comparing to the former solutions, but avoids complex parameter re-sharding/consistent hashing under varying pserver pods. However, it's hard to manage network bandwidth resources in a data center. There could be a complex network topology. 
+
+The forth solution is to create many pserver pods and keep the pserver pod number unchanged. Each parameter sharding will be mapped to a pserver pod. But the location of pserver pods could be changed.
+
+In general, one team's job has high prority on his own resources. For example, team A has a job A which require 3 pserver nodes, but team A only has two nodes for pserver. So it will "borrow" one from other team, and schedules pserver pods to 3 nodes, 2 of them is his own, 1 is from another team.
+
+It the other team launch his own job, it will tell job A to release the "borrowed" node. So the pserver pods of job A will be moved to his own 2 nodes.
+
```
skydoorkai, line 187: The main issue for elastic ps is not the bandwidth, but the memory size. If the number of ps nodes is reduced, the ps pods may have enough memory to hold the re-sharded model variables.

doc/parameter_server.md
```diff
@@ -0,0 +1,229 @@
...
+- Master defines a parameter sharding strategy, and generate unique key for each parameter(or sharded parameter)
+- Master defines a hash strategy from parameter key to pserver id 
+
+Workflow:
+
+
+1. Master creates several pservers and workers according to user's configuration
+2. Master generates parameter sharding strategy based on model definition provided by user
+3. Master initializes parameter, and send to pservers
+4. Master triggers pservers and workers and start training
```
chunyang-wen: Typo and grammar errors in items 4 and 5.

doc/parameter_server.md
```diff
@@ -0,0 +1,229 @@
...
+1. Pserver waits for workers pushing <key, gradient> to pserver
+2. Pserver gets the key and queries kv store to get corresponding parameter
+3. Pserver calls optimizer, applys <key, gradient> to <key, parameter>
+4. Pserver updates <key, parameter> back to kv store
+5. Pserver saves model checkpoint to disk periodially
+
+
+**Note**
+
+We could implement the kv store by ourselves, or we could use some already solution, such as redis.
```
chunyang-wen: already->existing

doc/parameter_server.md
```diff
@@ -0,0 +1,229 @@
...
+
+Workflow:
+
+1. Before forward layer starting, worker sends key to pserver, and pull <key, parameter> back from pserver
+2. Worker starts forward layer computation
+3. After backward layer finishing, worker generates <key, gradient>
+4. Worker pushes <key, gradient> to pserver
+
+
+There is also another kind of worker who does an evalution job. It define its own evalution computation process and dataloader of validation data. It pulls <key, parameter> from pserver, but never push data to pserver.
```
chunyang-wen: Currently, ElasticDL's worker treats evaluation tasks and training tasks equally. They just request tasks and process tasks. If we add a single validation node, the evaluation speed may suffer.

doc/parameter_server.md
```diff
@@ -0,0 +1,229 @@
...
+
+
+**Note**
+
+There could be some same item id in a minibatch data. So some gradient vector of embedding table will have the same item id. We need to sum these gradient before pushing to pserver.
+
+
+## Delayed model updating
+
+In order to reduce the communication overhead between workers and pservers, we propose a stragety called delayed model updating.
```
chunyang-wen: **stragety**

doc/parameter_server.md
```diff
@@ -0,0 +1,229 @@
...
+
+A worker runs a round of forward/backward computation using its local model, and keep the gradients in its local. After several rounds finishing, it push gradients to pservers.
```
chunyang-wen: **in its local**?
**rounds finishing**?

doc/parameter_server.md
QiJune: Exactly, here we are focusing on recommendation scenario, which big embedding table is stored at pserver, and worker only has a partial model.  So the memory is `relative medium`.

doc/parameter_server.md
QiJune: I mean ElasticDL should also have the ability to create more workers even the job starting for a while,  if hardware resources are added.

So, `someone` is needed to tell master to create more workers, the `someone` could be the cluster scheduler.

doc/parameter_server.md
QiJune: Yes, ElasticDL has no ability to control network bandwidth, or even get the current network status. A naive strategy for us is to create pserver pods in different nodes. I will update this sentence.

doc/parameter_server.md
QiJune: Yes, if memory is not enough, ps will be crashed. If bandwidth is not enough, psever will be slow. I will update this.

doc/parameter_server.md
QiJune: Done

doc/parameter_server.md
QiJune: Done

doc/parameter_server.md
QiJune: Done

doc/parameter_server.md
QiJune: Done

doc/parameter_server.md
QiJune: Done

wangkuiyi (approved):
This PR has been here for a while. I am giving this approval so @QiJune has the option to merge it, and polish it later.

PR #1182, opened 2019-09-15
[Clean Code]move lookup_embedding from worker to embedding layer

QiJune:
Lookup embedding is a method of embedding layer. This PR aims to make code of customized layer, such as embedding layer,  follow the principle `low coupling and high cohesion`.

### DIFF
--- a/elasticdl/python/elasticdl/layers/embedding.py
+++ b/elasticdl/python/elasticdl/layers/embedding.py
@@ -89,14 +89,15 @@ class Embedding(tf.keras.layers.Layer):
     def get_key(name_list):
         return "-".join(map(str, name_list))
 
-    def lookup_embedding(self, unique_ids, embedding_service_endpoint):
+    def lookup_embedding(self, unique_ids):
         ids = unique_ids.numpy()
         keys = [Embedding.get_key([self._name, id]) for id in ids]
         (
             embedding_vectors,
             unknown_keys_index,
         ) = EmbeddingService.lookup_embedding(
-            keys=keys, embedding_service_endpoint=embedding_service_endpoint
+            keys=keys,
+            embedding_service_endpoint=self.embedding_service_endpoint,
         )
 
         if unknown_keys_index:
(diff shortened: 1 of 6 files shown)


elasticdl/python/worker/worker.py
```diff
@@ -126,10 +124,7 @@ def _init_embedding_layer(self):
         """
         self._embedding_layers = find_layer(self._model, Embedding)
         for layer in self._embedding_layers:
-            layer.set_lookup_func(self.lookup_embedding)
-        if self._embedding_layers:
-            # TODO check that Redis IP/PORT is set
```
mhaoli, line 131: Perhaps do not remove this TODO?

elasticdl/python/elasticdl/layers/embedding.py
```diff
@@ -44,6 +48,7 @@ def __init__(
         self.supports_masking = mask_zero
         self.input_length = input_length
         self.combiner = combiner
+        self.embedding_service_endpoint = (embedding_service_endpoint,)
```
mhaoli: Why do we need to convert `embedding_service_endpoint` to a tuple?

elasticdl/python/tests/worker_test.py
```diff
@@ -160,53 +131,6 @@ def test_embedding_layer(self):
         )
         self.assertTrue(len(worker._embedding_layers) == 2)
 
-    def test_lookup_embedding(self):
```
mhaoli, line 163: Formerly, this test case is for the correctness of `worker.lookup_embedding`. Maybe do not delete it but moving it to `layer_test.py`?

elasticdl/python/worker/worker.py
QiJune: The new logic is that

```python
for layer in self._embedding_layers:
    layer.set_endpoint(self._embedding_service_endpoint)
```
It set redis IP/Port to embedding layer directly.

elasticdl/python/tests/worker_test.py
QiJune: It seems that former code compares the same thing:

```python
e_lookup, e_unknown = EmbeddingService.lookup_embedding(
   keys=["-".join([layer_name, str(id)]) for id in ids]
)
lookup_result = worker.lookup_embedding(
   ids=ids,
   layer_name=layer_name,
   embedding_table_dim=embedding_table_dim,
)
```

`worker.lookup_embedding` is a thin wrapper of `EmbeddingService.lookup_embedding`. It directly calls `EmbeddingService.lookup_embedding`. I think it's no need to make an unittest for these two functions.

elasticdl/python/elasticdl/layers/embedding.py
QiJune: Done, remove it.

mhaoli (approved):

PR #1185, opened 2019-09-16
[Clean Code]remove init parameters logic from worker

QiJune:
1.   The logic that workers initialize parameters and push to pserver is not correct.  

There are many workers, the parameters could be initialized many times and also pushed to pservers many times.

The correct logic is to initialize parameters in master node, and then broadcast to pservers.  Workers only pull parameters from pservers.

2.  I made an experiment:

```python
def custom_model():
    inputs = tf.keras.Input(shape=(28, 28), name="image")
    x = tf.keras.layers.Reshape((28, 28, 1))(inputs)
    x = tf.keras.layers.Conv2D(32, kernel_size=(3, 3), activation="relu")(x)
    x = tf.keras.layers.Conv2D(64, kernel_size=(3, 3), activation="relu")(x)
    x = tf.keras.layers.BatchNormalization()(x)
    x = tf.keras.layers.MaxPooling2D(pool_size=(2, 2))(x)
    x = tf.keras.layers.Dropout(0.25)(x)
    x = tf.keras.layers.Flatten()(x)
    outputs = tf.keras.layers.Dense(10)(x)

    return tf.keras.Model(inputs=inputs, outputs=outputs, name="mnist_model")

model = custom_model()
print(model.trainable_variables)
```

It seems that parameters have been already created when creating the model.

3. Another experiment for subclass API

```python
class CustomModel(tf.keras.Model):
    def __init__(self, channel_last=True):
        super(CustomModel, self).__init__(name="mnist_model")
        if channel_last:
            self._reshape = tf.keras.layers.Reshape((28, 28, 1))
        else:
            self._reshape = tf.keras.layers.Reshape((1, 28, 28))
        self._conv1 = tf.keras.layers.Conv2D(
            32, kernel_size=(3, 3), activation="relu"
        )
        self._conv2 = tf.keras.layers.Conv2D(
            64, kernel_size=(3, 3), activation="relu"
        )
        self._batch_norm = tf.keras.layers.BatchNormalization()
        self._maxpooling = tf.keras.layers.MaxPooling2D(pool_size=(2, 2))
        self._dropout = tf.keras.layers.Dropout(0.25)
        self._flatten = tf.keras.layers.Flatten()
        self._dense = tf.keras.layers.Dense(10)

    def call(self, input, training=False):
        x = self._reshape(input)
        x = self._conv1(x)
        x = self._conv2(x)
        x = self._batch_norm(x, training=training)
        x = self._maxpooling(x)
        if training:
            x = self._dropout(x, training=training)
        x = self._flatten(x)
        x = self._dense(x)
        return x

model = CustomModel()
model.build(input_shape=(None, 28, 28))
print(model.trainable_variables)
```

### DIFF
--- a/elasticdl/python/master/main.py
+++ b/elasticdl/python/master/main.py
@@ -112,8 +112,8 @@ def main():
             "get_input_shape", args.model_zoo, model_module, required=True
         )
         input_shape = get_input_shape()
-        input_shape = [None] + input_shape
-        model_inst.build(input_shape)
+        input_shape = tuple([None] + input_shape)
+        model_inst.build(input_shape=input_shape)
 
     optimizer = model_module[args.optimizer]()
 
--- a/elasticdl/python/master/main.py
+++ b/elasticdl/python/master/main.py
@@ -17,6 +17,7 @@ from elasticdl.python.common.constants import (
 from elasticdl.python.common.k8s_tensorboard_client import TensorBoardClient
 from elasticdl.python.common.log_util import get_logger
 from elasticdl.python.common.model_helper import (
+    _get_spec_value,
     find_layer,
     get_module_file_path,
     load_model_from_module,
@@ -106,6 +107,14 @@ def main():
     model_inst = load_model_from_module(
         args.model_def, model_module, args.model_params
     )
+    if not model_inst.trainable_variables:
+        get_input_shape = _get_spec_value(
+            "get_input_shape", args.model_zoo, model_module, required=True
+        )
+        input_shape = get_input_shape()
+        input_shape = [None] + input_shape
+        model_inst.build(input_shape)
+
     optimizer = model_module[args.optimizer]()
 
     if all(
--- a/elasticdl/python/worker/worker.py
+++ b/elasticdl/python/worker/worker.py
@@ -113,12 +113,9 @@ class Worker(object):
         """Set model instance to worker."""
         self._model = model_inst
         self._init_embedding_layer()
-        self._var_created = self._model.built
-        self._non_embed_vars = []
-        if self._var_created:
-            self._non_embed_vars = get_non_embedding_trainable_vars(
-                self._model, self._embedding_layers
-            )
+        self._non_embed_vars = get_non_embedding_trainable_vars(
+            self._model, self._embedding_layers
+        )
 
     def _init_embedding_layer(self):
         """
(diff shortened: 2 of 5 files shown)


skydoorkai:
Try a keras subclass model, which requires that the worker reports the. variables.

QiJune:
@skydoorkai Okay.. Let me see how to support parameter initialization of Keras subclass model in master.

model_zoo/mnist_subclass/mnist_subclass.py
```diff
@@ -48,21 +48,24 @@ def optimizer(lr=0.01):
...
             }
         else:
             feature_description = {
-                "image": tf.io.FixedLenFeature([28, 28], tf.float32),
+                "image": tf.io.FixedLenFeature(get_input_shape(), tf.float32),
                 "label": tf.io.FixedLenFeature([1], tf.int64),
             }
         r = tf.io.parse_single_example(record, feature_description)
-        features = {
-            "image": tf.math.divide(tf.cast(r["image"], tf.float32), 255.0)
```
QiJune, line 64: Not very sure, why we use a dict here? I remove it because `tf.keras.Model.build` API only takes Tensor as input

elasticdl/python/master/main.py
```diff
@@ -106,6 +107,14 @@ def main():
     model_inst = load_model_from_module(
         args.model_def, model_module, args.model_params
     )
+    if not model_inst.trainable_variables:
```
QiJune, line 110: Function API initializes parameters at once. And subclass API will call `build` method to initialize parameters.

elasticdl/python/worker/worker.py
```diff
@@ -268,15 +256,6 @@ def report_prediction_outputs(self, predictions):
...
 
-    def _create_variable_and_report(self, features):
-        # Use model.call to create variables, then report to ps
-        _ = self._model.call(features)
-        self._non_embed_vars = get_non_embedding_trainable_vars(
-            self._model, self._embedding_layers
-        )
-        self.report_variable()
-        self._var_created = True
-
```
mhaoli, line 279: We can't delete `_create_variable_and_report `. Worker should call it (or explicitly build model to create variables) even if it does not need to report variables to master. Without creating trainable variables, model does not have any variable when function `_process_minibatch ` calls function `get_model`, i.e. `model.trainable_variables` is an empty list. `get_model` will do nothing if `model.trainable_variables`.

skydoorkai:
We used to use model. build to create subclass variables, but that requires users provide with input shape.  Then we switched to current design to avoid the extra burden. 

also,  some tf new layers in tf2.0 have not supported  build() call yet.

QiJune:
@skydoorkai @LiMinghao1994  Emm..  It seems that it's more complicated than I thought, after I go through #677 . Let us check if there exits any elegant solution.

mhaoli:
> @skydoorkai @LiMinghao1994 Emm.. It seems that it's more complicated than I thought, after I go through #677 . Let us check if there exits any elegant solution.

Shall we close this PR? We may open another PR or reopen this PR if we find a better solution.

QiJune:
Close this PR now and will reopen it if find some elegant solution.

PR #1187, opened 2019-09-16
Fix #1166 organize args

### LINKED ISSUE #1166: Reorganize arguments of common/master/worker to avoid duplicate declaration
Currently the `argparse.ArgumentParser` is created in different places. It is a little hard if we want to add arguments from the master and then pass them to the worker.

+ `common/args.py`
+ `master/args.py`
+ `worker/main.py`

It is better that we can organize them according to rules. We can organize them by `training`, `evaluation` and `prediction`.

chunyang-wen:
This PR reorganizes the arguments of the master and the workers.

+ Move all the argument declarations into a single file.
+ Move share arguments of the master and the workers into a function.
  + This avoids duplicate declarations.


elasticdl/python/common/args.py
```diff
@@ -84,12 +86,7 @@ def parse_envs(arg):
 
 
 def add_common_params(parser):
```
terrytangyuan, line 88: Since we have "add_common_args_between_master_and_worker()" now, should we rename this too?

elasticdl/python/common/args.py
```diff
@@ -383,3 +329,146 @@ def _get_attr(instance, attribute):
...
+    if args.prediction_data_dir and not args.checkpoint_filename_for_init:
+        raise ValueError(
+            "checkpoint_filename_for_init is required for running "
+            "prediction job"
+        )
+
+    return args
+
+
+def parse_worker_args(worker_args=None):
```
terrytangyuan, line 456: I am not sure if it makes sense to put this here since it's common/args.py and these are worker specific.

elasticdl/python/common/args.py
chunyang-wen: I am not aware of the meaning of common arguments here. As `add_common_args` is used by your client, I do not modify it here.

elasticdl/python/common/args.py
chunyang-wen: You can assume that `common/args.py` is used to manage arguments of ElasticDL. It seems better to place similar functions in the same place to avoid duplication.

elasticdl/python/common/args.py
chunyang-wen: Does `add_common_args` mean the common arguments of training, prediction and evaluation? If so, we can use a more meaningful name then.

terrytangyuan: Yes I think so. They are shared for training, prediction and evaluation.

elasticdl/python/common/args.py
terrytangyuan: Okay

elasticdl/python/common/args.py
chunyang-wen: I'd rather not change the name. Do you have any suggestions for the replacement? Changing the name to `add_common_params_for_training_prediction_evaluation` seems cumbersome. I have added a comment for the function.

terrytangyuan: Where is the comment?

chunyang-wen: Oops! Pushed to the wrong branch. Updated, please have a look.

terrytangyuan (approved):

PR #1189, opened 2019-09-16
Support tf.function when model has ElasticDL embedding layer

### LINKED ISSUE #1188: Support tf.function when model has ElasticDL embedding layer
Currently, ElasticDL embedding layer looks up embedding vectors from Redis and use `tf.GradientTape.watch(embedding_vectors)` to tell gradient tape to record gradients of embedding vectors. However, in a function decorated by `tf.function`, `tape.watch` does not work.

We can use trainable variables instead. In a function decorated by `tf.function`, TensorFlow supports assigning values to variables, and gradient tape can automatically record gradients of trainable variables. So, we can create a TensorFlow variable called `batch_embedding` in advance. Every time worker does forward-pass, embedding layer assigns the embedding vectors to the `batch_embedding`.

One detail is worthy mention. Currently, ElasticDL supports the situation that an embedding layer is called more than once in one forward-pass. Please refer to this model as an example:

https://github.com/sql-machine-learning/elasticdl/blob/db0a1de92f3550878745feaea90c2f9fa0a2454b/elasticdl/python/tests/embedding_test_module.py#L7-L41

However, it is not simple to support this feature under `tf.function` because we plan to create TensorFlow variable in advance. We don't think this is an important feature. Thus we plan to not use `tf.function` if an embedding layer is called more than once.

mhaoli:
Fixes #1188.

This PR includes following:
* As #1188 states, create trainable variables in ElasticDL embedding layer.
* In `layer_test.py`, existing test cases only test embedding layer in eager mode. Modifying these test cases to test embedding layer both in eager mode and in graph mode (i.e. without `tf.function` and with `tf.function`).
* In `worker_test.py`, add a test case to test that embedding layer behaves the same when runs with  `tf.function` and without `tf.function`.

### DIFF
--- a/elasticdl/python/elasticdl/layers/embedding.py
+++ b/elasticdl/python/elasticdl/layers/embedding.py
@@ -57,19 +57,17 @@ class Embedding(tf.keras.layers.Layer):
         # BET's shape and ids' shape in `self._bet_ids_pair_graph` have `None`
         # dimension. This is because they have different shapes in different
         # iterations.
-        # `tf.Variable` requires initial value if shape has `None` dimension
-        bet_initial_value = tf.zeros((1, self.output_dim))
-        ids_initial_value = tf.zeros((1, 1), dtype=tf.int64)
+        # `tf.Variable` requires initial value if shape has `None` dimension.
         self._bet_ids_pair_graph = [
             (
                 tf.Variable(
-                    bet_initial_value,
+                    initial_value=tf.zeros((1, self.output_dim)),
                     shape=tf.TensorShape((None, self.output_dim)),
                     dtype=tf.float32,
                     trainable=True,
                 ),
                 tf.Variable(
-                    ids_initial_value,
+                    initial_value=tf.zeros((1, 1), dtype=tf.int64),
                     shape=tf.TensorShape(None),
                     dtype=tf.int64,
                     trainable=False,
@@ -185,7 +183,7 @@ class Embedding(tf.keras.layers.Layer):
                 self._bet_ids_pair_eagerly.append((batch_embedding, flat_ids))
                 batch_embedding = self._bet_ids_pair_eagerly[-1][0]
             else:
-                # In graph mode, assigning tensor to trainable variables is
+                # In graph mode, assigning tensors to trainable variables is
                 # allowed and tape can record the gradients of trainable
                 # variables automatically.
                 self._bet_ids_pair_graph[0][0].assign(batch_embedding)
@@ -221,7 +219,7 @@ class Embedding(tf.keras.layers.Layer):
                 )
                 batch_embedding = self._bet_ids_pair_eagerly[-1][0]
             else:
-                # In graph mode, assigning tensor to trainable variables is
+                # In graph mode, assigning tensors to trainable variables is
                 # allowed and tape can record the gradients of trainable
                 # variables automatically.
                 self._bet_ids_pair_graph[0][0].assign(batch_embedding)
--- a/elasticdl/python/elasticdl/layers/embedding.py
+++ b/elasticdl/python/elasticdl/layers/embedding.py
@@ -1,8 +1,12 @@
+import collections
 import numpy as np
 import tensorflow as tf
 from tensorflow.python.keras.utils import tf_utils
 
 from elasticdl.python.master.embedding_service import EmbeddingService
+
+
+EmbeddingAndIds = collections.namedtuple('EmbeddingAndIds', ['batch_embedding', 'batch_ids'])
 
 
 class Embedding(tf.keras.layers.Layer):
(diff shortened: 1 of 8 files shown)


elasticdl/python/elasticdl/layers/embedding.py
```diff
@@ -148,19 +172,27 @@ def call(self, input):
...
+                self.tape.watch(batch_embedding)
+                self._bet_ids_pair_eagerly.append((batch_embedding, flat_ids))
+                batch_embedding = self._bet_ids_pair_eagerly[-1][0]
+            else:
+                # In graph mode, assigning tensor to trainable variables is
+                # allowed and tape can record the gradients of trainable
+                # variables automatically.
+                self._bet_ids_pair_graph[0][0].assign(batch_embedding)
+                self._bet_ids_pair_graph[0][1].assign(flat_ids)
+                batch_embedding = self._bet_ids_pair_graph[0][0]
```
terrytangyuan: It's a little hard to read these indices in the above code, e.g. `[-1][0]`. Consider assigning them as variables so they are more meaningful.

elasticdl/python/elasticdl/layers/embedding.py
```diff
@@ -176,27 +208,43 @@ def _sparse_input_call(self, sparse_input):
...
+                    (batch_embedding, unique_ids)
+                )
+                batch_embedding = self._bet_ids_pair_eagerly[-1][0]
+            else:
+                # In graph mode, assigning tensor to trainable variables is
+                # allowed and tape can record the gradients of trainable
+                # variables automatically.
+                self._bet_ids_pair_graph[0][0].assign(batch_embedding)
+                self._bet_ids_pair_graph[0][1].assign(unique_ids)
+                batch_embedding = self._bet_ids_pair_graph[0][0]
```
terrytangyuan: Repeated code?

elasticdl/python/tests/example_test.py
```diff
@@ -124,7 +124,7 @@ def distributed_train_and_evaluate(
         )
 
         if dataset == "imagenet":
-            shards = {create_imagenet_recordio_file(16, feature_shape): 16}
+            shards = {create_imagenet_recordio_file(8, feature_shape): 8}
```
terrytangyuan, line 128: Why this change?

elasticdl/python/tests/layer_test.py
```diff
@@ -145,11 +142,19 @@ def test_embedding_layer(self):
...
+                lambda: module_call(model, inputs),
+            ]
+            for call_fn in call_fns:
+                outputs = call_fn()
+                values = outputs.numpy()[1]
+                for index, idx in enumerate(ids):
+                    correct_value = np.array(
+                        [idx] * output_dim, dtype=np.float32
+                    )
+                    self.assertTrue((values[index] == correct_value).all())
```
terrytangyuan: Repeated code?

elasticdl/python/tests/layer_test.py
```diff
@@ -160,11 +165,13 @@ def test_embedding_layer_with_input_length(self):
...
-        for index, idx in enumerate(flatten_ids):
-            correct_value = np.array([idx] * output_dim, dtype=np.float32)
-            self.assertTrue((values[index] == correct_value).all())
+        call_fns = [lambda: layer.call(ids), lambda: module_call(layer, ids)]
+        for call_fn in call_fns:
+            values = call_fn()
+            values = values.numpy().reshape(-1, output_dim)
+            for index, idx in enumerate(flatten_ids):
+                correct_value = np.array([idx] * output_dim, dtype=np.float32)
+                self.assertTrue((values[index] == correct_value).all())
```
terrytangyuan: Same here and also other places (double check though)

elasticdl/python/tests/example_test.py
mhaoli: Because  the default memory limit of docker is 2G. Running ResNet50 with batch size 16 uses nearly 2G memory and sometimes it fails because of OOM.

elasticdl/python/elasticdl/layers/embedding.py
mhaoli: How about using named tuple?

terrytangyuan: Sounds good to me

elasticdl/python/tests/example_test.py
terrytangyuan: Got it.

skydoorkai (approved):

elasticdl/python/elasticdl/layers/embedding.py
mhaoli: Done.

elasticdl/python/tests/layer_test.py
mhaoli: Remove most of repeat code. Leave two test cases unchanged. Because removing a small quantity of repeat code in these two function requires changing the logic of test case itself. (`test_embedding_layer_gradient` and `test_embedding_layer_gradient_with_sparse_inputs`)

PR #1200, opened 2019-09-17
do not use lock for async SGD in ReportGradient

### LINKED ISSUE #1167: Do not use lock in ReportGradient if async

skydoorkai:
fix #1167

### DIFF
--- a/elasticdl/python/master/servicer.py
+++ b/elasticdl/python/master/servicer.py
@@ -201,12 +201,6 @@ class MasterServicer(elasticdl_pb2_grpc.MasterServicer):
             )
 
     def _update_model(self):
-        if not self._use_async and not self._lock.locked():
-            # TODO (chengfu.wcy) `self._lock.locked` may be removed
-            # according to  changes in `ReportGradient` in async mode.
-            raise RuntimeError(
-                "Lock must be acquired when updating the model in sync mode"
-            )
         grad_var = []
 
         # (grad, var) pairs excluding keras Embedding layer and
@@ -229,11 +223,19 @@ class MasterServicer(elasticdl_pb2_grpc.MasterServicer):
 
         self._opt.apply_gradients(grad_var)
 
+        # need the lock for model version update in async SGD
+        if self._use_async:
+            self._lock.acquire()
         self._update_model_version()
-        self._gradient_sum.clear()
-        self._gradient_sum_indexed.clear()
-        self._edl_embedding_gradients.clear()
-        self._grad_n = 0
+        self._update_evaluation()
+        self._update_checkpoint()
+        if self._use_async:
+            self._lock.release()
+        else:
+            self._gradient_sum.clear()
+            self._gradient_sum_indexed.clear()
+            self._edl_embedding_gradients.clear()
+            self._grad_n = 0
 
     def get_model_version(self):
         return self._version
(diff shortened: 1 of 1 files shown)


elasticdl/python/master/servicer.py
```diff
@@ -229,11 +223,19 @@ def _update_model(self):
...
 
+        # need the lock for model version update in async SGD
+        if self._use_async:
+            self._lock.acquire()
         self._update_model_version()
-        self._gradient_sum.clear()
-        self._gradient_sum_indexed.clear()
-        self._edl_embedding_gradients.clear()
-        self._grad_n = 0
+        self._update_evaluation()
```
chunyang-wen, line 230: Can we wrap these clauses into a `with` statemennt?

elasticdl/python/master/servicer.py
```diff
@@ -401,16 +399,19 @@ def ReportGradient(self, request, _):
...
+            # TODO: do not accumulate gradients but apply directly.
+            pass
+
+        # staleness-aware learning rate modulation
+        if self._lr_modulation:
+            staleness = max(1, self._version - request.model_version)
+            self._lr_modulation.set_multiplier(1.0 / staleness)
+        if self._use_async or self._grad_n >= self._grad_to_wait:
+            self._update_model()
+        if not self._use_async:
```
chunyang-wen: `lock.acquire` and `lock.release` appears in multiple places. It is easy to mess things up.

elasticdl/python/master/servicer.py
ywskycn: May not easy to put `with` here, as the lock is only for `_use_async`.

elasticdl/python/master/servicer.py
ywskycn: One possible idea is to move some intermediate code outside to another func to make the acquire->func->release clearer.

skydoorkai: changed

elasticdl/python/master/servicer.py
```diff
@@ -229,11 +223,19 @@ def _update_model(self):
...
-        self._gradient_sum_indexed.clear()
-        self._edl_embedding_gradients.clear()
-        self._grad_n = 0
+        self._update_evaluation()
+        self._update_checkpoint()
+        if self._use_async:
+            self._lock.release()
+        else:
+            self._gradient_sum.clear()
+            self._gradient_sum_indexed.clear()
```
chunyang-wen, line 236:
```python
from contextlib import contextmanager
@contextmanager
def _possible_lock():
    if self._use_async:
        self._lock.acquire()

     yield
     if self._use_async:
         self._lock.release()

with _possible_lock():
  self._update_model_version()
  self._update_evaluation()
  self._update_checkpoint()

if not self._use_async:
  self._graidentxx
```

elasticdl/python/master/servicer.py
```diff
@@ -324,58 +326,68 @@ def ReportGradient(self, request, _):
...
+            else:
+                if arr.shape != self._model[k].numpy().shape:
+                    raise ValueError(
+                        "Gradient key: %s has incompatible dimension", k
+                    )
+                tmp[k] = arr
+
+        if not self._use_async:
+            self._lock.acquire()
+        self._process_gradients(
```
chunyang-wen, line 377: We can use a similar context as above comment.

chunyang-wen (approved):

PR #1201, opened 2019-09-17
Enable tests related to ODPS on Travis CI

terrytangyuan:
Changes include:
* Add `create_iris_odps_table()` to populate iris data in a temporary ODPS table. Note that this is through executing SQL statement intentionally since SQLFlow uses similar template that can be extended easily to test other SQL engines.
* Modify related tests to reflect the expected read result from iris ODPS table.
* Add secure Travis tokens for ODPS related environment variables (this is also being used in SQLFlow but I regenerated the tokens for uses in ElasticDL).


chunyang-wen:
Where is the data stored?  on internal servers?

chunyang-wen (approved):

terrytangyuan:
@chunyang-wen It’s in a public cloud.

Issue #1204, opened 2019-09-18
fix relative path of elasticdl-rbac.yaml in elasticdl_local.yaml


workingloong:
`cd elasticdl ` has been executed after git clone, so relative path of elasticdl-rbac.yaml in repo is elasticdl/manifests/examples/elasticdl-rbac.yaml. `kubectl apply`  should be exected after minikube start, otherwise timeout error will be raised.

PR #1205, opened 2019-09-18
modify relative path of elastic-rbac.yaml for kubectl apply

workingloong:
modify relative path of elastic-rbac.yaml for `kubectl apply` and execute `kubectl apply` after minikube start. #1204


QiJune (approved):
Cool!  LGTM

PR #1210, opened 2019-09-18
Automatically pass arguments

### LINKED ISSUE #1208: Change use_async and lr_staleness_modulation to boolean arguments and pass them
The arguments:
+ `use_async`
+ `lr_staleness_modulation`

are arguments with default value `False`. They are not boolean arguments.

And also pass them from client to master and worker

chunyang-wen:
Fixes #1208 

+ Support boolean arguments, which is more robust than default `store_true` or `store_false`
+ Automatically generate arguments for different jobs in `api.py`

### DIFF
--- a/elasticdl/python/elasticdl/api.py
+++ b/elasticdl/python/elasticdl/api.py
@@ -34,8 +34,9 @@ def train(args):
         _cluster_spec_def_in_docker(args.cluster_spec),
     ]
     container_args.extend(
-        build_arguments_from_parsed_result(args),
-        filter_args=["model_zoo", "cluster_spec", "worker_image"],
+        build_arguments_from_parsed_result(
+            args, filter_args=["model_zoo", "cluster_spec", "worker_image"]
+        )
     )
 
     _submit_job(image_name, args, container_args)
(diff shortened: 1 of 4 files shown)


elasticdl/python/elasticdl/api.py
```diff
@@ -23,70 +26,18 @@ def train(args):
...
-        args.model_def,
-        "--model_params",
-        args.model_params,
     ]
-    container_args.extend(["--image_pull_policy", args.image_pull_policy])
-    container_args.extend(["--restart_policy", args.restart_policy])
-    container_args.extend(["--volume", args.volume])
+    container_args.extend(
+        build_arguments_from_parsed_result(
+            args, filter_args=["model_zoo", "cluster_spec", "worker_image"]
```
skydoorkai, line 38: Why exclude model_zoo?

skydoorkai: Oh,  it is already set in container_args.

skydoorkai (approved):

terrytangyuan (approved):
Nice work!

PR #1212, opened 2019-09-18
Only run ODPS related tests if required env vars are set

terrytangyuan:
As pointed out in [Travis documentation](https://docs.travis-ci.com/user/encryption-keys/),  encrypted environment variables are not available for pull requests from forks. The only solution for now is to only run ODPS related tests if the environment variables are set (e.g. when PRs are created from branches instead of forks). [SQLFlow is currently using the same solution](https://github.com/sql-machine-learning/sqlflow/blob/develop/scripts/test_maxcompute.sh#L18).


QiJune (approved):
Great! LGTM

PR #1213, opened 2019-09-18
Add ODPSDataReader and relevant tests

terrytangyuan:
Part of #1105. Changes include:
* Added `ODPSDataReader` that implements `AbstractDataReader`.
* Added tests for existing subclasses of `AbstractDataReader`.
* Moved `create_iris_odps_table()` to a test utils module that can be reused across ODPS related tests.
* Renamed `_count_table_size()` to `get_table_size()` that can be reused in `ODPSDataReader`.


elasticdl/python/common/data_reader.py
```diff
@@ -1,9 +1,12 @@
+import math
 import os
 from abc import ABC, abstractmethod
 from contextlib import closing
 
 import recordio
 
+from elasticdl.python.common.odps_io import ODPSReader
+
```
chunyang-wen, line 8: should we create a module named `io` or  `reader` instead of putting them into `common` module?

elasticdl/python/common/data_reader.py
```diff
@@ -61,3 +64,45 @@ def create_shards(self):
             with closing(recordio.Index(p)) as rio:
                 f_records[p] = (start_ind, rio.num_records())
         return f_records
+
+
+class ODPSDataReader(AbstractDataReader):
+    def __init__(self, **kwargs):
+        AbstractDataReader.__init__(self, **kwargs)
+        self._kwargs = kwargs
```
chunyang-wen, line 70: Please add a validating method here and print useful information to users if any of the following required arguments is missing:
+ project
+ access_id
+ access_key
+ endpoint
+ num_processes
+ records_per_task

elasticdl/python/common/data_reader.py
```diff
@@ -61,3 +64,45 @@ def create_shards(self):
...
+        )
+        for batch in records:
+            yield batch
+
+    def create_shards(self):
+        shard_name_prefix = "shard_"
+        table_size = self._reader.get_table_size()
+        records_per_task = self._kwargs["records_per_task"]
+        shards = {}
+        num_shards = math.floor(table_size / records_per_task)
```
chunyang-wen:
```python
table_size // records_per_task
```

elasticdl/python/tests/odps_data_reader_test.py
```diff
@@ -0,0 +1,111 @@
...
+        }
+        self.assertEqual(expected_shards, reader.create_shards())
+
+        # Test records reading
+        records = list(reader.read_records(self.Task(0, 2, "shard_0")))
+        self.assertEqual(
+            [[6.4, 2.8, 5.6, 2.2, 2], [5.0, 2.3, 3.3, 1.0, 1]], records
+        )
+
+        # Teardown
```
chunyang-wen: Is it better we use `tearDown` from `TestCase`? I am afraid if any test fails, cleaning up operations will not be executed.

elasticdl/python/common/data_reader.py
```diff
@@ -61,3 +64,45 @@ def create_shards(self):
...
+        start_ind = 0
+        for shard_id in range(num_shards):
+            shards[shard_name_prefix + str(shard_id)] = (
+                start_ind,
+                records_per_task,
+            )
+            start_ind += records_per_task
+        num_records_left = table_size % records_per_task
```
chunyang-wen, line 110: I have done a similar work. FYI.
```python
import numpy as np
start = 0
table_size = 10
records_per_task = 4
shard_name_prefix = "shard_"
parts = np.arange(start, table_size, records_per_task).tolist()
parts_len = len(parts)
parts.append(table_size)
shards = dict(
    (shard_name_prefix+str(shard_id),
    (shard_id*records_per_task, parts[shard_id+1] - parts[shard_id])
    ) for shard_id in range(parts_len)
)

print(shards)
```

elasticdl/python/common/data_reader.py
```diff
@@ -61,3 +64,45 @@ def create_shards(self):
...
+        if num_records_left != 0:
+            shards[shard_name_prefix + str(num_shards)] = (
+                start_ind,
+                num_records_left,
+            )
+        return shards
```
chunyang-wen, line 116: Be careful that the size of a table can be a very large number. So `shards` may occupy a large part of the total memory.  It would be better that `shards` can be a generator. I have heard users complain to me that they cannot start their training job due to this shard generation process which causes OOM.

elasticdl/python/common/data_reader.py
terrytangyuan: Good suggestion. I'll rename them later together since this will cause conflicts with my other work-in-progress branch.

elasticdl/python/common/data_reader.py
terrytangyuan: Done.

elasticdl/python/common/data_reader.py
terrytangyuan: Thanks

elasticdl/python/tests/odps_data_reader_test.py
terrytangyuan: Done. Thanks.

elasticdl/python/common/data_reader.py
terrytangyuan: Good call. This could be done separately as it involves some changes in other places such as TaskDispatcher, etc.

chunyang-wen (approved):

PR #1214, opened 2019-09-19
Remove manually build args for worker

### LINKED ISSUE #1211: Remove manually create arguments in master/worker
In ElasticDL's master and worker. We manually create the arguments collection for them. It is a little hacky as we are adding more and more arguments.
After #1210 is merged, we can also update these two places where we manually add arguments.

chunyang-wen:
This PR fixes #1211 
+ Clear manually created arguments

Maybe we can discuss whether there exists any danger here.

### DIFF
--- a/elasticdl/python/master/main.py
+++ b/elasticdl/python/master/main.py
@@ -6,7 +6,11 @@ import grpc
 from kubernetes.client import V1EnvVar
 
 from elasticdl.proto import elasticdl_pb2_grpc
-from elasticdl.python.common.args import parse_envs, parse_master_args
+from elasticdl.python.common.args import (
+    build_arguments_from_parsed_result,
+    parse_envs,
+    parse_master_args,
+)
 from elasticdl.python.common.constants import (
     GRPC,
     JobType,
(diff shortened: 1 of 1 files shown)


elasticdl/python/master/main.py
```diff
@@ -227,31 +231,14 @@ def main():
...
-            args.dataset_fn,
-            "--loss",
-            args.loss,
-            "--optimizer",
-            args.optimizer,
-            "--eval_metrics_fn",
-            args.eval_metrics_fn,
-            "--model_def",
-            args.model_def,
             "--job_type",
```
terrytangyuan, line 236: Why these args are kept here?

elasticdl/python/master/main.py
```diff
@@ -6,7 +6,11 @@
 from kubernetes.client import V1EnvVar
```
terrytangyuan, line 6: Perhaps also change this for master in the same PR since the referenced issue is for fixing both of them?

elasticdl/python/master/main.py
chunyang-wen: They are not the arguments that parsed from command line.

terrytangyuan: Got it.

elasticdl/python/master/main.py
chunyang-wen: Do you mean the worker? The worker does not create argument collections such as `worker_args` or `container_args`.

terrytangyuan (approved):

PR #1217, opened 2019-09-19
Apply gradients directly for async

### LINKED ISSUE #1169: [async SGD] Do not copy edl_embedding_gradients/indexed_grads/tmp(other grads) to self._*, but use directly for update_model

skydoorkai:
fix #1169

### DIFF
--- a/elasticdl/python/master/servicer.py
+++ b/elasticdl/python/master/servicer.py
@@ -423,11 +423,14 @@ class MasterServicer(elasticdl_pb2_grpc.MasterServicer):
             indexed_grads = self._gradient_sum_indexed
             grads = self._gradient_sum
         if need_to_update_model:
-            if self._lr_modulation:
-                # staleness-aware learning rate modulation
-                staleness = max(1, self._version - request_version)
-                self._lr_modulation.set_multiplier(1.0 / staleness)
+            self._update_optimizer(request_version)
             self._update_model(grads, indexed_grads, edl_embedding_gradients)
+
+    def _update_optimizer(self, request_version):
+        if self._lr_modulation:
+            # staleness-aware learning rate modulation
+            staleness = max(1, self._version - request_version)
+            self._lr_modulation.set_multiplier(1.0 / staleness)
 
     def ReportTaskResult(self, request, _):
         if request.err_message:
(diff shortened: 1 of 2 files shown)


elasticdl/python/master/servicer.py
```diff
@@ -414,16 +410,23 @@ def _process_gradients(
...
-            # TODO: do not accumulate gradients but apply directly.
-            pass
 
         # staleness-aware learning rate modulation
         if self._lr_modulation:
             staleness = max(1, self._version - request_version)
             self._lr_modulation.set_multiplier(1.0 / staleness)
         if self._use_async or self._grad_n >= self._grad_to_wait:
-            self._update_model()
+            if not self._use_async:
```
terrytangyuan: The logic is a bit hard to read here since we have both:
```
if self._use_async or self._grad_n >= self._grad_to_wait:
```
and then 
```
if not self._use_async:
```
Even though it is valid by taking a closer look but it would be great if there's a good way to re-arrange the if conditions here.

skydoorkai: Revised.

elasticdl/python/master/servicer.py
```diff
@@ -414,16 +410,25 @@ def _process_gradients(
...
-            self._lr_modulation.set_multiplier(1.0 / staleness)
-        if self._use_async or self._grad_n >= self._grad_to_wait:
-            self._update_model()
+
+        need_to_update_model = self._use_async
+        if not self._use_async and self._grad_n >= self._grad_to_wait:
+            need_to_update_model = True
+            # get gradient average for sync SGD
+            for k in self._gradient_sum:
+                if not self._use_async:
```
chunyang-wen: Redundant check for `self._use_async`

skydoorkai: fixed.

elasticdl/python/master/servicer.py
```diff
@@ -414,16 +410,25 @@ def _process_gradients(
...
+                        self._gradient_sum[k] / self._grad_to_wait
+                    )
+            edl_embedding_gradients = self._edl_embedding_gradients
+            indexed_grads = self._gradient_sum_indexed
+            grads = self._gradient_sum
+        if need_to_update_model:
+            if self._lr_modulation:
+                # staleness-aware learning rate modulation
+                staleness = max(1, self._version - request_version)
+                self._lr_modulation.set_multiplier(1.0 / staleness)
```
chunyang-wen: `_lr_modulation` here is a little confusing. I have read the code. It takes its effect through the optimizer instance. Can we make a function such as `_udpate_optimizer` here to clarify the purpose?

chunyang-wen (approved):
LGTM

PR #1218, opened 2019-09-19
Added dummy model to test ODPSDataReader with local Keras

terrytangyuan:
Part of #1105, including:
* Added dummy model to test ODPSDataReader with local Keras.
* Fixed an issue with yielding data records when data is eager tensors with multiple values.

### DIFF
--- a/elasticdl/python/common/dataset_utils.py
+++ b/elasticdl/python/common/dataset_utils.py
@@ -17,7 +17,7 @@ def create_dataset_from_tasks(tasks):
         def gen(self):
             for task in self._tasks:
                 for data in self._data_reader.read_records(task):
-                    if data:
+                    if data is not None:
                         yield data
 
     generator = _Generator(tasks)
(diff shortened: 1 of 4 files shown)


elasticdl/python/common/dataset_utils.py
```diff
@@ -17,7 +17,7 @@ def __init__(self, tasks):
         def gen(self):
             for task in self._tasks:
                 for data in self._data_reader.read_records(task):
-                    if data:
+                    if data is not None:
```
chunyang-wen: Why?  I think `if data` is better.

terrytangyuan: I’ve seen issues like “ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()” today with some datasets. Similar to #1112. It’s safer to check this explicitly.

chunyang-wen: Why does `read_records` return `None`?

terrytangyuan: There was some mistake during implementation of new data reader. I just reverted this change since it’s unnecessary for now. Will add this in the future if really needed.

chunyang-wen (approved):

PR #1229, opened 2019-09-23
Support get model frequency

### LINKED ISSUE #1140: Worker supports get_model_frequency argument in async-SGD
According to [design doc](https://github.com/sql-machine-learning/elasticdl/blob/develop/elasticdl/doc/async_sgd_design.md#change-in-worker)

To support SSP strategy, the worker pulls the model from PS in every staleness_threshold minibatch step. Also, the worker needs to update the local model with the computed gradients. model pull/updates do not include embedding variables, as we directly access the embedding vectors in the embedding service.

mhaoli:
Fixes #1140.


elasticdl/python/common/args.py
```diff
@@ -465,6 +465,14 @@ def parse_master_args(master_args=None):
             "checkpoint_filename_for_init is required for running "
             "prediction job"
         )
+    if not args.use_async and args.get_model_steps > 1:
+        raise ValueError(
+            "get_model_steps should be 1 when using synchronous SGD."
+        )
+    if args.use_async and args.grads_to_wait > 1:
```
skydoorkai, line 473: The default value for `grads_to_wait` is 2. Should we change it to 1 so async does not need to set it ?

elasticdl/python/tests/in_process_master.py
```diff
@@ -15,9 +17,15 @@ def ReportVariable(self, req):
...
 
     def ReportGradient(self, req):
+        for callback in self._callbacks:
+            if test_call_back.ON_REPORT_GRADIENT_BEGIN in callback.call_times:
+                callback()
         return self._m.ReportGradient(req, None)
 
     def ReportEvaluationMetrics(self, req):
+        for callback in self._callbacks:
+            if test_call_back.ON_REPORT_GRADIENT_BEGIN in callback.call_times:
```
skydoorkai: should be `ON_REPORT_EVALUATION_METRICS_BEGIN`?

elasticdl/python/worker/worker.py
```diff
@@ -469,6 +504,22 @@ def run(self):
             mode = Mode.EVALUATION
         else:
             mode = Mode.TRAINING
+
+        # need to get model from PS, i.e. `train_with_local_model=False`
+        # at following situations:
```
skydoorkai: Make it clear:
The worker needs to get model from PS if `train_with_local_model=False`. 
This happens when:

elasticdl/python/worker/worker.py
```diff
@@ -469,6 +504,22 @@ def run(self):
...
+        #     local_update_count >= worker._get_model_steps
+        # otherwise, train with local model, i.e. `train_with_local_model=True`
+        train_with_local_model = False
+
+        local_update_count = 0
+        last_training_minibatch_failed = False
+        # initialize `last_minibatch_do_evaluation` to True in order to set
+        # `train_with_local_model` to False inside forloop for the first
+        # minibatch.
+        last_minibatch_do_evaluation = True
```
skydoorkai: Is it better to set:
`local_update_count == self._get_model_steps`  instead?

elasticdl/python/worker/worker.py
```diff
@@ -477,11 +528,36 @@ def run(self):
...
+                    last_minibatch_do_evaluation = False
+                    local_update_count += 1
+                    if err_msg:
+                        last_training_minibatch_failed = True
+                    else:
+                        last_training_minibatch_failed = False
+                        if local_update_count < self._get_model_steps:
+                            self._update_local_model()
+                elif task.type == elasticdl_pb2.EVALUATION:
+                    last_minibatch_do_evaluation = True
```
skydoorkai: This will be evaluation-only job, so no need to set here.

elasticdl/python/common/args.py
mhaoli: Good point. Done.

elasticdl/python/tests/in_process_master.py
mhaoli: Fixed.

elasticdl/python/worker/worker.py
mhaoli: Done.

elasticdl/python/worker/worker.py
mhaoli: Done.

elasticdl/python/worker/worker.py
mhaoli: Good point. Done.

elasticdl/python/common/args.py
```diff
@@ -465,6 +465,14 @@ def parse_master_args(master_args=None):
             "checkpoint_filename_for_init is required for running "
             "prediction job"
         )
+    if not args.use_async and args.get_model_steps > 1:
+        raise ValueError(
+            "get_model_steps should be 1 when using synchronous SGD."
```
chunyang-wen: We can ignore this argument and warn the users instead of crashing our program.

elasticdl/python/common/args.py
```diff
@@ -465,6 +465,14 @@ def parse_master_args(master_args=None):
...
+        )
+    if args.use_async and args.grads_to_wait > 1:
+        raise ValueError(
+            "grads_to_wait should be 1 when using asynchronous SGD."
+        )
```
chunyang-wen, line 477: The same as the above comment.

elasticdl/python/common/args.py
mhaoli: Done.

elasticdl/python/common/args.py
mhaoli: Done

chunyang-wen:
LGTM @skydoorkai Please double check the logic.

skydoorkai (approved):

PR #1245, opened 2019-09-26
Use num_minibatches_per_task instead of records_per_task as the ElasticDL command parameter.

### LINKED ISSUE #1232: Ensure records per task to be integral multiple of mini-batch size.
Currently, EDL train command contains two parameters: minibatch_size and records_per_task. Because there is no constraint between the value of these two parameters(such as records_per_task = 10, minibatch_size = 8), the records in one mini-batch may come from two tasks.
In order to make the process logic simple and clear, we make sure that records_per_task to be integral multiple of minibatch_size. But we can't assume that user will set these two parameter according to the constraints. Change the parameter name from records_per_task to minibatch_count_per_task.

brightcoder01:
Fix #1232


elasticdl/python/common/args.py
```diff
@@ -173,7 +173,12 @@ def add_common_params(parser):
         help="The name of the Kubernetes namespace where ElasticDL "
         "pods will be created",
     )
-    parser.add_argument("--records_per_task", type=int, required=True)
+    parser.add_argument(
+        "--minibatches_per_task",
+        type=int,
+        help="Minibatch number per task",
```
terrytangyuan: "Number of minibatches per task"

elasticdl/python/common/args.py
```diff
@@ -173,7 +173,12 @@ def add_common_params(parser):
         help="The name of the Kubernetes namespace where ElasticDL "
         "pods will be created",
     )
-    parser.add_argument("--records_per_task", type=int, required=True)
+    parser.add_argument(
+        "--minibatches_per_task",
```
terrytangyuan: Maybe `num_minibatches_per_task` would be better?

elasticdl/python/master/main.py
```diff
@@ -94,11 +94,14 @@ def main():
         args.evaluation_data_dir,
         args.prediction_data_dir,
     )
+
+    # Make sure that records_per_task is integral multiple of minibatch_size
```
skydoorkai: Rephrase this comment. No need to `make sure` as it is guaranteed as a multiple of minibatch_size from the calculation below.

brightcoder01: Good idea, will remove it.

elasticdl/python/common/args.py
brightcoder01: Good idea, updated.

elasticdl/python/common/args.py
brightcoder01: *minibatches_per_task* is using the same pattern with records_per_task. How about keep aligned with the existed parameters.

brightcoder01: Good idea, updated.

terrytangyuan:
@brightcoder01 It looks like @chunyang-wen has a question in https://github.com/sql-machine-learning/elasticdl/issues/1232#issuecomment-535529462. Could you respond to that?

brightcoder01:
> @brightcoder01 It looks like @chunyang-wen has a question in [#1232 (comment)](https://github.com/sql-machine-learning/elasticdl/issues/1232#issuecomment-535529462). Could you respond to that?

Thanks for the remind. Just replied.

skydoorkai (approved):

PR #1257, opened 2019-09-27
Allow to pass additional params to data reader constructor

### LINKED ISSUE #1256: Support passing additional parameters to data reader
For example, information such as the list of column names should be supported by `ODPSDataReader`.

terrytangyuan:
Fixes #1256. Changes include:
* Allow to pass additional params to data reader constructor. Tests are added for `create_data_reader()` that demonstrates this new feature.
* Add `--data_reader_params` in CLI and support this in `Worker`.
* Extract `get_dict_from_params_str()` that is reusable for parsing model params and data reader params. Changed to use semi-colon as the kv separator since comma is too common in Python. Tests are added to demonstrate this.

### DIFF
--- a/elasticdl/python/common/args.py
+++ b/elasticdl/python/common/args.py
@@ -412,6 +412,13 @@ def add_common_args_between_master_and_worker(parser):
         default=1,
         help="Worker will get_model from PS every these steps.",
     )
+    parser.add_argument(
+        "--data_reader_params",
+        type=str,
+        default="",
+        help="The dictionary of data reader parameters in a string that will be "
+             'used to instantiate the data reader, e.g. "columns=["a", "b"]"',
+    )
 
 
 def parse_master_args(master_args=None):
--- a/elasticdl/python/common/args.py
+++ b/elasticdl/python/common/args.py
@@ -417,7 +417,7 @@ def add_common_args_between_master_and_worker(parser):
         type=str,
         default="",
         help="The data reader parameters in a string separated by semi-colon "
-        'used to instantiate the data reader, e.g. "columns=["a", "b"]"',
+        'used to instantiate the data reader, e.g. "param1=1; param2=2"',
     )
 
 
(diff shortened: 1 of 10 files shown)


elasticdl/python/worker/worker.py
```diff
@@ -61,8 +63,11 @@ def __init__(
...
                 function/class in the model zoo, e.g.
                 "cifar10_subclass.CustomModel".
-            model_params: The dictionary of model parameters in a string that
-                will be used to instantiate the model,
+            model_params: The dictionary of model parameters in a string
+                separated by semi-colon used to instantiate the model,
+                e.g. "param1=1,param2=2".
+            data_reader_params: The data reader parameters in a string
+                separated by semi-colon used to instantiate the data reader,
                 e.g. "param1=1,param2=2".
```
ywskycn: Here we still use comma, but we use colon in other files..

elasticdl/python/common/args.py
```diff
@@ -403,15 +403,22 @@ def add_common_args_between_master_and_worker(parser):
         "--model_params",
         type=str,
         default="",
-        help="The dictionary of model parameters in a string that will be "
-        'used to instantiate the model, e.g. "param1=1,param2=2"',
+        help="The model parameters in a string separated by semi-colon "
+        'used to instantiate the model, e.g. "param1=1; param2=2"',
```
ywskycn, line 407: I'm not sure `comma`, `colon` which one better. What do ppl usually use in ml area?

elasticdl/python/worker/worker.py
terrytangyuan: Fixed.

elasticdl/python/common/args.py
terrytangyuan: This is defined by ourselves. For example, we can use either `--model_params="a=1；b=2" ` or `--model_params="a=1, b=2" `. But using semi-colon as separate will save us from writing a bug-free parser that parses string like `"c=["a", "b"], d=1"` where we cannot simply use `split(",")` since this string contains too many commas.

ywskycn (approved):

PR #1258, opened 2019-09-27
Fix task end index that should not exceed the max index for this shard

terrytangyuan:
We should not create task with end index that exceeds the maximally possible number of records available in this shard. This PR includes:
* Takes the maximally possible number of records available in this shard into account when finding the end index for each task.
* Fixed the task end index in the test case that should be 20 instead of 22 since 20 is the maximally possible number of records available in this shard.
* Renamed the variables and added comments so it's more readable.

### DIFF
--- a/elasticdl/python/master/task_dispatcher.py
+++ b/elasticdl/python/master/task_dispatcher.py
@@ -105,7 +105,9 @@ class _TaskDispatcher(object):
                 )
                 # If the start index is not smaller than end index,
                 # we need to find the correct end index by taking the start
-                # index into account.
+                # index into account. We should not create task with
+                # end index that exceeds the maximally possible number of
+                # records available in this shard.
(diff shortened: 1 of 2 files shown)


ywskycn:
> We should not create task with end index that exceeds the maximally possible number of records available in this shard. This PR includes:

So now we may have a task which is assigned 10 records, but actually it only needs to process 8 records and reaches the end of the shard? But in that case, the task just finishes after processing 8, still ok, right?

terrytangyuan:
Yes, all records will be assigned in tasks. Prior to this PR, the `test_create_tasks_with_non_zero_start_ind` test case will generate a task that reads records from index 19 to 22 which will cause `OutOfRange` exception since there are only 20 records available.

ywskycn (approved):

PR #1263, opened 2019-09-30
support arbitrary dtype in protobuf

### LINKED ISSUE #1262: Support Arbitrary Data type in protobuf
Currently, we only support data type float32 in protobuf. 

I am planning to refactor the evaluation process to fix issue #1082. To support metrics inherited from `tf.Metric`, worker need to transport labels and model outputs to master. We can't suppose that labels and model outputs are of type `float32`. So we need to support arbitrary data type in protobuf.

mhaoli:
Fixes #1262.

### DIFF
--- a/elasticdl/proto/elasticdl.proto
+++ b/elasticdl/proto/elasticdl.proto
@@ -1,6 +1,7 @@
 syntax = "proto3";
 
 import "google/protobuf/empty.proto";
+import "elasticdl/proto/tensor_dtype.proto";
 
 package master;
 
@@ -55,7 +56,7 @@ message Tensor {
     repeated int32 indices = 3;
 
     // Dtype of the tensor, e.g. "int64", "float32".
-    string dtype = 4;
+    TensorDtype dtype = 4;
 }
 
 message Model {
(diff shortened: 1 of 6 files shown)


elasticdl/proto/elasticdl.proto
```diff
@@ -53,6 +53,9 @@ message Tensor {
     // Indices will be tf.IndexedSlices.indices if the tensor is in the form
     // of tf.IndexedSlices. Ohterwise indices will be None.
     repeated int32 indices = 3;
+
+    // Dtype of the tensor, e.g. "int64", "float32".
+    string dtype = 4;
```
QiJune: Data type should be an enum type, string is too flexible.

skydoorkai: `enum` is clear. `string` can support arbitrary types (supported by Numpy) and simple to code.

mhaoli: Use enum type instead.

QiJune (approved):
Excellent

elasticdl/python/tests/ndarray_test.py
```diff
@@ -11,10 +11,6 @@
 
 class ConverterTest(unittest.TestCase):
     def test_ndarray_to_tensor(self):
-        # Wrong type, should raise
```
terrytangyuan, line 14: Should we use this space to test `DT_INVALID` instead of removing this test?

mhaoli: Good point. Added.

terrytangyuan (approved):

PR #1265, opened 2019-09-30
Change the parameter docker_image_prefix name

ywskycn:
We have a `docker_image_prefix` parameter, which actually maps to the `repository` in Docker [document](https://docker-py.readthedocs.io/en/stable/api.html) and CLI. To be consistent, it would be better to change it to `repository`.


wangkuiyi (approved):
LGTM!

PR #1266, opened 2019-10-01
Add elasticdl clean command to clean up local docker images

### LINKED ISSUE #1202: Add cleanup command to clean local built docker images
Currently, each ElasticDL job will build a local docker image, causing lots of images after submitting some jobs. Better to provide a command `elasticdl clean` to clean up some local docker images.
- `elasticdl clean --all` will clean all local docker images;
- `elasticdl clean --prefix ...` will only clean docker images with the given prefix. Here `prefix` refers to repository.

ywskycn:
To resolve #1202. Add `elasticdl clean` command to clean up local images. Support to different configs:
- `elasticdl clean --docker_image_repository repo1` is to clean all images with repository equals to `repo1`.
- `elasticdl clean --all` is to clean all local docker images.

### DIFF
--- a/elasticdl/python/elasticdl/api.py
+++ b/elasticdl/python/elasticdl/api.py
@@ -105,19 +105,25 @@ def predict(args):
 
 
 def clean(args):
-    if args.docker_image_prefix and args.all:
-        raise ValueError("--docker_image_prefix and --all cannot be specified at the same time")
-    if not args.docker_image_prefix and not args.all:
-        raise ValueError("--docker_image_prefix and --all")
-    
-    if args.all:
-    remove_images()
-    # if args.docker_image_prefix and args.all:
-    #     # throw exception
-    # if args.docker_image_prefix:
-    #     # Only clean images with the given prefix
-    # elif args.all:
-    #     # Clean all
+    if args.docker_image_repository and args.all:
+        raise ValueError(
+            "--docker_image_repository and --all cannot "
+            "be specified at the same time"
+        )
+    if not (args.docker_image_repository or args.all):
+        raise ValueError(
+            "Either --docker_image_repository or --all "
+            "needs to be configured"
+        )
+    repository = (
+        args.docker_image_repository if args.docker_image_repository else ""
+    )
+    remove_images(
+        docker_image_repository=repository,
+        docker_base_url=args.docker_base_url,
+        docker_tlscert=args.docker_tlscert,
+        docker_tlskey=args.docker_tlskey,
+    )
 
 
 def _submit_job(image_name, client_args, container_args):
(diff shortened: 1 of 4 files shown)


elasticdl/python/elasticdl/api.py
```diff
@@ -103,6 +104,28 @@ def predict(args):
...
+            "--docker_image_repository and --all cannot "
+            "be specified at the same time"
+        )
+    if not (args.docker_image_repository or args.all):
+        raise ValueError(
+            "Either --docker_image_repository or --all "
+            "needs to be configured"
+        )
+    repository = (
+        args.docker_image_repository if args.docker_image_repository else ""
```
terrytangyuan: `docker_image_repository` already defaults to `""` so this line doesn't seem necessary?

elasticdl/python/elasticdl/image_builder.py
```diff
@@ -81,6 +79,55 @@ def build_and_push_docker_image(
     return image_name
 
 
+def remove_images(
```
terrytangyuan, line 82: Is there an option `-f` for force removal?

elasticdl/python/elasticdl/image_builder.py
```diff
@@ -81,6 +79,55 @@ def build_and_push_docker_image(
...
+    for image in images:
+        image_id = image.get("Id")
+        if image_id:
+            logger.info("Removing image %s" % image_id)
+            try:
+                client.remove_image(image_id)
+            except docker.errors.APIError as e:
+                logger.warning("Failed to delete image %s: %s" % (image_id, e))
+    # If removing all images, we also run prune to remove untagged images
+    if not docker_image_repository:
```
terrytangyuan, line 123: Also check `--all`?

elasticdl/python/elasticdl/api.py
ywskycn: Good catch. Yes, we can directly use `docker_image_repository` here. I introduced `--all` here to make sure users really want to clean up all images. `docker_image_repository` = ""  also means `--all`.

elasticdl/python/elasticdl/image_builder.py
ywskycn: `docker_image_repository` = "" actually means removing `all`, in the docker python API.

elasticdl/python/elasticdl/image_builder.py
ywskycn: `force` option available in docker python API for cases where you want to remove an image in a force way, without considering dependencies and containers. I put `force`=False here by default.

terrytangyuan (approved):

PR #1268, opened 2019-10-02
Support saving metadata in data reader

terrytangyuan:
The addition of this metadata information would be useful in the following scenarios:
* When we want to cache some meta information (might be time consuming) that can be reused across  each `read_records()` call.
* When we want to validate the consistency of the result from each `read_records()` call.
* When users want to do more validation around the retrieved data in `dataset_fn()` instead of previously having to remember the indices for each column themselves, e.g. `record[0:3]` that represents the features and `record[4]` that represents the labels.

Right now only `ODPSDataReader` uses this for saving column names information, we can add more metadata information if needed in the future.

### DIFF
--- a/elasticdl/python/data/data_reader.py
+++ b/elasticdl/python/data/data_reader.py
@@ -9,6 +9,11 @@ from elasticdl.python.common.constants import ODPSConfig
 from elasticdl.python.data.odps_io import ODPSReader
 
 
+class Metadata(object):
+    def __init__(self, column_names):
+        self.column_names = column_names
+
+
 class AbstractDataReader(ABC):
     def __init__(self, **kwargs):
         pass
@@ -44,6 +49,13 @@ class AbstractDataReader(ABC):
         """
         return None
 
+    @property
+    def metadata(self):
+        """This method returns the `Metadata` object that contains
+         some metadata collected for the read records, such as the
+         list of column names."""
+        return Metadata(column_names=None)
+
 
 class RecordIODataReader(AbstractDataReader):
     def __init__(self, **kwargs):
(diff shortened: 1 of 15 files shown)


elasticdl/python/data/data_reader.py
```diff
@@ -78,18 +90,28 @@ def create_shards(self):
...
     def __init__(self, **kwargs):
         AbstractDataReader.__init__(self, **kwargs)
         self._kwargs = kwargs
+        self._metadata = Metadata(column_names=None)
 
     def read_records(self, task):
         reader = self._get_reader(
             table_name=self._get_odps_table_name(task.shard_name)
         )
+        if self._metadata.column_names is None:
```
ywskycn, line 108: So, if `_metadata.colume_names` is not None, but different from `get("columns")`, do we do any validation here?

model_zoo/odps_iris_dnn_model/odps_iris_dnn_model.py
```diff
@@ -17,13 +17,44 @@ def optimizer(lr=0.1):
...
+            ]
+            features = tf.concat(features, -1)
+            return tf.reshape(features, features_shape)
+
+        features_shape = (4, 1)
+        labels_shape = (1,)
+        if mode != Mode.PREDICTION:
+            if label_col_name not in metadata.column_names:
+                raise ValueError(
+                    'Missing the label column "%s" in from the retrieved '
```
ywskycn: Why not use `"` as the outer wrap, and `'%s'`?

terrytangyuan: Good idea. Fixed so it's consistent with the string below

elasticdl/python/data/data_reader.py
terrytangyuan: No we don't need validation here in this case since `self._kwargs.get("columns")` won't get changed at all (this is only specified in the constructor).

ywskycn (approved):

Issue #1252, opened 2019-09-27
How to do evaluation


QiJune:
Users usually do an evaluation after several steps of training to see the performance of the model. Besides, when doing checkpoint, we usually want to make an evaluation of the checkpoint.

So, the evaluation might have two kinds:

- Evaluation when doing checkpoint. 

Here, we should introduce a global barrier to sync all workers and pservers. And then, master sends evaluation task to workers. Workers just pull parameters from pservers and will not send gradient. And the evaluation metric is calculated in master.

- Other Evaluation

We allow some in-consistency, some workers do evaluation, some workers do training, pservers still update the model. The evaluation metric is calculated in worker locally.

workingloong:
I summarize the existing evaluation design in EDL and some proposals are given in the 
https://github.com/workingloong/elasticdl/blob/worker_task_execution/doc/evaluation_design.md

PR #1273, opened 2019-10-08
[Draft] Add pserver detail part

QiJune:
[Here](https://github.com/QiJune/elasticdl/blob/ps_detail/doc/pserver.md) is for a better review.


doc/pserver.md
```diff
@@ -0,0 +1,188 @@
...
+
+### Compoments
+
+
+![pserver](./images/pserver.png)
+
+
+PServer contains two main compoments:
+
+- KVStore Servier
```
brightcoder01: Servier -> Server

doc/pserver.md
```diff
@@ -0,0 +1,188 @@
...
+**Question 1** out-of-memory
+
+Workers push gradients to the gradient queue of KVStore, pserver optimizer gets gradient from the queue, and does parameter optimization.
+
+This is a classical producer-consumer problem. We have to ensure that the speed of optimizer processing gradients is large than the speed of workers pushing gradients. Otherwise, the gradient queue will become larger and larger and lead to out-of-memory.
+
+We could set a limit size to the queue, if the condition is satisfied:
+
+- just throwing away upcoming gradients from workers
+- blocking `push_gradient` return from KVStore to workers
```
brightcoder01, line 150: blocking is not recommended. It will cause the GRPC call timeout. And the hang will waste up the RPC ThreadPool, the following RPC calls will also fail.

doc/pserver.md
```diff
@@ -0,0 +1,188 @@
...
+
+
+PServer contains two main compoments:
+
+- KVStore Servier
+- Optimizer
+
+1. The worker initializes a model, and push parameters to KVStore.
+2. Before each step of training, the worker pulls the latest model from KVStore.
+3. After forward/backward computation, the worker push gradients to KVStore. And KVStore will put the gradients to a Queue waiting for processing.
```
brightcoder01: KVStore is just for storage, provides put/get Api;
Optimizer is responsible for calculation;
The gradient queue can be put in a third component or just hold by PServer;
I think separate the gradient queue from KVStore will be cleaner

doc/pserver.md
```diff
@@ -0,0 +1,188 @@
...
+
+![pserver](./images/pserver.png)
+
+
+PServer contains two main compoments:
+
+- KVStore Servier
+- Optimizer
+
+1. The worker initializes a model, and push parameters to KVStore.
```
mhaoli: `push` --> `pushes`?

doc/pserver.md
```diff
@@ -0,0 +1,188 @@
...
+        pass
+
+    def get_param(self):
+        pass
+
+    def set_param(self):
+        pass
+
+    def get_gradient(self):
+        pass
```
mhaoli: May these functions be internal functions?

doc/pserver.md
```diff
@@ -0,0 +1,188 @@
...
+
+    # embedding related interface is exposed explictly
+    def pull_embedding_param(self, request, _):
+        pass
+
+    def push_embedding_param(self, request, _):
+        pass
+
+    def push_embedding_gradient(self, request, _):
```
mhaoli: We should push all gradients in one gprc. So there is no need for `push_embedding_gradient`.

doc/pserver.md
```diff
@@ -0,0 +1,188 @@
...
+        pass
+
+    def get_embedding_param(self):
+        pass
+
+    def set_embedding_param(self):
+        pass
```
mhaoli: Same. May these functions be internal functions?

doc/pserver.md
```diff
@@ -0,0 +1,188 @@
...
+        FP16 = 4;
+        FP32 = 5;
+        FP64 = 6;
+    }
+    string name = 1;
+    DataType data_type = 2;
+    repeated int64 dim = 3;
+    bytes content = 4;
+    repeated int64 indices = 5;
+    int64 version = 6;
```
mhaoli: Maybe consider supporting name and version in the next step?

doc/pserver.md
QiJune: KVStore provides remote API, such as `pull_param/push_param` for workers, and also provides local API such as `get_param/set_param` for the optimizer.

If we make gradient queue as an independent component, then the queue will provide remote API `push_gradient` for worker and local API `get_gradient` for optimizer.

I am fine with making gradient queue an independent component.

doc/pserver.md
QiJune: Currently, we put `apply_gradient` inside `ReportGradient` service. The `apply_gradient` will consume some time, it's a kind of blocking. 

I mean worker producing gradients faster than pserver consuming, so we have to let workers wait for some time. Or we could set a while-loop here to check queue size. If the queue size down to some value, we could let the grpc call return.

doc/pserver.md
QiJune: These `get/set` APIs are local API to operate the parameter DB. The optimizer will use these APIs.

The `push/pull` APIs are remote API for workers.

These APIs are both used for external purposes.

doc/pserver.md
QiJune: Got it, will remove this.

doc/pserver.md
QiJune: Now, we pull a submodel from pserver, and push gradients to pserver. (In next step, we may want to handle these layer by layer)
So the version is actually for submodel.  And for name,  it's actually the key of KVStore.

How about adding another message:

```python
message Tensor {
    enum DataType {
        BOOL = 0;
        INT16 = 1;
        INT32 = 2;
        INT64 = 3;
        FP16 = 4;
        FP32 = 5;
        FP64 = 6;
    }
    string name = 1;
    DataType data_type = 2;
    repeated int64 dims = 3;
    bytes content = 4;
    repeated int64 indices = 5;
}

message Model {
    int64 version = 1
    repeated Tensor tensors = 2;
}
```

doc/pserver.md
mhaoli: Got it.

doc/pserver.md
mhaoli: Agree.

doc/pserver.md
```diff
@@ -0,0 +1,226 @@
...
+In sync mode, optimizer needs to wait for a certain number of gradients, and then get the gradient after addition.
+
+We could implement a customized queue structure to support such logic efficiently.
+
+
+### Checkpoint
+
+
+
+### Evaluation
```
QiJune: We are still under discussing, please refer to #1252

QiJune: After discussing with @workingloong , we reach such summary:

1. Evaluation, one way:

We allow some in-consistency, some workers do evaluation, some workers do training, and pservers still update the model.


2. Checkpoint, two ways:

- Master sends sync signal to all workers and pservers, and training job is stopped. Evaluation job will be started. After evaluation, pservers will save checkpoint respectively. And after checkpoint is saved, the training job will be started again.

- Master send signal to all pservers, and tell pservers to save checkpoint. Then, master launch another evaluation worker pods and pserver pods to do evaluation job. The evaluation pserver pod will load the latest checkpoint. The training job will be still runing at the same time.

doc/pserver.md
```diff
@@ -0,0 +1,226 @@
...
+![pserver](./images/pserver.png)
+
+
+PServer contains three main compoments:
+
+- KVStore
+- Gradient Queue
+- Optimizer
+
+1. The worker initializes a model, and pushes parameters to KVStore.
```
brightcoder01, line 22: How do we choose which worker does this initialization job?

chunyang-wen: push **gradients**?

doc/pserver.md
```diff
@@ -0,0 +1,226 @@
...
+2. Before each step of training, the worker pulls the latest model from KVStore.
+3. After forward/backward computation, the worker push gradients to Gradient Queue waiting for processing.
```
chunyang-wen: Please add a proper `a` or `the` at multiple places here.

doc/pserver.md
```diff
@@ -0,0 +1,226 @@
...
+    rpc pull_model(Model) returns (google.protobuf.Empty) {}
+    rpc pull_embedding_param(Tensor) returns (EmbeddingResponse) {}
+}
+
+service GradientQueue {
+    rpc push_gradient(Model) returns (google.protobuf.Empty) {}
+}
+```
+
+### Gradient Queue
```
skydoorkai, line 123: As discussed, we will use a queue in sync-SGD. In async-SGD, we do not use the queue but apply_gradient directly in the 1st version. 
We may consider queue for async-SGD later.

doc/pserver.md
```diff
@@ -0,0 +1,226 @@
...
+So the RPC service will be defined as following:
+
+```proto
+message EmbeddingResponse{
+    Tensor value = 1;
+    repeated int64 unknown_indices = 2;
+}
+
+service KVStore{
+    rpc push_model(Model) returns (google.protobuf.Empty) {}
```
skydoorkai: `push_model` corresponds to current `ReportVariable`.
`pull_model` corresponds to current `GetModel`, the return is not empty.
`pull_embedding_param` correspnds to current `lookup_embedding`. May change name to `lookup_embedding` or `pull_embedding_vectors`.

doc/pserver.md
```diff
@@ -0,0 +1,226 @@
...
+        pass
+    
+    def get_gradient(self):
+        pass
+        
+    def put_gradient(self):
+        pass
+```
+
+### Parameters Transfromation
```
skydoorkai: We need to add a subsection to describe the data structure in PS for storing model variables and embedding vectors.
For a model variable, the key is the variable name, and the value is `tf.Variable`.
For an embedding vector, the key is embedding layer name + id, the value is a Numpy array.

doc/pserver.md
QiJune: Currently, all workers will do the initialization job, and push parameters to KVStore. If the KVStore has the parameter already, it will discard the later parameter.

doc/pserver.md
QiJune: Done

doc/pserver.md
QiJune: Done

doc/pserver.md
```diff
@@ -0,0 +1,257 @@
...
+
+
+
+### KVStore
+
+For a common model variable, we use save it as a `tf.Variable` in Parameter DB. For the embedding table, we introduce a customized data structure.
+
+
+```python
+class EmbeddingTable(object):
```
QiJune, line 86: Here, I think that we should have a data structure of EmbeddingTable. One embedding layer has one embedding table.  It avoids frequent `get_key`/`split_key` operations, and has better organization of EmbeddingTable parameter.

The embedding table now store <id, embedding_vector> pairs. This is a Python class now, we could re-implement it as a C++ structure in the future for better performance if necessary.


If there is a embedding gradient, the optimization logic could be:

```python
y, idx = tf.unique(grad.indices)
grad.indices = idx.numpy()
param = self.kvstore.get_embedding_vector(grad.name, y.numpy())
param_var = convert_to_var(param)
grads_and_params = list(zip([convert_to_tensor(grad)], [param_var]))
self.opt.apply_gradients(grads_and_params)
param.value = param_var.numpy()
self.kvstore.set_embeddint_vector(param.name, param.indices, param.value)
```

doc/pserver.md
QiJune: Done

doc/pserver.md
mhaoli: If we only store <id, embedding_vector> pairs, we need to create multiple embedding tables for one embedding layer. Because we also have slot variables such as momentum. For slot variables, the key in redis now is `<layer_name>-<slot_name>-<id>`.

mhaoli: Also, the optimization logic has implemented in `OptimizerWrapper` now.

QiJune: Yes, exactly. If we store embedding table param as an EmbeddingTable object, we also need to make embedding table mementum slot as a EmbeddingTable object.

I write a simple test script to record the time:


```python
import numpy as np
from random import randrange
from time import process_time


id_num = 100000
embedding_dim = 128
id_limit = 10000000
layer_name = "tom"

embedding_vectors = np.random.uniform(-1, 1, size=(id_num, embedding_dim))
set_ids = [randrange(id_limit) for i in range(id_num)]
get_ids = [randrange(id_limit) for i in range(id_num)]

class EmbeddingTable(object):
    def __init__(self, name):
        self.name = name
        self.vectors = {}

    def set(self, ids, value):
        for index, id in enumerate(ids):
            row = value[index,:]
            self.vectors[id] = row

    def get(self, ids):
        unknow_ids = []
        know_ids = []
        res = []
        for id in ids:
            if id not in self.vectors:
                unknow_ids.append(id)
            else:
                know_ids.append(id)
                res.append(self.vectors[id])

        res = np.stack(res)
        return res, unknow_ids



class EmbeddingTable2(object):
    def __init__(self):
        self.db = {}

    def db_set(self, ids, value):
        for index, id in enumerate(ids):
            key = layer_name + "_" + str(id)
            row = value[index,:]
            self.db[key] = row

    def db_get(self, ids):
        unknow_ids = []
        know_ids = []
        res = []
        for id in ids:
            key = layer_name + "_" + str(id)
            if key not in self.db:
                unknow_ids.append(id)            
            else:
                know_ids.append(id)
                res.append(self.db[key])

        res = np.stack(res)
        return res, unknow_ids


start = process_time()

t1 = EmbeddingTable(layer_name)
t1.set(set_ids, embedding_vectors)
t1.get(get_ids)

end = process_time()
print("%.2gs" % (end-start))


start = process_time()
t2 = EmbeddingTable2()
t2.db_set(set_ids, embedding_vectors)
t2.db_get(get_ids)

end = process_time()
print("%.2gs" % (end-start))
```


The result is 
```
0.072s
0.15s
```

The first one is almost twice faster than the second one.

Anyway, I mean that we should extract EmbeddingTable as an independent data structure. It provides`get/set` interface.  We could find better and better concrete implementations. Or even we will implement it in C++, and expose it to Python later.

doc/pserver.md
```diff
@@ -0,0 +1,261 @@
...
+```
+
+Model could also be used as gradients collection.
+
+So the RPC service will be defined as following:
+
+```proto
+message EmbeddingResponse{
+    Tensor value = 1;
+    repeated int64 unknown_indices = 2;
```
skydoorkai: No need to include unknown_indices, as PS will initialize the embedding vectors if not exist.

QiJune:
Please refer to #1295

PR #1274, opened 2019-10-08
[WIP] PS design overview part and embedding replica part

skydoorkai:
This PR should be combined with other parts (PR https://github.com/sql-machine-learning/elasticdl/pull/1273).

Also, need to use the same definitions in the design.

### DIFF
--- a/doc/ps_design.md
+++ b/doc/ps_design.md
@@ -1,0 +1,19 @@
+# ElasticDL Parameter Server Design Doc
+
+## Overview
+Currently there is one parameter server (PS) co-existed with the master. In order to support multiple PS nodes and PS fault tolerance, we need to seperate PS from the master. Besides a KV store for model variables and embedding tables, PS should also support model variable update and sparse embedding table update using gradients. For a sparse embedding table update, it will update a number of embedding vectors in the embedding table.
+
+The master will create *N* PS pods, with *N* as the number of PS nodes. Each model variable and embedding vector has a corresponding PS node. Thus, every PS node has a subset of model variables and embedding tables.
+
+The master will monitor the PS pods status similar as it does for the worker pods. In case a PS pod fails, the master will try to relaunch the PS pod. Since PS pods have higher priority than worker pods, if there are still some running worker pods, the relaunch will succeed by using either idled or preempted Kubernetes resouces. If the relaunch fails, there are no worker pods left. The Elastic job has to wait for resources for the PS pod and worker pods.
+
+Each worker has a local copy of model variables. After the master relaunches a PS node, the PS pod can recover model variables from workers. For embedding vectors, PS must create replicas to support fault tolerance. For each PS node *Ni*, it will store *M* replicas in following *M* PS nodes *Ni+1, Ni+2, ... Ni+M*. We should use division remainder for PS node indices with the divided number as *N*. The relaunched PS can recover embedding vectors from one of its replicas. If there are more than *M* continously-indexed PS nodes failing, at least one PS node fails with all of its replica PS nodes. The ElasticDL job has to recover from a recent checkpoint.
+
+## Parameter Server (PS)
+
+## Interactions among Master, PS and Worker
+### Master - PS
+### Master - Worker
+### Worker - PS
+
+## Embedding Replicas in PS
--- a/doc/ps_design.md
+++ b/doc/ps_design.md
@@ -1,13 +1,13 @@
 # ElasticDL Parameter Server Design Doc
 
 ## Overview
-Currently there is one parameter server (PS) co-existed with the master. In order to support multiple PS nodes and PS fault tolerance, we need to seperate PS from the master. Besides a KV store for model variables and embedding tables, PS should also support model variable update and sparse embedding table update using gradients. For a sparse embedding table update, it will update a number of embedding vectors in the embedding table.
+Currently, there is one parameter server (PS) co-existed with the master. In order to support multiple PSs and PS fault tolerance, we need to separate PS from the master. Besides a KV store for model variables and embedding tables, PS should also support model variable update and sparse embedding table update using gradients. For a sparse embedding table update, it will update a number of embedding vectors in the embedding table.
 
-The master will create *N* PS pods, with *N* as the number of PS nodes. Each model variable and embedding vector has a corresponding PS node. Thus, every PS node has a subset of model variables and embedding tables.
+The master will create *N* PSs, with *N* as the number of PSs specified by the user. Each model variable and embedding vector has a corresponding PS. Thus, every PS has a subset of model variables and embedding tables.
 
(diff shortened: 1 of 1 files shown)


mhaoli:
LGTM. @QiJune please double check.

doc/ps_design.md
```diff
@@ -0,0 +1,19 @@
...
+
+The master will monitor PS pods status similar to what it does for worker pods. In case a PS pod fails, the master will try to relaunch it. Since PS pods have higher priority than worker pods, if there are still some running worker pods, the relaunch will succeed by using either idled or preempted Kubernetes resources. If the relaunch fails, there are no worker pods left. The Elastic job has to wait for resources for the PS pod and worker pods.
+
+Each worker has a local copy of the model variables. After the master relaunches a PS, the PS pod can recover model variables from workers. For embedding vectors, PS must create replicas to support fault tolerance. For each PS *P(i)*, it will store *M* replicas in the following *M* PSs *P(i+1 % N), P(i+2 % N), ... P(i+M % N)*. The relaunched PS can recover embedding vectors from one of its replicas. If there are more than *M* continuously-indexed PSs failing, at least one PS fails with all of its replica PSs. The ElasticDL job has to recover from a recent checkpoint.
```
brightcoder01: Since PS recover the embedding table from the neighborhood, how about recovering the variable from the neighborhood PS instead of worker too?
Remove the PS's dependency on the worker.
In the resource shortage case, all the workers are preempted, we are launch a PS node at this time, and we can't recover the PS from any worker.

doc/ps_design.md
```diff
@@ -0,0 +1,19 @@
+# ElasticDL Parameter Server Design Doc
+
+## Overview
+Currently, there is one parameter server (PS) co-existed with the master. In order to support multiple PSs and PS fault tolerance, we need to separate PS from the master. Besides a KV store for model variables and embedding tables, PS should also support model variable update and sparse embedding table update using gradients. For a sparse embedding table update, it will update a number of embedding vectors in the embedding table.
+
+The master will create *N* PSs, with *N* as the number of PSs specified by the user. Each model variable and embedding vector has a corresponding PS. Thus, every PS has a subset of model variables and embedding tables.
+
+The master will monitor PS pods status similar to what it does for worker pods. In case a PS pod fails, the master will try to relaunch it. Since PS pods have higher priority than worker pods, if there are still some running worker pods, the relaunch will succeed by using either idled or preempted Kubernetes resources. If the relaunch fails, there are no worker pods left. The Elastic job has to wait for resources for the PS pod and worker pods.
```
brightcoder01: idled -> idle

doc/ps_design.md
skydoorkai: If we save all variables in the PS replicas, we need *M* times more memory for PS to store the model and its replicas and much more bandwidth for replication. 
The logic for PS to recover variables from workers is the same as model variable initialization (`ReportVariable`). 
If all workers are preempted, PS has to recover from a recent checkpoint.
In a K8S cluster with fixed resources, a lite PS (with less memory) has a much smaller chance to preempt all worker pods than a heavy PS(with *M*x large memory).

doc/ps_design.md
```diff
@@ -1,19 +1,80 @@
...
 ## Interactions among Master, PS and Worker
-### Master - PS
-### Master - Worker
-### Worker - PS
+
 
 ## Embedding Replicas in PS
+An ElasticDL job has *N* PSs. Embedding vectors are partitioned into these *N* PSs. The user provides *M*, the number of replicas for embedding vectors. *M* must be smaller than *N* as each PS uses other PSs to store its embedding replicas.
+
+Assume *E(i)* is the embedding vectors in PS *PS(i)*, it has *M* replicas which are stored in PSs from *P(i + 1 % N)* to *P(i + M % N)*. Also, *PS(i)* has replicas for *E(i - M % N)* to *E(i - 1 % N)*.
```
mhaoli, line 20: Both symbols *PS(i)* and *P(i)* refer to the i-th PS in this doc.

doc/ps_design.md
```diff
@@ -1,19 +1,80 @@
...
 
 ## Embedding Replicas in PS
+An ElasticDL job has *N* PSs. Embedding vectors are partitioned into these *N* PSs. The user provides *M*, the number of replicas for embedding vectors. *M* must be smaller than *N* as each PS uses other PSs to store its embedding replicas.
+
+Assume *E(i)* is the embedding vectors in PS *PS(i)*, it has *M* replicas which are stored in PSs from *P(i + 1 % N)* to *P(i + M % N)*. Also, *PS(i)* has replicas for *E(i - M % N)* to *E(i - 1 % N)*. 
+
+*PS(i)* stores *E(i)* as a dictionary so that every embedding vector has a corresponding key.  *PS(i)* maintains *M* updated embedding vector key sets *UKS_i(j) for j from 0 to M - 1*. When *PS(i)* sparsely updates its embedding vectors *E(i)*, it also add the updated embedding vector keys into these *M* sets. 
+
+
+*PS(i)* also periodically synchronize the replicas stored in it from *E(i - M % N)* to *E(i - 1 % N)*. The synchronization frequency can be several seconds.
```
mhaoli, line 25: Maybe add a sentence illustrating *UKS_i(j)* is used for synchronization? Otherwise there is a logic gap between these two paragraphs.

doc/ps_design.md
```diff
@@ -1,19 +1,80 @@
...
+# GRPC service for PS to recover embedding vectors after relaunch
+rpc GetReplica(SynchronizeEmbeddingRequest) returns (SynchronizeEmbeddingResponse);
+```
+
+Each PS has a thread dedicated for replica synchronization. In this dedicated thread, PS will synchronize the stored replicas in it.
+
+```
+# T is the number of seconds for synchronization frequency
+# Assume current PS is PS(i), self._stub[index] is the stub for PS(i - index)'s GRPC server.
+# self.replicas[index] is the replica for PS(i - index).
```
mhaoli, line 50: Is the range of `index` [0,M) or [1, M]?
If the range is [0, M), `self.replicas[index]` should be the replica for PS(i - index - 1).

doc/ps_design.md
QiJune: The index should be *P((i + 1) % N)* , *P((i + M) % N)*?

doc/ps_design.md
```diff
@@ -0,0 +1,80 @@
...
+
+message SynchronizeEmbeddingResponse {
+    map<string, Tensor> embedding_vectors = 1;
+}
+
+# GRPC service for replica synchronization
+rpc SynchronizeEmbedding(SynchronizeEmbeddingRequest) returns (SynchronizeEmbeddingResponse);
+
+# GRPC service for PS to recover embedding vectors after relaunch
+rpc GetReplica(SynchronizeEmbeddingRequest) returns (SynchronizeEmbeddingResponse);
```
QiJune, line 42: Not very clear this service used for. Could you please give more explanation?

QiJune:
Please refer to #1295

PR #1275, opened 2019-10-08
Refactor the process of calculating evaluation metrics

### LINKED ISSUE #1082: Refactor the process of calculating evaluation metrics
# Currently Existing Problems
Currently, during the process of evaluation, workers report metrics calculated from every minibatch to master. We assume that the metrics reported in every minibatch was independent and master calculates the average of those metrics.

However, metrics inherited from `tf.keras.metrics` does not behave like this, e.g. `tf.keras.metrics.AUC`. These are streaming metrics, i.e. each time they are called, they calculate the metrics from all the minibatches they have received so far. More specifically, when `tf.keras.metrics.AUC` receives the model outputs and labels of the second batch, it calculates AUC of the first two batches rather than the second batch only. 

As a result, in order to support metrics inherited from `tf.keras.metrics`, we should refactor the process of calculating evaluation metrics.

# New Evaluation Metrics Interface
Referring to the evaluation metric interface of Keras, we plan to support two kinds of interfaces for calculating evaluation metrics:
* Keras-native streaming metrics, i.e. metric classes in `tf.keras.metrics`.
* Stateless functions will also be supported. However, we require the arguments of the stateless functions to be `(lables, model_outputs)`, just like [Keras](https://keras.io/metrics/#custom-metrics) has required. 

An example of new interfaces:
```
def eval_metrics_fn():
    return {
        "acc": tensorflow.keras.metrics.Accuracy(),
        "mse": tensorflow.keras.metrics.MeanSquaredError(),
        "acc_fn": lambda lables, outputs: tf.equal(
            tf.cast(labels, tf.int32), 
            tf.cast(outputs, tf.int32), 
        ),
    }
```

# How to calculate evaluation metrics
For metric instances of `tf.keras.metrics.Metric`, the logic of calculating evaluation metrics can be represented by the following code (take AUC for an example):
```
metric_instance = tf.keras.metrics.AUC()
for minibatch_data in dataset:
    inputs, labels = minibatch_data
    outputs = forward_pass(inputs)
    metric_instance.update_state(labels, outputs)
print(metric_instance.result())
```

For stateless functions, Keras provides `MeanMetricWrapper` which wraps a stateless function into a streaming metric instance.

Please note that evaluation process may involve multiple workers, each worker calculates a subset of evaluation data. Thus, workers should report model outputs and labels to the master, and the master call `metric_instance.update_state`.

# Different Metrics for Different Model Outputs
Some models may have multiple outputs. According to compile method part in  Keras [official tutorial](http://faroit.com/keras-docs/2.0.4/models/model/), Keras supports users to specify different metrics for different model outputs, such as `metrics={'output_a': 'accuracy', 'output_b': 'AUC'}`. This is necessary because calculating AUC requires that model has multiple outputs.
ElasticDL support models with multiple outputs now. We should also support different metrics for different outputs. Thus we provide two kinds of evaluation metric interfaces:
```
# Interface for models with only one output. 
# ElasticDL will calculate "acc" and "mse" using the model output.
def eval_metrics_fn_for_one_output():
    return {
        "acc": tensorflow.keras.metrics.Accuracy(),
        "mse": tensorflow.keras.metrics.MeanSquaredError(),
    }

# Interface for models with multiple outputs. 
# For model output "output_a", ElasticDL will calculate "acc" and "mse". 
# For "output_b", ElasticDL will calculate "auc".
def eval_metrics_fn_for_multiple_outputs():
    return {
        "output_a": {
            "acc": tensorflow.keras.metrics.Accuracy(),
            "mse": tensorflow.keras.metrics.MeanSquaredError(),
        },
        "output_b": {
            "auc": tensorflow.keras.metrics.AUC(),
        },
    }
```

mhaoli:
Fixes #1082. Changes include:
* Support two kinds of evaluation metric interface.
* Workers report model outputs and labels to the master, and the master calculates evaluation metrics.
* This PR supports users to specify different metrics for different model outputs. `model_zoo/deepfm_functional_api` can be a good example. Model have two outputs, named `logits` and `probs`. We calculate `accuracy` metric for output `logits` and `auc` metric for output `probs`. Also note that calculating loss only uses `logits`.


elasticdl/python/master/evaluation_service.py
```diff
@@ -13,12 +12,48 @@
 class _EvaluationJob(object):
     """Representation of an evaluation job"""
 
-    def __init__(self, model_version, total_tasks=-1):
+    def __init__(self, metrics_dict, model_version, total_tasks=-1):
+        """
+        Args:
+            metrics_dict: A python dictionary. If model has only one output,
```
skydoorkai, line 19: Specify that each metric uses only one model output.

elasticdl/python/master/evaluation_service.py
```diff
@@ -13,12 +12,48 @@
...
+        first_metrics = list(metrics_dict.values())[0]
+        if isinstance(first_metrics, dict):
+            self._model_have_multiple_outputs = True
+            self._metrics_dict = metrics_dict
+        else:
+            # When model has only one output, save it in a dict in order to
+            # keep the same data structure as the `metrics_dict` when model
+            # has multiple outputs.
+            self._model_have_multiple_outputs = False
+            self._metrics_dict = {"default": metrics_dict}
```
skydoorkai: metrics_dict is a dict with output name as its keys. Thus change "default" to "output".

mhaoli: Done.

elasticdl/python/master/evaluation_service.py
mhaoli: Done.

elasticdl/python/master/evaluation_service.py
```diff
@@ -38,18 +74,30 @@ def report_evaluation_metrics(
...
-            if k in self._evaluation_metrics:
-                self._evaluation_metrics[k] += tensor_to_ndarray(v)
-            else:
-                self._evaluation_metrics[k] = np.copy(tensor_to_ndarray(v))
-        self._completed_minibatches += 1
+        labels = tensor_to_ndarray(labels)
+        for key, tensor in model_outputs.items():
+            metrics = self._metrics_dict.get(key, {})
+            if not metrics:
+                continue
```
workingloong: `if key not in self.metrics_dict: continue` may be more readable.

elasticdl/proto/elasticdl.proto
```diff
@@ -112,7 +112,8 @@ message ReportEvaluationMetricsResponse {
 message ReportEvaluationMetricsRequest {
     // Model version used to compute evaluation metrics.
     int32 model_version = 1;
-    map<string, Tensor> evaluation_metrics = 2;
+    map<string, Tensor> model_outputs = 2;
+    Tensor labels = 3;
```
chunyang-wen, line 116: can we only use a `map` instead of a `map` plus `labels`?

mhaoli: The keys of `model_outputs` are defined by users. It seems difficult to put `labels` into that map?

elasticdl/python/master/evaluation_service.py
```diff
@@ -13,12 +12,49 @@
...
-        self._completed_minibatches = 0
-        self._evaluation_metrics = defaultdict()
+        self._init_metrics_dict(metrics_dict)
+
+    def _init_metrics_dict(self, metrics_dict):
+        if not metrics_dict:
+            raise ValueError(
+                "Doing evaluation with empty evaluation metrics dict."
+            )
+        first_metrics = list(metrics_dict.values())[0]
```
chunyang-wen, line 43: It is a little weird here. I have noticed that `get_evaluation_summary` is dependent on this variable `self._model_have_multiple_outputs`. It is better we have a place to unify those two cases and make other functions consistent.

elasticdl/proto/elasticdl.proto
chunyang-wen: By making the reporting data more general, related logic can also be more flexible.

elasticdl/python/master/evaluation_service.py
mhaoli: Yes, it is a little weird here. This is because I want to support two kinds of user interface, as illustrated in the docstring of `_EvaluationJob.__init__`. 

Function `__init__` unifies these two cases into a consistent structure. 
Function `report_evaluation_metrics` calculates metrics through this structure. 
Function `get_evaluation_summary` returns the calculated metrics and logger will print the returned metrics. `get_evaluation_summury` should return the metrics in the similar form as the user interface.

elasticdl/proto/elasticdl.proto
mhaoli: It is difficult to make reporting data more general. We can modify the proto file in another PR if we have a better way.

elasticdl/python/master/evaluation_service.py
```diff
@@ -13,12 +12,49 @@
...
         self._total_tasks = total_tasks
         self._completed_tasks = 0
-        self._completed_minibatches = 0
-        self._evaluation_metrics = defaultdict()
+        self._init_metrics_dict(metrics_dict)
+
+    def _init_metrics_dict(self, metrics_dict):
+        if not metrics_dict:
+            raise ValueError(
+                "Doing evaluation with empty evaluation metrics dict."
```
terrytangyuan: "Evaluation metrics dictionary must not be empty."

elasticdl/python/master/evaluation_service.py
```diff
@@ -13,12 +12,49 @@
...
+        first_metrics = list(metrics_dict.values())[0]
+        if isinstance(first_metrics, dict):
+            self._model_have_multiple_outputs = True
+            self._metrics_dict = metrics_dict
+        else:
+            # When model has only one output, save it in a dict in order to
+            # keep the same data structure as the `metrics_dict` when model
+            # has multiple outputs.
+            self._model_have_multiple_outputs = False
+            self._metrics_dict = {"output": metrics_dict}
```
terrytangyuan: Move "output" to a var/const since it's also used in line 99 which is easy to get messed up

elasticdl/python/tests/test_module.py
```diff
@@ -34,8 +34,8 @@ def optimizer(lr=0.1):
     return tf.optimizers.SGD(lr)
 
 
-def eval_metrics_fn(predictions, labels):
-    return {"mse": tf.reduce_mean(tf.square(predictions - labels))}
+def eval_metrics_fn():
+    return {"mse": lambda labels, outputs: tf.square(outputs - labels)}
```
terrytangyuan, line 38: I haven't looked closely yet but double check whether this is `labels, outputs` or `outputs, labels` (same in other examples)

elasticdl/python/worker/worker.py
```diff
@@ -366,11 +365,12 @@ def training_process_eagerly(self, features, labels):
         grads = tape.gradient(loss, self.get_trainable_items())
         return loss, grads
 
+    # TODO (yunjian.lmh): `evaluation_process` and `predict_process` are the
+    # same now.
```
terrytangyuan: Why isn't "labels" used?

model_zoo/deepfm_functional_api/deepfm_functional_api.py
```diff
@@ -110,16 +110,13 @@ def _parse_data(record):
     return dataset
 
 
-def eval_metrics_fn(predictions, labels):
-    labels = tf.reshape(labels, [-1])
-    predictions = tf.reshape(predictions, [-1])
-    global AUC_metric
```
terrytangyuan, line 116: Is it safe to remove "global" now?

elasticdl/python/master/evaluation_service.py
```diff
@@ -13,12 +12,49 @@
...
+        for output_name, metrics in self._metrics_dict.items():
+            for metric_name, metric in metrics.items():
+                if not isinstance(metric, metrics_module.Metric):
+                    metrics[metric_name] = metrics_module.MeanMetricWrapper(
```
terrytangyuan, line 58: Where is `MeanMetricWrapper` defined and why is it used? Maybe add a comment here

elasticdl/python/master/evaluation_service.py
mhaoli: Done.

elasticdl/python/tests/test_module.py
mhaoli: I have checked that here should be `labels, outputs`. 

Because this PR calculates evaluation metrics in the same way as Keras, we can refer to the "Custom metrics" section in this [link](http://faroit.com/keras-docs/2.0.4/metrics/).

elasticdl/python/worker/worker.py
mhaoli: Because workers report model outputs and labels to the master, and the master calculating evaluation metrics.

model_zoo/deepfm_functional_api/deepfm_functional_api.py
mhaoli: Yes, because `eval_metrics_fn` returns the AUC metric instance now. Master saves this instance in `_EvaluationJob`.

elasticdl/python/master/evaluation_service.py
mhaoli: `MeanMetricWrapper` is defined in  `tf.keras.metrics`. 
The idea to use `MeanMetricWrapper` to wrap stateless functions is borrowed from Keras.
I will add a comment here.

elasticdl/python/master/evaluation_service.py
mhaoli: Done.

terrytangyuan (approved):

Issue #1248, opened 2019-09-26
Rename arguments "--*_data_dir" to "--*_data"


terrytangyuan:
Currently, if we want to use ODPS data source, we can pass ODPS table names to "--*_data_dir" arguments instead of previously passing file directories if we want to use RecordIO files as data source. I am proposing the following changes so the data sources are not "file"-dependent:
```
training_data_dir -> training_data
evaluation_data_dir -> evaluation_data
prediction_data_dir -> prediction_data
```
We can add help information like this:
```
This can be either the data directory that contains RecordIO files or an ODPS table name used for training.
```
If there are any suggestions on other names, please let me know in this issue as well.

chunyang-wen:
Is it better to append a string "path" at the end?

terrytangyuan:
@chunyang-wen But in the ODPS case, this should be a table name instead. Either a "path" or "directory" doesn't make much sense.

chunyang-wen:
@terrytangyuan For ODPS, path=odps://project.table. Previously we use path for ODPS table. I agree with you.

ywskycn:
Let's rename the parameter.

PR #1279, opened 2019-10-09
Add integration test for ElasticDL job with ODPS data source

terrytangyuan:
This setups the integration test on Travis which includes:
* Sets up the testing data table on ODPS
* Submits ElasticDL training job using ODPS data source
* Validates the pod statuses
* Destroys the testing data table on ODPS

### DIFF
--- a/scripts/client_test.sh
+++ b/scripts/client_test.sh
@@ -82,7 +82,6 @@ elif [[ "$JOB_TYPE" == "odps" ]]; then
       --log_level=INFO \
       --image_pull_policy=Never \
       --output=model_output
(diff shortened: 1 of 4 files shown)


scripts/client_test.sh
```diff
@@ -57,9 +57,31 @@ elif [[ "$JOB_TYPE" == "predict" ]]; then
...
-      --job_name=test-mnist-predict \
+      --job_name=test-predict \
       --log_level=INFO \
       --image_pull_policy=Never
+elif [[ "$JOB_TYPE" == "odps" ]]; then
+    elasticdl train \
+      --image_base=elasticdl:ci \
+      --model_zoo=model_zoo \
+      --model_def=odps_iris_dnn_model.odps_iris_dnn_model.custom_model \
+      --training_data_dir=$ODPS_TABLE_NAME \
```
ywskycn, line 68: This parameter should be renamed, right?

ywskycn (approved):

scripts/client_test.sh
terrytangyuan: Yes, but probably need more people's feedback on https://github.com/sql-machine-learning/elasticdl/issues/1248 before renaming those parameters

PR #1293, opened 2019-10-10
Add function for removing last slash from parameter `model_zoo`

### LINKED ISSUE #1287: Parameter model_zoo cannot take a path end with `/`
If the parameter `model_zoo` is a path end with slash, the following exception will show up:
```
[2019-10-10 13:15:47,574] [INFO] [image_builder.py:227:_print_docker_progress]

  File "/miniconda2/envs/elasticdl-rc0/bin/elasticdl", line 10, in <module>
    sys.exit(main())
  File "/miniconda2/envs/elasticdl-rc0/lib/python3.7/site-packages/elasticdl/python/elasticdl/client.py", line 46, in main
    args.func(args)
  File "/miniconda2/envs/elasticdl-rc0/lib/python3.7/site-packages/elasticdl/python/elasticdl/api.py", line 24, in train
    docker_tlskey=args.docker_tlskey,
  File "/miniconda2/envs/elasticdl-rc0/lib/python3.7/site-packages/elasticdl/python/elasticdl/image_builder.py", line 74, in build_and_push_docker_image
    _build_docker_image(client, ctx_dir, df.name, image_name)
  File "/miniconda2/envs/elasticdl-rc0/lib/python3.7/site-packages/elasticdl/python/elasticdl/image_builder.py", line 241, in _build_docker_image
    _print_docker_progress(line)
  File "/miniconda2/envs/elasticdl-rc0/lib/python3.7/site-packages/elasticdl/python/elasticdl/image_builder.py", line 224, in _print_docker_progress
    raise RuntimeError("Docker image build: " + error)
RuntimeError: Docker image build: COPY failed: stat /var/lib/docker/tmp/docker-builder531748767/model_zoo: no such file or directory
```

ywskycn:
To fix #1287.

### DIFF
--- a/elasticdl/python/common/args.py
+++ b/elasticdl/python/common/args.py
@@ -561,3 +561,10 @@ def build_arguments_from_parsed_result(args, filter_args=None):
         "--" + k if i % 2 == 0 else k for i, k in enumerate(arguments)
     ]
     return arguments
+
+
+def remove_end_slash(val):
+    """Remove the end slash if it has"""
+    if val and val.endswith("/"):
+        return val[0:-1]
+    return val
--- a/elasticdl/python/common/args.py
+++ b/elasticdl/python/common/args.py
@@ -561,10 +561,3 @@ def build_arguments_from_parsed_result(args, filter_args=None):
         "--" + k if i % 2 == 0 else k for i, k in enumerate(arguments)
     ]
     return arguments
-
-
-def remove_end_slash(val):
-    """Remove the end slash if it has"""
-    if val and val.endswith("/"):
-        return val[0:-1]
-    return val
--- a/README.md
+++ b/README.md
@@ -2,6 +2,7 @@
 
 [![Travis-CI Build Status](https://travis-ci.org/sql-machine-learning/elasticdl.svg?branch=develop)](https://travis-ci.org/sql-machine-learning/elasticdl)
 [![License: MIT](https://img.shields.io/badge/License-MIT-blue.svg)](https://opensource.org/licenses/MIT)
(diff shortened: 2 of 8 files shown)


elasticdl/python/common/args.py
```diff
@@ -561,3 +561,10 @@ def _str_ignore_none(s):
         "--" + k if i % 2 == 0 else k for i, k in enumerate(arguments)
     ]
     return arguments
+
+
+def remove_end_slash(val):
```
terrytangyuan: Consider renaming to something like `maybe_strip_trailing_slash()`?

elasticdl/python/common/args.py
```diff
@@ -561,3 +561,10 @@ def _str_ignore_none(s):
...
+    """Remove the end slash if it has"""
+    if val and val.endswith("/"):
+        return val[0:-1]
```
terrytangyuan: `val[:-1]` should be sufficient

elasticdl/python/common/args.py
```diff
@@ -561,3 +561,10 @@ def _str_ignore_none(s):
         "--" + k if i % 2 == 0 else k for i, k in enumerate(arguments)
     ]
     return arguments
+
+
+def remove_end_slash(val):
+    """Remove the end slash if it has"""
+    if val and val.endswith("/"):
```
terrytangyuan: This can be simplified to `if val.endswith("/"):` since `"".endswith("/")` would return `False` already

terrytangyuan:
We should probably use this instead of writing our own function to normalize the path in many different cases and on different operating systems: https://docs.python.org/3/library/os.path.html#os.path.normpath

terrytangyuan (approved):

PR #1295, opened 2019-10-11
complete parameter server detailed design doc

QiJune:
[Here](https://github.com/QiJune/elasticdl/blob/new_ps_detail/docs/designs/ps_design.md) is for a better review.

Some naming convention:

- parameter: all trainable parameter in a Model
- variable: non-embedding parameter
- embedding vector:  embedding layer parameter
- embedding meta info: embedding layer parameter is initialized lazily in PS, worker need send this to PS in the first time. Embedding meta info includes: embedding vector dim, embedding vector initializer and slot variables initializer.

- pull_variable: worker pull non-embedding parameter from PS
- pull_embedding_vector: worker pull embedding parameter from PS
- push_model: worker push variable and Embedding Meta Info to PS


doc/ps_design.md
```diff
@@ -0,0 +1,382 @@
...
+        FP16 = 4;
+        FP32 = 5;
+        FP64 = 6;
+    }
+    string name = 1;
+    DataType data_type = 2;
+    repeated int64 dims = 3;
+    bytes content = 4;
+    repeated int64 indices = 5;
+}
```
QiJune: Question 1, Tensor proto definition:
-  set name to Tensor, and use `repeated Tensor param` 
- or use `map<string, Tensor> param`


I prefer to set `name` to Tensor.  Map is designed for random access. If we do not have such demand, vector is enough.

doc/ps_design.md
```diff
@@ -0,0 +1,382 @@
...
+    def __init__(self, name):
+        self.name = name
+        self.vectors = {}
+        
+    def get(self, indices):
+        pass
+        
+    def set(self, indices, value):
+        pass        
+```
```
QiJune: Question 2 EmbeddingTable structure or EmbeddingTableDB structure

KVStore has an embedding_table_db object.

```python
class EmbeddingTableDB(object):
    def get_embedding_vector(self, layer_name, slot_name, indices):
        pass

    def set_embedding_vector(self, layer_name, slot_name, indices, value):
        pass
    
    def get_incremental_replica(self):
        pass
``` 


We should take this data structure out,  the interface is determined, but we could consider better implementation later, even implemented it in C++.

doc/ps_design.md
```diff
@@ -0,0 +1,382 @@
...
+* Initialization of parameters in PS.
+* Relaunch of PS.
+* Workers get model parameters from PS.
+* Workers push gradients to PS.
+* PS reports submodel version to the master.
+* The master tells PS to save checkpoint.
+
+
+### The master starts PS
+When an ElasticDL task starts, `master.main` is responsible for starting PS. After starting PS, `master.main` starts the master and workers, and tells them the endpoints of PS.
```
QiJune: PS is started as a k8s service. So if the pserver pod is killed, and set up again, the connection of the worker is also valid.

@ywskycn  Could you please help to comment on this?

doc/ps_design.md
mhaoli: Agree.

doc/ps_design.md
```diff
@@ -0,0 +1,382 @@
...
+There are two ways for PS to get model variables and model meta info, one is to read from a checkpoint file, one is to obtain them from workers.
+
+When `master.main` starts PS, `master.main` decides how to initialize PS. If `master.main` passes an argument specifying the checkpoint file name to PS, PS reads from the checkpoint. Otherwise, PS does nothing but waiting for the first `get_model` from worker. In the reponse of `get_model` call, PS tells the worker to initialize model, and report model variables and meta info to the PS.
+
+Please Note that the worker only initializes model variables. ElasticDL adopts lazy initialization for embedding vectors. Please refer to "[Workers get model parameters from PS](#Workers-get-model-parameters-from-PS)" section.
+
+### Relaunch of PS
+In case a PS pod fails, the master will try to relaunch one PS and it should recover model variables and embedding tables.
+
+For model variables, PS can recover from workers.
```
QiJune: Worker will call `get_model`, but find that the new PS has no model on it. So the new PS will request worker to send its model.

The logic here is the same with parameter Initialization of PS.

mhaoli: Yes, that's the logic I want to say. I will modify this sentence to make things clear.

doc/ps_design.md
skydoorkai: Just define the functions, not a class. So the function implementation can be changed later.

QiJune: The functions will manipulate the data. Now, we use a simple dictionary to store all the embedding related param/slot variable etc.  The data structure inside could also be another more efficient one.
Function combines data is actually a class. I mean making the function as an interface, the definition is of the function keeps the same.

doc/ps_design.md
```diff
@@ -0,0 +1,375 @@
...
+
+So the RPC service will be defined as following:
+
+```proto
+
+service PServer{
+    rpc push_model(Model) returns (google.protobuf.Empty) {}
+    rpc pull_model(Model) returns (Model) {}
+    rpc pull_embedding_vector(Tensor) returns (EmbeddingResponse) {}
+    rpc push_gradient(Model) returns (google.protobuf.Empty)
```
QiJune: Question 3, name convention

- `compute_task` following python style
- or `ComputeTask`

If we use gRPC only with Python, personally, I think it's better to follow Python pep8 style.
There is also another discussion [here](https://github.com/grpc/grpc/issues/5014)

doc/ps_design.md
QiJune: The interface should be changed like this:

```python
class EmbeddingTableDB(object):
    def get_embedding_vector(self, layer_name, slot_name, indices, embedding_meta_info):
        pass
```
Since the embedding table is lazily initialized.

doc/ps_design.md
```diff
@@ -0,0 +1,375 @@
...
+    string name = 1;
+    repeated int64 dims = 2;
+    string initializer = 3;
+}
+
+message Model {
+    int64 version = 1;
+    repeated Tensor tensors = 2;
+    repeated EmbeddingTableInfo embedding_table_info = 3;
+}
```
QiJune: `tensors` field only contains non-embedding variable. Embedding table related meta info needs another field.

doc/ps_design.md
```diff
@@ -0,0 +1,375 @@
...
+Before each forward-pass, workers need to get all model parameters from PS. Currently, workers call function `get_model()` to get parameters.
+
+When workers want to get model parameters from PS, PS may not possess all the embedding vectors needed because ElasticDL adopts lazy initialization for embedding vectors, i.e. iniatializing embedding vectors when they are needed in workers. Thus, if a worker wants to pull some embedding vectors that are not existing in PS, PS will create and initialize these embedding vectors and return their value to the worker.
+
+### Push Gradients
+After backward-pass, workers push gradients to PS.
+
+### PS reports submodel version to the master
+The master needs to know the model version to decide when to save checkpoint and when to evaluate model. PS regularly reports the version of the submodel it possessed to the master. 
+
```
QiJune: If this is a gRPC call, please also add the corresponding gRPC service definition.

doc/ps_design.md
```diff
@@ -0,0 +1,375 @@
...
+### The master tells PS to save checkpoint
+When the master decides to save checkpoint, the master tells all the pservers to save checkpoint. Every pserver saves the submodel it possessed into a separate file.
+
+
+
+## Embedding Replicas in PS
+An ElasticDL job has *N* PSs. Embedding vectors are partitioned into these *N* PSs. The user provides *M*, the number of replicas for embedding vectors. *M* must be smaller than *N* as each PS uses other PSs to store its embedding replicas.
+
+Assume *E(i)* is the embedding vectors in PS *PS(i)*, it has *M* replicas which are stored in PSs from *P(i + 1 % N)* to *P(i + M % N)*. Also, *PS(i)* has replicas for *E(i - M % N)* to *E(i - 1 % N)*. 
+
```
QiJune: *P((i + 1) % N)* to *P((i + M) % N)* here?

skydoorkai:
(1) naming for PS
In the doc, PS and `pserver` are used. Also, as we are using multiple PSs, we need to decide what word is used for an individual PS, and what for the general PS (consisted of multiple PSs (nodes).)
One suggestion is to use PS and PS nodes. A PS has multiple nodes. Then it will be clear in the doc what we are referring to.

(2) There are two types of parameters: variables and embedding vectors. In the naming of related doc/functions, we can use variable/embedding_vector directly instead of parameters if we are referring to one of the specific parameters.

doc/ps_design.md
```diff
@@ -0,0 +1,375 @@
...
+## PServer
+
+### Compoments
+
+![pserver](./images/pserver.png)
+
+PServer contains three main compoments:
+
+- KVStore
```
skydoorkai: PS does not have a unified KVStore, but a few independent dict/KVStore for variables, embedding vectors and a few embedding replicas.

doc/ps_design.md
```diff
@@ -0,0 +1,375 @@
...
+- Gradient Queue
```
skydoorkai: Gradient Queue is for the implementation of the model updates. Only sync-SGD is used currently. It can be described in the model updates implementation, not as a component of PS

doc/ps_design.md
```diff
@@ -0,0 +1,375 @@
...
+- Optimizer
+
+1. The worker initializes a model, and pushes parameters to KVStore.
```
skydoorkai: This workflow is not needed in PS section. PS part only describes PS features (variable/embedding_vector lookup/update) and their implementation.

The next section for "Interactions among master, PS, worker" can explain the interactions.

doc/ps_design.md
skydoorkai: Who will call `get_embedding_vector`? GRPC service? If so, PS should already have `embedding_meta_info`, and we should not send it in every GRPC calls.

QiJune:
@skydoorkai 

(1) naming for PS. 

How about using `pserver` for general PS, and `PS node/or PS nodes` for concrete a pserver node?

(2) naming for parameter

How about using `variable` and `embedding_table`?

The model is defined as the following:

```proto
message EmbeddingTableInfo {
    string name = 1;
    repeated embedding_vector_dims = 2;
    string initializer = 3;
}

message Model {
    int64 version = 1;
    repeated Tensor variables = 2;
    repeated EmbeddingTableInfo embedding_tables = 3;
}
```

doc/ps_design.md
skydoorkai: *PS(i)* has *E(i - M % N)* to *E(i - 1 % N)* (the replicas for P(i-M%N) to P(i-1%N)). 
Need to rephrase this sentence.

skydoorkai (commented):
> @skydoorkai
> 
> (1) naming for PS.
> 
> How about using `pserver` for general PS, and `PS node/or PS nodes` for concrete a pserver node?
> 
> (2) naming for parameter
> 
> How about using `variable` and `embedding_table`?
> 
> The model is defined as the following:
> 
> ```protobuf
> message EmbeddingTableInfo {
>     string name = 1;
>     repeated embedding_vector_dims = 2;
>     string initializer = 3;
> }
> 
> message Model {
>     int64 version = 1;
>     repeated Tensor variables = 2;
>     repeated EmbeddingTableInfo embedding_tables = 3;
> }
> ```

GetModel will only get variables, not embedding vectors. `embedding_tables` should not be included.

QiJune:
> GetModel will only get variables, not embedding vectors. `embedding_tables` should not be included.

Yes, but PushModel needs to push variables and embedding meta info.  A direct thought, the model is composed of two parts, variables and embedding_tables. The embedding table field could be empty when GetModel.

doc/ps_design.md
QiJune: Yes, the KVStore is just a composed data structure, it's a very thin wrapper.

doc/ps_design.md
QiJune: I mean that operator priority here, `%` is higher than `-`.  So it should be ((i + 1) % N) and ((i + M) % N)

doc/ps_design.md
QiJune: Not GRPC service. It's used by Optimizer. The optimizer will get embedding vectors from the DB to do updating.

doc/ps_design.md
QiJune: Okay, I will put gradient queue to Optimizer part.

doc/ps_design.md
QiJune: This is only a simple introduction, trying to let users get a better understanding of the role of PS. I will make it more simpler.

doc/ps_design.md
```diff
@@ -0,0 +1,374 @@
...
+### KVStore
+
+For a common model variable, we use save it as a `tf.Variable` in Parameter DB. For the embedding table, we introduce a customized data structure.
+
+
+```python
+class EmbeddingTable(object):
+    def __init__(self, name):
+        self.name = name
+        self.vectors = {}
```
QiJune: Add:
```python
self.meta_info = meta_info
```

doc/ps_design.md
```diff
@@ -0,0 +1,374 @@
...
+
+
+
+So the RPC service will be defined as following:
+
+```proto
+
+service PServer{
+    rpc push_model(Model) returns (google.protobuf.Empty) {}
+    rpc pull_model(Model) returns (Model) {}
```
QiJune: rename `pull_model` to `pull_variables`

doc/ps_design.md
```diff
@@ -0,0 +1,337 @@
...
+
+So the RPC service will be defined as following:
+
+```proto
+
+service PServer{
+    rpc push_model(Model) returns (google.protobuf.Empty) {}
+    rpc pull_variable(Model) returns (Model) {}
+    rpc pull_embedding_vector(Tensor) returns (EmbeddingResponse) {}
+    rpc push_gradient(Model) returns (google.protobuf.Empty)
```
skydoorkai: push_gradient has a non-empty response. Sync-SGD may reject gradients.

doc/ps_design.md
```diff
@@ -0,0 +1,337 @@
...
+![pserver](./images/pserver.png)
+
+PS contains two main compoments:
+
+- KVStore
+- Optimizer
+
+The worker initializes a model, and pushes parameters to KVStore. Before each step of training, the worker pulls the latest model from KVStore. After a round of forward/backward computation, the worker pushes gradients to the PS waiting for processing. Then, optimizer of PS will look up the corresponding parameter from KVStore. At last, it applies gradients to parameters, and updates parameter back to KVStore.
+
+### Tensor Data Structure
```
skydoorkai: Use a top-down approach to describe PS. `KVStore` first, then `EmbeddingTable`, then `Tensor`.
Bottom-up apporach to start with a detail will make the reader get lost.

skydoorkai:
Need to change the names in the image.

doc/ps_design.md
QiJune: Done

doc/ps_design.md
QiJune: Done

docs/designs/ps_design.md
```diff
@@ -0,0 +1,429 @@
...
+### Workers get model variables from PS
+Before each forward-pass, workers need to get all model variables from PS. Currently, workers call function `get_model()` to get variables.
+
+Workers get embedding vectors from PS when the forward-pass function of the ElasticDL embedding layer is called. PS may not possess all the embedding vectors needed because ElasticDL adopts lazy initialization for embedding vectors, i.e. iniatializing embedding vectors when they are needed. Thus, if a worker wants to pull some embedding vectors that are not existing in PS, PS will create and initialize these embedding vectors and return their value to the worker.
+
+```python
+message EmbeddingTable{
+    string name;
+    Tensor embeddings;
+}
```
QiJune: There is a name field in Tensor already,  the name could be `embedding_layer0` or `embedding_layer0-momentum`. So, there is no need to add an extra EmbeddingTable message

docs/designs/ps_design.md
```diff
@@ -0,0 +1,422 @@
...
+If `master.main` passes an argument specifying the checkpoint file name to PS, PS reads from the checkpoint. Every PS node scans the variables and embedding vectors in the checkpoint, and saves the part which belong to it after parameter sharding.
+
+Otherwise, PS obtains model variables and model meta info from workers. PS does nothing but waiting for the first `pull_variable` call from worker. In the reponse of `pull_variable` call, PS tells the worker to initialize model, and report model variables and meta info to the PS. This process can be represented in the pseudocode:
+
+```python
+message PullModelRequest{
+    enum MethodType {
+        MINIMUM = 0;
+        FIXED = 1;
+    }
```
QiJune, line 261: Not very sure what's the usage of the `MethodType`,  the definition is copied from the current code base.

docs/designs/ps_design.md
```diff
@@ -0,0 +1,424 @@
...
+
+Each PS will provide a gRPC service for the replica synchronization.
+
+```
+message SynchronizeEmbeddingRequest {
+    int32 replica_index = 1;
+}
+
+message SynchronizeEmbeddingResponse {
+    map<string, Tensor> embedding_vectors = 1;
```
QiJune, line 382: The is a `name` field in `Tensor`, so we may use

```proto
repeated Tensors embedding_vectors = 1;
```

wangkuiyi (approved):
Let us merge this design. And please @QiJune organize a meeting sharing the design and raise discussions.

docs/designs/ps_design.md
mhaoli: OK. I will modify this proto after discussion.

docs/designs/ps_design.md
mhaoli: OK. I will modify this proto after discussion.

PR #1296, opened 2019-10-11
Report per task

### LINKED ISSUE #1259: Evaluation minibatches may be calculated repeatedly
Currently, each task contains multiple minibatches. In the evaluation phase, workers calculate each minibatch in turn. Every time a minibatch calculation is completed, the worker immediately passes the calculated evaluation metrics to the master. When the worker completes all the minibatches in a task, the worker tells the master the results of the entire task.
If a worker A is killed in the process of calculating a task, the master will find another worker B to calculate the failed task, but it will not clear the metrics of the minibatch already calculated in the task. So some minibatches are calculated both by worker A and worker B, which will lead to errors in the final evaluation metrics.

chunyang-wen:
fixes #1259 
+ report evaluation metrics by task instead of by batch

### DIFF
--- a/elasticdl/python/common/constants.py
+++ b/elasticdl/python/common/constants.py
@@ -37,3 +37,4 @@ class Redis(object):
 
 class MetricsDictKey(object):
     MODEL_OUTPUT = "output"
+    LABEL = "label"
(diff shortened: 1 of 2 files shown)


mhaoli (approved):
LGTM

PR #1298, opened 2019-10-11
Fix odps test in data_reader_test.py

mhaoli:
Keras model requires model input to have shape `[batch_size, feature_shape]`, even if batch size equals to 1. However, `test_odps_data_reader_integration_with_local_keras` in `tests/data_reader_test.py` generate features with shape `[feature_shape]`.


terrytangyuan (approved):
Thanks!

PR #1270, opened 2019-10-04
Update TensorFlow to official 2.0.0 release

terrytangyuan:
Signed-off-by: terrytangyuan <terrytangyuan@gmail.com>

PR #1301, opened 2019-10-12
Add write lock in optimizer wrapper

mhaoli:
This is a follow-up PR of #1270. 

#1270 used a temporary solution to fix conflict when updating TensorFlow to 2.0.0: setting some useless variables to None. This PR adds a lock and deletes them from the list.


skydoorkai (approved):
LGTM

Issue #1303, opened 2019-10-13
Define wide&deep model  and train with ODPS table


workingloong:
One format of ODPS table is as followd:

| dense_feature |sparse_feature  | label |
| ------ | ------ | ------ |
| "1.3,0.5,0.7,0.9" | "1 4 6 "| 1 |
| "2.3,0.2,1.7,-0.9" | "1 7 9 13 56" | 0 |

The dense_feature column are float fixed-length arrays and is saved as string with separator. The sparse_feature column are int variable-length arrays and is also saved as string with separator . The dataset_fn should parse dense_feature to tensor and sparse_feature to sparseTensor.

PR #1304, opened 2019-10-13
Define wide&deep model to train with odps table

workingloong:
Define wide&deep classification model and dataset_fn. The dataset_fn can parse FElib output data to generate Dataset. #1303


terrytangyuan:
Is the dataset somewhere public? How could users run this example by themselves? If this example doesn't take too long to run, we could probably add it to integration test as well similar to our dummy iris model.

Issue #1302, opened 2019-10-12
dtype of column in odps table may be not float32


workingloong:
The dtype of column in odps table may be int32, float, boolean, string and so on. So, the records_output_types in ODPSDataReader can not be fixed to tf.float32.
https://github.com/sql-machine-learning/elasticdl/blob/aef9d66e5d99ed2144ed4ba36933aaf85738a327/elasticdl/python/data/data_reader.py#L143-L144

I suggest that, records_output_types in ODPSDataReader is fixed to tf.string. And, we should convert data from odps_io.ODPSReader.read_batch to string.
https://github.com/sql-machine-learning/elasticdl/blob/aef9d66e5d99ed2144ed4ba36933aaf85738a327/elasticdl/python/data/odps_io.py#L223-L225

```
batch_record.append(
    [str(record[column]) for column in columns]
)
```
Then, user can cast the string to data type they want in defined dataset_fn

terrytangyuan:
@workingloong Yes, `float32` was only the temporary plan. We should switch to something more robust.

PR #1307, opened 2019-10-14
Modify data type to tf.string in ODPS reader

workingloong:
Create tf.data.Dataset with dtype tf.string and cast string to float32 in dataset_fn for odps_iris_dnn_model. https://github.com/sql-machine-learning/elasticdl/issues/1302

### DIFF
--- a/elasticdl/python/data/data_reader.py
+++ b/elasticdl/python/data/data_reader.py
@@ -141,7 +141,7 @@ class ODPSDataReader(AbstractDataReader):
 
     @property
     def records_output_types(self):
-        return tf.float32
+        return tf.string
 
     @property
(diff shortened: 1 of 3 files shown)


terrytangyuan (commented):
Have you tested this change? (Note that the ODPS related tests will only run on Travis when you are developing from branches instead of forks.

workingloong:
> Have you tested this change? (Note that the ODPS related tests will only run on Travis when you are developing from branches instead of forks.

Yes, I have tested this change in minikube for iris_dnn_model with my owner's odps table.

workingloong:
I have created new pr #1309 from branches, so I close the pr.

PR #1313, opened 2019-10-15
polish ps design

QiJune:
[Here](https://github.com/sql-machine-learning/elasticdl/blob/ps_design_v0/docs/designs/ps_design.md) is for a better review.


docs/designs/ps_design.md
```diff
@@ -1,375 +1,191 @@
...
+In order to support the scalability of model size and PS fault tolerance in ElasticDL, ElasticDL needs a distributed parameter server. 
+
+If a model has a very large size, it may not fit in a single parameter server (PS) memory. For example, many recommending and ranking models take very high-dimensional (and sparse) inputs, thus require large embedding tables. Processing all model parameters in a single PS requires huge computation and I/O bandwidth when the model size becomes very large. By distributing the model parameters into multiple PS pods, a distributed PS can solve the memory, the computation and the I/O bandwidth issues when scaling up the model size.
+
+The master can monitor PS pods status similar to what it does for worker pods. In case a PS pod fails, the master will relaunch it. Since PS pods have higher priority than worker pods, if there are still some running worker pods, the relaunch will succeed by using either idle or preempted Kubernetes resources.
```
chunyang-wen: pods' status
a higher priority
the worker pods

docs/designs/ps_design.md
```diff
@@ -1,375 +1,191 @@
...
+
+The master can monitor PS pods status similar to what it does for worker pods. In case a PS pod fails, the master will relaunch it. Since PS pods have higher priority than worker pods, if there are still some running worker pods, the relaunch will succeed by using either idle or preempted Kubernetes resources. 
+
+After the relaunch of a PS pod, the PS pod needs to recover the model parameters distributed in it. There are two kinds of model parameters in ElasticDL.
 
-The master will monitor PS nodes status similar to what it does for workers. In case a PS node fails, the master will try to relaunch it. Since PS nodes have higher priority than workers, if there are still some running workers, the relaunch will succeed by using either idle or preempted Kubernetes resources. If the relaunch fails, there are no workers left. The Elastic job has to wait for resources for the PS node and workers.
+* variable: trainable TensorFlow variable defined in a Keras model.
+* embedding vector: embedding vector in an [embedding layer](../../elasticdl/python/elasticdl/layers/embedding.py)
```
chunyang-wen, line 14: The embedding vector is a special kind of variable. Is it suitable to differentiate them from normal variables? If you want to emphasize the size of the embedding table, maybe we can change the description here to describing why the model size keeps increasing.

docs/designs/ps_design.md
```diff
@@ -1,375 +1,191 @@
...
 
 ## PS
 
-### Compoments
+We will distribute model parameters into multiple PS pods, which is called parameter sharding. There is a hash function that maps a parameter to a PS pod id. For a variable, the key of hash function is its name. For a embedding vector, the key is its embedding layer name combining an item id. We could use a simple round-robin policy, *PS(i)=hash(key) mod N*, at first. Each PS pod only holds a subset of the whole model.
 
-![pserver](../images/pserver.png)
+We use a KVStore to store model parameters in PS. There is a KVStore instance in each PS pod. The KVStore instances from the PS pods forms a distributed KVStore, which could be scaled easily to support a model with a large size.
```
chunyang-wen: forms->form

docs/designs/ps_design.md
```diff
@@ -1,375 +1,191 @@
...
 
-PS contains two main compoments:
+We also need an optimizer to update the model parameters stored in PS pods. To update a single parameter, the optimizer needs to get the parameter, apply the gradient to the parameter, and then write the parameter back. It involves one time read and one time write. There will be huge accesses to parameters during a training job. Since model parameters are store at PS, it's better to make the optimization at the same place to reduce the cost of accessing parameters.
 
-- KVStore
-- Optimizer
+Besides KVStore and optimizer, the PS also need to provide necessary RPC services to workers. Workers will pull the latest model parameters from PS, and push gradients to PS in each iteration of training.
```
chunyang-wen: need->needs
Besides [the] KVStore and optimizer?

docs/designs/ps_design.md
```diff
@@ -1,375 +1,191 @@
...
 
-The worker initializes a model, and pushes parameters to KVStore. Before each step of training, the worker pulls the latest model from KVStore. After a round of forward/backward computation, the worker pushes gradients to the PS waiting for processing. Then, optimizer of PS will look up the corresponding parameter from KVStore. At last, it applies gradients to parameters, and updates parameter back to KVStore.
+Thus, we propose that PS has three basic components:
```
chunyang-wen: [a] PS ?

docs/designs/ps_design.md
```diff
@@ -1,375 +1,191 @@
...
-        pass
+### KVStore
 
-    def get_embedding_vector(self, name, indices):
-        pass
+The KVStore needs to support both embedding vector parameter and non-embedding parameter.
 
-    def set_parameter(self, name, value):
-        pass
+Since `tf.keras.optimizer` only accept `tf.Variable` as the parameter. To avoid unnecessary memory copy, we save non-embedding parameter as a `tf.Variable` in variable DB directly.
```
chunyang-wen, line 48: What do you mean by *variable DB*?

docs/designs/ps_design.md
```diff
@@ -1,375 +1,191 @@
...
 
     def set(self, indices, value):
         pass
 ```
 
-The name of embedding table is the embedding layer name. EmbeddingTable uses a dictionary `vectors` to store `<id, embedding_vector>` pairs.
+The name is the embedding layer name of embedding vectors. It uses a dictionary `vectors` to store embedding vector.
 
-Since embedding table is lazily initialized in PS, it also has `dim` and `initializer` fields.
+Since embedding vectors are lazily initialized in PS, it also has `dim` and `initializer` fields. Inside the `get` interface of `EmbeddingVectors `, if the id is not in the `vectors` dictionary, the corresponding value will be initialized and return back.
```
chunyang-wen, line 81: return back?

chunyang-wen: Maybe we should introduce variable db and embedding db first?

docs/designs/ps_design.md
```diff
@@ -1,375 +1,191 @@
...
-
-In the first way, there may be several gRPC threads running in parallel. This will introduce race condition on parameter updating, some gradients may be overwrited.
-
-The second way ensure each gradient could be applied, and decoupling these two procedures, `push_gradient` of worker and `apply_gradient` of optimizer. But the second way introduces more staleness in updating model, and may influence the final training accuracy.
-
-We may consider the second way later.
+### Optimizer
 
-In sync-SGD, optimizer needs to wait for a certain number of gradients, and then get the gradient after addition. We could implement a customized gradient queue structure to support such logic efficiently.
+The optimizer of PS is responsible for applying gradients to parameters in KVStore. Embedding table parameter needs to be handled carefully, since its not a standard `tf.Variable`. We have already implemented an [OptimizeWrapper](https://github.com/sql-machine-learning/elasticdl/blob/develop/elasticdl/python/master/optimizer_wrapper.py) to handle this. We will move it to from master to pserver part.
```
chunyang-wen: its->it is

docs/designs/ps_design.md
```diff
@@ -1,375 +1,191 @@
...
 
-The master will create a PS with *N* PS nodes, where *N* is specified by the user. Each model variable and embedding vector has a corresponding PS node. Thus, every PS node has a subset of model variables and embedding tables.
+In order to support the scalability of model size and PS fault tolerance in ElasticDL, ElasticDL needs a distributed parameter server. 
+
+If a model has a very large size, it may not fit in a single parameter server (PS) memory. For example, many recommending and ranking models take very high-dimensional (and sparse) inputs, thus require large embedding tables. Processing all model parameters in a single PS requires huge computation and I/O bandwidth when the model size becomes very large. By distributing the model parameters into multiple PS pods, a distributed PS can solve the memory, the computation and the I/O bandwidth issues when scaling up the model size.
```
terrytangyuan: single parameter server (PS) memory -> memory of a single parameter server

docs/designs/ps_design.md
```diff
@@ -1,375 +1,191 @@
...
-    def __init__(self):
-        self.grad_queue = queue.Queue()
-    
-    def get_gradient(self):
-        pass
-        
-    def put_gradient(self):
-        pass
-```
+The optimizer supports two kinds of parameter updating strategies: sync-SGD and async-SGD. In sync-SGD, the optimizer needs to wait for a certain number of gradients, and then apply the gradients to parameters. In async-SGD, the  `apply_gradient` function of optimizer inside will be called inside `push_gradient` RPC service directly.
```
chunyang-wen, line 105: Use a table or unordered list to describe these two kinds of updating strategies.

docs/designs/ps_design.md
```diff
@@ -1,375 +1,191 @@
...
 
 ### RPC Service
 
-PS provides RPC service for workers.
-
-Since each PS node will store a subset of the whole model. A worker will push/pull submodels to/from all PS nodes.
+PS provides RPC service for workers. There are three important events that PS should interact with workers:
```
chunyang-wen, line 109: The actions seem to belong to the workers. But here we are introducing the events of PS. A little confusing here.

docs/designs/ps_design.md
```diff
@@ -1,375 +1,191 @@
...
-    string initializer = 3;
-}
-
-message Model {
-    int64 version = 1;
-    repeated Tensor variables = 2;
-    repeated EmbeddingTableInfo embedding_table_info = 3;
-}
-```
+After starting, PS does not contain any parameter. For model variables, ElasticDL should initialize them before training process. For embedding vectors, ElasticDL adopts lazy initialization, i.e. initialize them when they are needed in training process.
```
chunyang-wen: If we restore from a pre-trained model, the PSs do contain parameters.
lazy initialization-> a lazy initialization
training process -> the training process

docs/designs/ps_design.md
```diff
@@ -1,375 +1,191 @@
...
+In order to support the scalability of model size and PS fault tolerance in ElasticDL, ElasticDL needs a distributed parameter server. 
+
+If a model has a very large size, it may not fit in a single parameter server (PS) memory. For example, many recommending and ranking models take very high-dimensional (and sparse) inputs, thus require large embedding tables. Processing all model parameters in a single PS requires huge computation and I/O bandwidth when the model size becomes very large. By distributing the model parameters into multiple PS pods, a distributed PS can solve the memory, the computation and the I/O bandwidth issues when scaling up the model size.
+
+The master can monitor PS pods status similar to what it does for worker pods. In case a PS pod fails, the master will relaunch it. Since PS pods have higher priority than worker pods, if there are still some running worker pods, the relaunch will succeed by using either idle or preempted Kubernetes resources. 
+
+After the relaunch of a PS pod, the PS pod needs to recover the model parameters distributed in it. There are two kinds of model parameters in ElasticDL.
```
terrytangyuan: "." -> ":"

docs/designs/ps_design.md
```diff
@@ -1,375 +1,191 @@
...
 
-The master will monitor PS nodes status similar to what it does for workers. In case a PS node fails, the master will try to relaunch it. Since PS nodes have higher priority than workers, if there are still some running workers, the relaunch will succeed by using either idle or preempted Kubernetes resources. If the relaunch fails, there are no workers left. The Elastic job has to wait for resources for the PS node and workers.
+* variable: trainable TensorFlow variable defined in a Keras model.
+* embedding vector: embedding vector in an [embedding layer](../../elasticdl/python/elasticdl/layers/embedding.py)
+
+Each worker has a local copy of all variables. The relaunched PS pod can recover variables from any of the workers. For embedding vectors, PS must create replicas to support fault tolerance. Each PS pod has its embedding vector replicas stored in other PS pods. The relaunched PS pod can recover embedding vectors from its replicas.
```
terrytangyuan: embedding vectors are variables as well so the "all variables" is a bit confusing

docs/designs/ps_design.md
```diff
@@ -1,375 +1,191 @@
...
-message Model {
-    int64 version = 1;
-    repeated Tensor variables = 2;
-    repeated EmbeddingTableInfo embedding_table_info = 3;
-}
-```
+After starting, PS does not contain any parameter. For model variables, ElasticDL should initialize them before training process. For embedding vectors, ElasticDL adopts lazy initialization, i.e. initialize them when they are needed in training process.
 
-So the RPC service will be defined as following:
+Since a single PS pod may not have enough memory for big model, workers are responsible for random initializing model variables. After initializing, workers push initialized model variables to corresponding PS pod. Here is the RPC call definition for pushing initialized model.
```
chunyang-wen, line 119: I don't understand the statements here. Why the workers take the responsibility of initializing can solve the memory problem of a single PS?

docs/designs/ps_design.md
terrytangyuan: Define KVStore and perhaps mention what technology will be used for this

docs/designs/ps_design.md
```diff
@@ -1,375 +1,191 @@
...
 
+For model variables, a worker needs to pull model variables from all PS pods before the forward-pass.
 
-```python
-class PServer(elasticdl_pb2_grpc.PServerServicer):
-    def __init__(self, kvstore, grad_queue, opt):
-        self.kvstore = KVStore()
-        self.grad_queue = GradientQueue()
-        self.opt = Optimizer(opt, self.kvstore, self.grad_queue)
+For embedding vectors, ElasticDL should only pull embedding vectors that are used in this iteration. This is because embedding vectors used in each iteration only account for a small proportion of the embedding tables. Only when the `call` function of embedding layer is called do we know which embedding vectors will be used in this function. Thus, embedding layer is responsible for pulling embedding vectors from PS in its `call` function.
```
chunyang-wen: Multiple places need a/an/the. Please double check the doc.

docs/designs/ps_design.md
```diff
@@ -1,375 +1,191 @@
...
+We will distribute model parameters into multiple PS pods, which is called parameter sharding. There is a hash function that maps a parameter to a PS pod id. For a variable, the key of hash function is its name. For a embedding vector, the key is its embedding layer name combining an item id. We could use a simple round-robin policy, *PS(i)=hash(key) mod N*, at first. Each PS pod only holds a subset of the whole model.
 
-![pserver](../images/pserver.png)
+We use a KVStore to store model parameters in PS. There is a KVStore instance in each PS pod. The KVStore instances from the PS pods forms a distributed KVStore, which could be scaled easily to support a model with a large size.
 
-PS contains two main compoments:
+We also need an optimizer to update the model parameters stored in PS pods. To update a single parameter, the optimizer needs to get the parameter, apply the gradient to the parameter, and then write the parameter back. It involves one time read and one time write. There will be huge accesses to parameters during a training job. Since model parameters are store at PS, it's better to make the optimization at the same place to reduce the cost of accessing parameters.
```
terrytangyuan, line 27: "make the optimization at the same place" -> what do you mean by "the same place"?

docs/designs/ps_design.md
```diff
@@ -1,375 +1,191 @@
...
-- KVStore
-- Optimizer
+Besides KVStore and optimizer, the PS also need to provide necessary RPC services to workers. Workers will pull the latest model parameters from PS, and push gradients to PS in each iteration of training.
 
-The worker initializes a model, and pushes parameters to KVStore. Before each step of training, the worker pulls the latest model from KVStore. After a round of forward/backward computation, the worker pushes gradients to the PS waiting for processing. Then, optimizer of PS will look up the corresponding parameter from KVStore. At last, it applies gradients to parameters, and updates parameter back to KVStore.
+Thus, we propose that PS has three basic components:
 
+| component| functionality |hardware |
+| :----  |:----  |:----  |
+|KVStore  | provides distributed storage of model parameters|memory|
```
terrytangyuan, line 35: add "storage" besides "memory"?

docs/designs/ps_design.md
```diff
@@ -1,375 +1,191 @@
...
 
-For model variables, PS can recover from workers in the same way as the variable initialization.
+## PS Fault Tolerance
 
-For embedding tables, the `master.main` tells PS through in starting command that PS should recover from replica. If there is no replica, PS has to recover from checkpoint. More details can be find in section [Embedding Replicas in PS](#Embedding-Replicas-in-PS).
+We support PS fault tolerance by relaunching any failed PS pod and recovering its model parameters. 
 
-### Workers get model variables from PS
-Before each forward-pass, workers need to get all model variables from PS. Currently, workers call function `get_model()` to get variables.
+The master will create a distributed PS with *N* PS pods, where *N* is specified by the user. In case a PS pod fails, the master will relaunch it by Kubernetes APIs. As we discussed in [Overview](#overview), the relaunch will succeed as long as there are still running worker pods.
```
chunyang-wen, line 166: How to handle the failing PSs during applying gradients?

docs/designs/ps_design.md
```diff
@@ -1,375 +1,191 @@
...
-    def set_parameter(self, name, value):
-        pass
+Since `tf.keras.optimizer` only accept `tf.Variable` as the parameter. To avoid unnecessary memory copy, we save non-embedding parameter as a `tf.Variable` in variable DB directly.
 
-    def set_embedding_vector(self, name, indices, value):
-        pass
-```
+For the embedding vector parameter, there are two reasons that we could not initialize it before training:
+
+- Feature ID is usually calculated by the hash function. The training dataset is too large to get the range of the ID before training.
```
terrytangyuan, line 52: Define "Feature ID"

docs/designs/ps_design.md
```diff
@@ -1,375 +1,191 @@
...
+- Sometimes, there comes a new feature ID in online learning.
```
terrytangyuan, line 53: You mean the set of feature IDs cannot be defined at once and is growing?

docs/designs/ps_design.md
```diff
@@ -1,375 +1,191 @@
 # Parameter Server Design
 
-
 ## Overview
-Currently, there is one parameter server (PS) co-existed with the master. In order to support PS with multiple nodes and PS fault tolerance, we need to separate PS from the master. Besides a KV store for model parameters, PS should also support updating parameters using gradients. Currently, models have two kinds of parameter, variable and embedding vector. PS should support model variable update and embedding table sparse update.
 
-The master will create a PS with *N* PS nodes, where *N* is specified by the user. Each model variable and embedding vector has a corresponding PS node. Thus, every PS node has a subset of model variables and embedding tables.
+In order to support the scalability of model size and PS fault tolerance in ElasticDL, ElasticDL needs a distributed parameter server.
```
wangkuiyi, line 5: In order to support PS fault tolerance, so we need a distributed PS -- this reasoning seems reversed.

docs/designs/ps_design.md
wangkuiyi: Big models don't fit in the memory of a single PS process. Then what?

docs/designs/ps_design.md
wangkuiyi: What's the relationship between this paragraph and the previous one? This paragraph mentions the concept *master* abruptly without any introduction in the previous paragraph.

docs/designs/ps_design.md
wangkuiyi: It doesn't sense for each worker has a local copy of all variables, because this document mentioned earlier that the model parameters could be too big to fit in the memory of a single PS process.

wangkuiyi (commented):
There seem too many logical loopholes in this version. Is this a WIP?

docs/designs/ps_design.md
brightcoder01: from any of the workers -> from any worker or from any one of the workers

docs/designs/ps_design.md
```diff
@@ -1,375 +1,191 @@
...
-    string name = 1;
-    DataType data_type = 2;
-    repeated int64 dim = 3;
-    bytes content = 4;
-    repeated int64 indices = 5;
-}
-```
-
-Correspondingly, we have a `Tensor` Python class.
+There could be multiple embedding vectors from different embedding layer. We save all these embedding vectors in an independent embedding vector DB of KVStore.
```
brightcoder01, line 83: It means the KVStore contains multiple embedding vectors of different embedding layer.
How does KVStore figure out which embedding vector group to look up?

docs/designs/ps_design.md
```diff
@@ -1,375 +1,191 @@
...
-        pass
+Here are the RPC call definitions for pulling model variables and pulling embedding vectors.
 
-    # embedding param is handled lazily
-    def pull_embedding_vector(self, reques, _):
-        pass
+```proto
+service PServer{
+    rpc pull_variable(PullModelRequest) returns (PullModelResponse);
+    rpc pull_embedding_vector(Tensor) returns (Tensor);
```
brightcoder01: Tensor -> Where do we define this object? Is it the **Tensor** proto from tensorflow?

docs/designs/ps_design.md
```diff
@@ -1,375 +1,191 @@
...
 }
 ```
 
-For the details of RPC service definition, please refer to next section.
+**Pull model parameters**
 
-The interfaces of PServer could be like this:
+Since ElasticDL saves parameters in PS, workers should pull parameters from PS in each iteration of training/evaluation process. 
 
+For model variables, a worker needs to pull model variables from all PS pods before the forward-pass.
```
brightcoder01, line 131: Does it mean for each iteration, each worker will make multiple PRC calls to each PS code to get the non-embedding variables? I'm wondering it will cost much time to do it.

docs/designs/ps_design.md
mhaoli: Because a single PS pod only contains a subset of model parameters, while each worker contains a complete model. Thus the memory of single PS pod may be smaller than the memory of single worker pod.

docs/designs/ps_design.md
QiJune: Here we have two kinds of parameters, one is non-embedding parameter, the other is embedding parameter. Non-embedding parameter could be represented by a standard `tf.Variable`, but embedding parameter could not.

So, we limit variable to only non-embedding parameters in this doc.

docs/designs/ps_design.md
QiJune: In the same pod. I will fix it.

docs/designs/ps_design.md
QiJune: The parameters are stored in memory, for fast access. If we want to make a checkpoint, we will dump data in KVStore from memory to storage.

docs/designs/ps_design.md
QiJune: variable DB is a simple dictionary, the key is the name of a `tf.Variable`, the value is the `tf.Variable` itself.

docs/designs/ps_design.md
QiJune: Yes, so the size embedding table of feature ID could not be decided. We could not take it as a common `tf.Variable`, we need a customized data structure to handle it.

docs/designs/ps_design.md
QiJune: There are two steps: 

- give embedding layer name, `KVStore` returns a `EembeddingVectors` object
- give ids info, `EembeddingVectors` returns embedding vectors

docs/designs/ps_design.md
skydoorkai: The overall data transmitted does not change. Multiple RPC calls can be called concurrently from a thread pool. This will take less time.

PR #1316, opened 2019-10-16
re-polish ps design by QiJune

QiJune:
[Here](https://github.com/QiJune/elasticdl/blob/ps_design_v1/docs/designs/ps_design.md) is for a better review.


docs/designs/ps_design.md
```diff
@@ -1,53 +1,62 @@
...
+Sharding the model to multiple parameter servers is a common choice. The reasons are as following:
 
-## PS
+- Sometimes, the model size if out of the memory of a process. For example, many recommending and ranking models take very high-dimensional (and sparse) inputs, thus require large embedding tables.
+- Communication load between workers and parameter servers will be balanced. More parameter servers usually means more network bandwidth.
+- Model optimization could be parallelized on multiple parameter servers.
 
-### Compoments
+In addition, we want to launch machine learning jobs in a Kubernetes supported cluster. The pods in Kubernetes are scheduled with priority, and could be preempted at any time. At the same time, hardware failure problem is also nonnegligible in a large scale distributed system. Here, we introduce another role called master to watch the status of worker pods and parameter server pods. It calls Kubernetes API to start or relaunch worker pods and optionally parameter server pods, and scheduling tasks to worker pods.
```
wangkuiyi: I understand this paragraph has an intention to explain the existence of the master process in a job. To clearly explain this topic, the author needs to know more about Kubernetes-native architecture and its successor, the Kubernetes Operator-based design strategy. Otherwise, it is difficult to explain why we need a master process, other than a Kubernetes Operator.

Again, do we need to explain why we need a master here?

docs/designs/ps_design.md
```diff
@@ -1,53 +1,62 @@
 # Parameter Server Design
 
-
 ## Overview
-Currently, there is one parameter server (PS) co-existed with the master. In order to support PS with multiple nodes and PS fault tolerance, we need to separate PS from the master. Besides a KV store for model parameters, PS should also support updating parameters using gradients. Currently, models have two kinds of parameter, variable and embedding vector. PS should support model variable update and embedding table sparse update.
 
-The master will create a PS with *N* PS nodes, where *N* is specified by the user. Each model variable and embedding vector has a corresponding PS node. Thus, every PS node has a subset of model variables and embedding tables.
+A typical parameter server architecture of data parallelism distributed training system contains two roles:
```
wangkuiyi: It seems that enumeration prevents us from explaining the relationship between the bullets -- workers and parameter servers.

> Many distributed training systems use multiple works and the data parallelism architecture. Data parallel means that each worker computes the opinion about how to update the model, say, gradients, from its local data. So we need a mechanism to help workers exchange their opinions. The parameter server is such an idea for gradient aggregation and model updates.

docs/designs/ps_design.md
```diff
@@ -1,53 +1,62 @@
...
+- parameter server(or PS for short), maintaining the global model, and providing parameter pull/push/optimize/checkpoint service
+- worker, computing gradients of model parameters using local data and pushing gradients to parameter servers.
 
-Each worker has a local copy of the model variables. After the master relaunches a PS node, the PS node can recover model variables from workers. For embedding vectors, PS must create replicas to support fault tolerance. For each PS node *PS(i)*, it will store *M* replicas in the following *M* PS nodes from *PS(i+1 % N)* to *PS(i+M % N)*. The relaunched PS node can recover embedding vectors from one of its replicas. If there are more than *M* continuously-indexed PS nodes failing, at least one PS node fails with all of its replicas. The ElasticDL job has to recover from a recent checkpoint.
+Sharding the model to multiple parameter servers is a common choice. The reasons are as following:
```
wangkuiyi: Again, the enumeration here prevents us from explaining the ideas progressively. Please be aware that these ideas are not on the same logic level.  A single-node parameter server architecture is a bottleneck of communication and computation anyway, in spite of the model size.  So it would be easier to say:

> Many systems use multiple parameter server instances in a job, especially when the model could be large and overruns the memory space of a process. In such cases, we can shard the model and distribute the pieces to multiple parameter server instances. Even if the model is not too big, model sharding prevents the case that a single-node parameter server architecture becomes the bottleneck of communication and computation.

docs/designs/ps_design.md
QiJune: @wangkuiyi Since we want to keep the parameter server fault-tolerant, once a PS pod is killed, we have to restore it. The master here is responsible for creating a new PS pod. It's hard to skip the introduction of the master role.

I will add another link to the readme doc of our project, https://github.com/sql-machine-learning/elasticdl#background . There are more details of why we introduce  a master role in the readme doc.

docs/designs/ps_design.md
QiJune: Done

docs/designs/ps_design.md
QiJune: Done

docs/designs/ps_design.md
```diff
@@ -58,322 +67,202 @@ class EmbeddingTable(object):
...
-        res = PullModelResponse()
-        if self._need_initialize_model:
-            res.need_push_model = True
-            return res
-        res.need_push_model = False
-        res.model = self._get_model() # get model in this PS node
-        return res
+| component| functionality |hardware resource|
+| :----  |:----  |:----  |
+|KVStore  | provides distributed storage of model parameters|memory|
```
QiJune: Remove KVStore in the graph.

docs/designs/ps_design.md
```diff
@@ -1,53 +1,62 @@
...
+*PS(i) = hash(key) mod N*
 
-    def set_embedding_vector(self, name, indices, value):
-        pass
-```
+Each parameter server pod will store some embedding vectors and dense tensor parameter, it only holds a subset of the whole model.
+
+The number of parameter server pods will keep unchanged during a job. We set the priority of parameter server pods higher than worker pods. Thus, worker pods of a lower priority job will be killed first to satisfy another job. If all worker pods of a job are dead , this job is actually stopped. We will restart the job again when the hardware resources are sufficient.
+
+## KVStore
```
QiJune: Remove KVStore data structure.

docs/designs/ps_design.md
```diff
@@ -1,53 +1,56 @@
...
+- How to make model sharding in parameter server pods?
+- What kind of data structure is used to keep parameters in the memory of parameter server pods?
+- How parameters are initialized in parameter server pods?
+- How parameters are updated in parameter server pods?
+- How to make the parameter server architecture fault-tolerant?
 
-The worker initializes a model, and pushes parameters to KVStore. Before each step of training, the worker pulls the latest model from KVStore. After a round of forward/backward computation, the worker pushes gradients to the PS waiting for processing. Then, optimizer of PS will look up the corresponding parameter from KVStore. At last, it applies gradients to parameters, and updates parameter back to KVStore.
+## Model Sharding
 
+We will distribute model parameters into multiple PS pods, which is called model sharding. There is a hash function that maps a parameter to a PS pod id. And this parameter will be stored at this corresponding PS pod.
```
mhaoli: `And this parameter will be stored at this corresponding PS pod.` looks strange to me. I tried to modify these two sentences to:
> We can use a hash function p(x) to decide that parameter x will be stored at PS pod p(x).

docs/designs/ps_design.md
```diff
@@ -1,53 +1,56 @@
...
-    def set_parameter(self, name, value):
-        pass
+The number of parameter server pods will keep unchanged during a job. We set the priority of parameter server pods higher than worker pods. Thus, worker pods of a lower priority job will be killed first to satisfy another job. If all worker pods of a job are dead , this job is actually stopped. We will restart the job again when the hardware resources are sufficient.
 
-    def set_embedding_vector(self, name, indices, value):
-        pass
-```
+## Model Data Structure
+
+We use two hashmaps to store embeding table parameters and dense tensor parameters.
```
mhaoli, line 47: Some places in this doc use `parameter`, and some places use `parameters`. It would be better to keep the singular and plural numbers consistent.

docs/designs/ps_design.md
```diff
@@ -58,322 +61,180 @@ class EmbeddingTable(object):
         self.vectors = {}
```
mhaoli: It seems a little strange that variable `vectors` is a python dictionary. I modified the variable name to `embeddings` in my PR. Maybe better to change a name.

docs/designs/ps_design.md
```diff
@@ -58,322 +61,180 @@ class EmbeddingTable(object):
...
+The name of an embedding table is actually the embedding layer name. The embedding table uses a dictionary `vectors` to store embedding vectors, the key is the item id, the value is the embedding vector.
 
-Since embedding table is lazily initialized in PS, it also has `dim` and `initializer` fields.
+Please note that the embedding tables in the parameter server pods which have the same name, form the big embedding table for a certain embedding layer.
 
-### Tensor Data Structure
+Since embedding vectors are lazily initialized in parameter server, `EmbeddingTable` also has `dim` and `initializer` fields. Inside the `get` interface of `EmbeddingTable `, if the id is not in the `vectors` dictionary, the corresponding value will be initialized.
 
-To support data communication between workers and PS, we introduce a `Tensor` proto message:
+There could be multiple embedding table from different embedding layer. We will create an `EmbeddingTable` instance for each embedding layer. These instances are stored at another hashmap. The key is embedding layer name, the value is the embedding table itself.
```
mhaoli: multiple embedding tables
different embedding layers

QiJune: Done

docs/designs/ps_design.md
QiJune: Done.  Renaming to `embedding_vectors`.

docs/designs/ps_design.md
QiJune: Done

docs/designs/ps_design.md
QiJune: Done. Add `the` or `a` before `parameter`.

PR #1317, opened 2019-10-16
PS design doc (combined version)

skydoorkai:
[Here](https://github.com/sql-machine-learning/elasticdl/blob/haitao_ps_design/docs/designs/ps_design.md) is for a better review.


docs/designs/ps_design.md
```diff
@@ -1,377 +1,94 @@
 # Parameter Server Design
+This document describes the design of a distributed parameter server (PS) to support training scalability and PS fault tolerance in ElasticDL.
```
wangkuiyi: The appearance of "ElasticDL" is abrupt. We could change the title to be "The Parameter Server of ElasticDL", or talk about ElasticDL by inserting a sentence before this one.

wangkuiyi: The concept of fault-tolerance appears abruptly too. Reader would be surprised -- why do we need fault-tolerance?

wangkuiyi: fault tolerance => fault-tolerance

docs/designs/ps_design.md
```diff
@@ -1,377 +1,94 @@
...
 
 
-## Overview
-Currently, there is one parameter server (PS) co-existed with the master. In order to support PS with multiple nodes and PS fault tolerance, we need to separate PS from the master. Besides a KV store for model parameters, PS should also support updating parameters using gradients. Currently, models have two kinds of parameter, variable and embedding vector. PS should support model variable update and embedding table sparse update.
+## Motivation
+Parameter server based distributed training uses data parallelism to speed up training. PS stores model parameters. There are multiple workers that get model parameters from PS, compute gradients using different training data and send computed gradients to PS. PS iteratively updates these model parameters using gradients sent by workers. A PS based distributed training can use an arbitrary number of workers to support the scalability of training data size.
```
wangkuiyi: It is not "parameter server-based distributed training uses data parallelism"; instead, it is "data parallelism-based training systems who use data parallelism".

wangkuiyi: We need to reorder these sentences in this paragraph or rewrite them.

docs/designs/ps_design.md
```diff
@@ -1,377 +1,94 @@
...
 
-The master will create a PS with *N* PS nodes, where *N* is specified by the user. Each model variable and embedding vector has a corresponding PS node. Thus, every PS node has a subset of model variables and embedding tables.
+If a model has a very large size, it may not fit in the memory of a single parameter server. For example, many recommending and ranking models use embedding layers. The parameter of an embedding layer is an embedding table, which consists of multiple embedding vectors. When the number of embedding vectors is too large to store in a single PS, we need a distributed PS. A distributed PS contains multiple PS pods.  The distributed PS can partition model parameters and store different partitions in different PS pods.
```
wangkuiyi: If a model has a very large size, it may not fit in the memory ...

=>

Big models might overrun the memory space of a single parameter server process.

wangkuiyi: I guess the author wanted to express the following logic derivation process in this paragraph:

1. We want to have one or more parameter server instances in each job.
1. A reason is that models could be large and overrun the memory space of each single process.
1. In such cases, we need to shard the model.
1. Even if the model is not too big and fits in the memory of a single parameter server instance,
1. we might still want model sharding,
1. because this distribute the parameter access communication from the workers in the job.
1. and it also distributes the computation like embedding vector combination and parameter optimization over multiple parameter server instances.

docs/designs/ps_design.md
```diff
@@ -1,377 +1,94 @@
...
 
-![pserver](../images/pserver.png)
+After the relaunch of a PS pod, the PS pod needs to recover its partition of model parameters. For a worker, a minibatch of training data in a forward-pass contains some embedding vectors, which is a subset of embedding tables (the parameters of the corresponding embedding layers). The worker pulls all non-embedding parameters and a subset of embedding tables from PS pods in the training. The PS pod can recover non-embedding parameters from workers but not embedding tables.
 
-PS contains two main compoments:
+In order to recover the embedding table partition in the relaunched PS pod, the distributed PS needs to have embedding table replicas. A PS pod can store its embedding table partition replicas in other PS pods. For example, assume there are *N* PS pods from *PS<sub>0</sub>* to *PS<sub>N-1</sub>*, *PS<sub>i</sub>* can stores its replica in *PS<sub>(i + 1) % N</sub>*. The relaunched PS pod  *PS<sub>i</sub>* can recover its embedding table partition from its replica in  *PS<sub>(i + 1) % N</sub>*.
```
wangkuiyi: It might not be a good idea to use HTML tag <sub> in Markdown files. Some renderers do not recognize such tags. It also makes it less comprehensive to readers. I would usually Google "subscript x" and paste the Unicode points in the Markdown files.

skydoorkai: Mac cannot display some Unicode of alphabet subscripts correctly, such as 'm'.
The `<sub>` tag works fine on github and on Mac.

docs/designs/ps_design.md
```diff
@@ -1,377 +1,97 @@
...
 
-Each worker has a local copy of the model variables. After the master relaunches a PS node, the PS node can recover model variables from workers. For embedding vectors, PS must create replicas to support fault tolerance. For each PS node *PS(i)*, it will store *M* replicas in the following *M* PS nodes from *PS(i+1 % N)* to *PS(i+M % N)*. The relaunched PS node can recover embedding vectors from one of its replicas. If there are more than *M* continuously-indexed PS nodes failing, at least one PS node fails with all of its replicas. The ElasticDL job has to recover from a recent checkpoint.
+After the relaunch of a PS pod, the PS pod needs to recover its partition of model parameters. The model may contain one or more embedding layers with embedding tables as their parameters. If so, a minibatch of training data in a worker contains some embedding vectors, which is a subset of embedding tables. The worker pulls all non-embedding parameters and and only a subset of embedding tables from PS pods in the training. Thus, the PS pod can recover non-embedding parameters from workers but not embedding tables.
```
QiJune: > If so, a minibatch of training data in a worker contains some embedding vectors, which is a subset of embedding tables.

training data contains embedding vectors?

docs/designs/ps_design.md
```diff
@@ -1,377 +1,97 @@
...
-
-```python
-def serialize_to_pb(tensor, pb):
-    pass
+Each embedding layer has an embedding table which maps a discrete ID *i* to an embedding vector *vᵢ*. Because many recommending and ranking models have large embedding tables, we partition each embedding table and store every partition in an unique PS pod. For an embedding vector *vᵢ*, we select *PS<sub>i % N</sub>* to store it.
 
-def deserialize_from_pb(pb, tensor):
-    pass
+## Model Parameter Storage
+Each PS node has a dictionary data structure to store its partition of model parameters.
```
QiJune: node --> pod

docs/designs/ps_design.md
QiJune: And I think it's better to introduce model parameter partition/storage in PS first, then we  dive into how to make parameters in PS fault-tolerance.

docs/designs/ps_design.md
```diff
@@ -1,377 +1,97 @@
...
 
-The master will monitor PS nodes status similar to what it does for workers. In case a PS node fails, the master will try to relaunch it. Since PS nodes have higher priority than workers, if there are still some running workers, the relaunch will succeed by using either idle or preempted Kubernetes resources. If the relaunch fails, there are no workers left. The Elastic job has to wait for resources for the PS node and workers.
+In ElasticDL, a master is responsible for creating PS pods and worker pods using Kubernetes APIs. The master launches PS pods with high priority and launches worker pods with low priority. The master also monitor PS and worker pods status. In case a PS pod fails, the master will relaunch it using Kubernetes APIs. Since PS pods have a higher priority than worker pods, if there are still some worker pods running, the relaunch will succeed by using either idle or preempted Kubernetes resources. If no worker pods left, ElasticDL has to wait for Kubernetes resources to continue the training.
```
mhaoli: `monitor` -> `monitors`

docs/designs/ps_design.md
mhaoli: duplicated `and` here.

mhaoli: Agree with @QiJune . It is a little confusing for readers that why workers have a full set of non-embedding parameters, but only have a subset of embedding parameters. You may introduce the reason first. Or maybe introduce parameter partition/storage first. Or you can ignore this detail and only tell readers that PS recover parameters from replica in this section, and illustrate this detail in the *Model Parameter Recovery* section.

docs/designs/ps_design.md
```diff
@@ -1,377 +1,97 @@
...
-}
+We choose to store each non-embedding parameter using the parameter name as its key, and a [`tf.Variable`](https://www.tensorflow.org/api_docs/python/tf/Variable) instance as its value. This is because that we want to update non-embedding parameters directly by [TensorFlow optimizers](https://www.tensorflow.org/api_docs/python/tf/optimizers).
 
-message Model {
-    int64 version = 1;
-    repeated Tensor variables = 2;
-    repeated EmbeddingTableInfo embedding_table_info = 3;
-}
-```
+If a model has one or more embedding layers, a minibatch of training data contains a set of discrete IDs. These discrete IDs correspond to a set of embedding vectors. The Worker needs to pull these embedding vectors from their corresponding PS pods using the embedding layer name and the discrete IDs. To store an embedding vector, We use its corresponding embedding layer name and discrete ID to form a pair, and use this pair as the key and the embedding vector itself as the value.
```
mhaoli: The Worker -> The worker

docs/designs/ps_design.md
```diff
@@ -1,173 +1,194 @@
...
 
-## PS
+In order to recover the embedding table partition in the relaunched PS pod, PS needs to replicate embedding tables. A PS pod can store its embedding table partition replicas in other PS pods. For example, assume there are *N* PS pods from *PS₀* to *PS<sub>N-1</sub>*, *PSᵢ* can stores its replica in *PS<sub>(i + 1) % N</sub>*. The relaunched PS pod  *PSᵢ* can recover its embedding table partition from its replica in  *PS<sub>(i + 1) % N</sub>*.
 
-### Compoments
+In the following sections, we will explain the distributed PS design in detail, including how to [partition the model](#model-parameter-partition), [store model parameters](#model-parameter-storage), [access parameters from workers](#model-parameter-access-from-worker), [initialize parameters](#model-parameter-initialization), [update parameters from gradients](#model-parameter-update) and support PS fault-tolerance with [replicas](#embedding-replica).
 
-![pserver](../images/pserver.png)
+## Model Parameter Partition
+For a distributed PS with *N* PS pods, each PS pod stores a model parameter partition.
```
QiJune: The number of PS pod **N** keeps unchanged during the training job.

docs/designs/ps_design.md
```diff
@@ -1,173 +1,194 @@
...
 
-The KVStore could be like following:
+We choose to store each non-embedding parameter using the parameter name as its key, and a [`tf.Variable`](https://www.tensorflow.org/api_docs/python/tf/Variable) instance as its value. This is because that we want to update non-embedding parameters directly by [TensorFlow optimizers](https://www.tensorflow.org/api_docs/python/tf/optimizers).
 
-```python
-class KVStore(object):
-    def __init__(self):
-        self.variable_db = {}
-        self.embedding_table_db = {}
+If a model has one or more embedding layers, a minibatch of training data contains a set of discrete IDs. These discrete IDs correspond to a set of embedding vectors. The Worker needs to pull these embedding vectors from their corresponding PS pods using the embedding layer name and the discrete IDs. To store an embedding vector, We use its corresponding embedding layer name and discrete ID as the keys and the embedding vector itself as the value.
```
QiJune: a 1-D numpy.ndarray as the value.

mhaoli (approved):

PR #1324, opened 2019-10-17
Split requirements to required deps and extras

### LINKED ISSUE #1286: Only keep necessary dependencies in requirements.txt
Currently, we have one single `requirements.txt` file which contains some unnecessary dependencies used for testing only such as `pytest`, `pytest-cov`, and `mock`. They should not be part of `requirements.txt` and should not be distributed with `elasticdl` Python package.

terrytangyuan:
Fixes #1286. After this, pip install will only install the required dependencies in `requirements.txt`. If developers wants to install all the optional dependencies in `requirements-dev.txt`, they can do so by installing the extra `develop` module via `pip install -e ".[develop]"`.


chunyang-wen (approved):
Awesome.

PR #1328, opened 2019-10-21
Combine unit and integration tests

terrytangyuan:
Most of these files are not covered in unit tests but covered in integration tests instead so we should not take them into account.

Note that somehow the uploaded coverage report is still outdated so I added some additional tests and the issue has been fixed.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -13,16 +13,5 @@ coverage:
 
 ignore:
   - "elasticdl/python/tests"
-  - "elasticdl/python/common/k8s_client.py"
-  - "elasticdl/python/common/k8s_tensorboard_client.py"
-  - "elasticdl/python/master/k8s_worker_manager.py"
-  - "elasticdl/python/master/main.py"
-  - "elasticdl/python/master/tensorboard_service.py"
-  - "elasticdl/python/worker/main.py"
-  - "elasticdl/python/elasticdl/api.py"
-  - "elasticdl/python/elasticdl/client.py"
-  - "elasticdl/python/elasticdl/image_builder.py"
-  - "elasticdl/python/data/recordio_gen"
-  - "elasticdl/python/data/odps_io.py"
 
 comment: false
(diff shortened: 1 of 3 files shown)


terrytangyuan:
Somehow the uploaded coverage report is still outdated. I am adding some additional tests to see if the issue is fixed.

ywskycn:
> Most of these files are not covered in unit tests but covered in integration tests instead so we should not take them into account.

You mean these files will bring down the coverage rate?

terrytangyuan:
@ywskycn Yes, previously the coverage percentage was around 50% which is pretty low and it has been boosted to 80% after this PR. Most of those code rely on Minikube and ODPS and they are actually covered in Travis.

.codecov.yml
```diff
@@ -13,5 +13,16 @@ coverage:
 
 ignore:
   - "elasticdl/python/tests"
+  - "elasticdl/python/common/k8s_client.py"
```
wangkuiyi: May we know why should we exclude these source files from unit test coverage stats?

ywskycn:
> @ywskycn Yes, previously the coverage percentage was around 50% which is pretty low and it has been boosted to 80% after this PR. Most of those code rely on Minikube and ODPS and they are actually covered in Travis.

But that should be fine, right? Otherwise, we always need to think about whether we should exclude newly added files or not.

terrytangyuan:
@ywskycn @wangkuiyi I think we can probably combine non-k8s and k8s-related tests so the relevant code doesn't have to be excluded in coverage calculation. For ODPS related code, we still have to exclude them since Travis builds will skip ODPS related tests for pull requests from forks. What do you think?

terrytangyuan:
Discussed with @ywskycn offline, I combined all tests to one stage on Travis and the k8s related code should be included and covered in test coverage report now. No code will be excluded from test coverage anymore.

ywskycn (approved):

PR #1330, opened 2019-10-21
ElasticDL serving design

brightcoder01:


docs/designs/serving_design.md
```diff
@@ -0,0 +1,101 @@
+# Serving Design
+
+## Background
+
+Model serving is an essential part in an end-to-end machine learning lifecycle. Publishing the trained model as a service in production can make it valuable in the real world. In the literature, we focus on how to save the model trained in ElasticDL for serving.
+
+[SavedModel](https://www.tensorflow.org/guide/saved_model?hl=zh_cn) is the universal serialization format for TensorFlow models. It's language neutral and can be loaded by multiple frameworks (such as TFServing, TFLite, TensorFlow.js and so on). We choose to export the ElasticDL model to the SavedModel format. In this way, we can leverage various mature solutions to serving our model in different scenarios.
```
ywskycn: Let's use https://www.tensorflow.org/guide/saved_model for the `SavedModel` link.

docs/designs/serving_design.md
```diff
@@ -0,0 +1,101 @@
...
+
+|                            | Master Central Storage |  AllReduce  |            Parameter Server              |
+|:--------------------------:|:----------------------:|:-----------:|:----------------------------------------:|
+| Small or Medium Size Model |       SavedModel       |  SavedModel |               SavedModel                 |
+| Large Size Model           |          N/A           |     N/A     | Distributed Parameter Server for Serving |
+
+Small or medium size model\
+A single serving process can load the entire model into memory. No matter which training strategy to choose, we will export the model to the SavedModel format. For master central storage and AllReduce strategy, each worker has a replica of the entire model and then export it directly. For parameter server strategy, the model contains unsaveable elasticdl.layers.Embedding layer and the variables are stored in the PS. We will use the process in [the section below](#Export-the-model-with-elasticdl.layers.Embedding-to-SavedModel) to export the model. And then we can deploy it using the existed serving frameworks like TFServing. **We focus on this case in this article.**
```
ywskycn, line 19: Why we need to differentiate `small` and `medium` here, why not just `small`, as these two sizes are handled the same way.

docs/designs/serving_design.md
```diff
@@ -0,0 +1,101 @@
...
+
+The model size can vary from several kilobytes to several terabytes. Exporting to savedModel can't apply to all the scenarios. We divide the model size into two categories: *Small or medium size* and *large size*. The small or medium size model can be loaded by a process, and the latter can not fit in a single process. Training and serving strategies are different between these two cases. Please check the following table:
+
+|                            | Master Central Storage |  AllReduce  |            Parameter Server              |
+|:--------------------------:|:----------------------:|:-----------:|:----------------------------------------:|
+| Small or Medium Size Model |       SavedModel       |  SavedModel |               SavedModel                 |
+| Large Size Model           |          N/A           |     N/A     | Distributed Parameter Server for Serving |
```
ywskycn, line 14: What is the meaning of `Master Central Storage`?

docs/designs/serving_design.md
```diff
@@ -0,0 +1,101 @@
...
+        if type(layer) == tf.keras.layers.Embedding:
+            embedding_params = EmbeddingService.get_all_embedding_params(layer)
+            layer.embeddings.assign(embedding_params)
+```
+
+## Execute task to save model with fault-tolerance
+
+We designed the master-worker architecture and task dispatch&recover mechanism in ElasticDL to make the job execution fault tolerant. Please check [overall design](./overall_design.md).\
+For model saving work, we use the same mechanism. After completing all the training/evaluation/prediction tasks, master will generate a SaveModel task and insert it into the todo task queue. The first worker pulling this task will execute the model saving process. Please check the diagram below:
+
```
ywskycn, line 123: From the diagram, "If all the tasks are done, generate a SaveModel task", so the last finished worker will do the `SaveModel` task? As other workers already quit..

docs/designs/serving_design.md
```diff
@@ -0,0 +1,101 @@
+# Serving Design
+
+## Background
+
+Model serving is an essential part in an end-to-end machine learning lifecycle. Publishing the trained model as a service in production can make it valuable in the real world. In the literature, we focus on how to save the model trained in ElasticDL for serving.
+
+[SavedModel](https://www.tensorflow.org/guide/saved_model?hl=zh_cn) is the universal serialization format for TensorFlow models. It's language neutral and can be loaded by multiple frameworks (such as TFServing, TFLite, TensorFlow.js and so on). We choose to export the ElasticDL model to the SavedModel format. In this way, we can leverage various mature solutions to serving our model in different scenarios.
+
+The model size can vary from several kilobytes to several terabytes. Exporting to savedModel can't apply to all the scenarios. We divide the model size into two categories: *Small or medium size* and *large size*. The small or medium size model can be loaded by a process, and the latter can not fit in a single process. Training and serving strategies are different between these two cases. Please check the following table:
```
terrytangyuan: savedModel -> SavedModel

docs/designs/serving_design.md
```diff
@@ -0,0 +1,101 @@
...
+
+|                            | Master Central Storage |  AllReduce  |            Parameter Server              |
+|:--------------------------:|:----------------------:|:-----------:|:----------------------------------------:|
+| Small or Medium Size Model |       SavedModel       |  SavedModel |               SavedModel                 |
+| Large Size Model           |          N/A           |     N/A     | Distributed Parameter Server for Serving |
+
+Small or medium size model\
```
terrytangyuan: No need for `"\"`

docs/designs/serving_design.md
terrytangyuan, line 19: Where is "master central storage and AllReduce strategy" defined?

docs/designs/serving_design.md
```diff
@@ -0,0 +1,101 @@
...
+
+Large size model\
+A single serving process will run out of memory while loading the model. We partition the model variables into multiple shards, store them in distributed parameter server for serving. The inference engine will execute the serving graph, query the variable values from the distributed parameter server as needed and finish the calculation. It's necessary to upgrade the serving framework to support this. **We will discuss this case in a separate design in the next step.**
+
+In this article, we want to achieve these three goals in ElasticDL:
+
+1. Guarantee consistency between training and serving for data input.
```
terrytangyuan: What kind of consistency? I think you meant the processing logic for the input data, right?

docs/designs/serving_design.md
```diff
@@ -0,0 +1,101 @@
...
+        return keras.models.clone_model(model, clone_function=_clone_function)
+    else:
+    # replace embedding attribute for subclass model
+        for attr_name, attr_value in model.__dict__.items():
+            if type(attr_value) == keras.layers.Embedding:
+                setattr(model, attr_name, edl_Embedding(attr_value.output_dim))
+        return model
+```
+
+However, tf.saved_model.save can not export the replaced model to SavedModel. Because ElasticDL.Embedding use tf.py_function to invoke Rpc to interact with the parameter server. It is not mapped to any native TensorFlow op. As a result we choose to save the origin model with native keras embedding layer, replace the embedding parameters with the trained parameters of elasticdl.layers.embedding.
```
terrytangyuan: can not -> cannot

terrytangyuan: use -> uses

docs/designs/serving_design.md
```diff
@@ -0,0 +1,101 @@
...
+    for layer in model.layers:
+        if type(layer) == tf.keras.layers.Embedding:
+            embedding_params = EmbeddingService.get_all_embedding_params(layer)
+            layer.embeddings.assign(embedding_params)
+```
+
+## Execute task to save model with fault-tolerance
+
+We designed the master-worker architecture and task dispatch&recover mechanism in ElasticDL to make the job execution fault tolerant. Please check [overall design](./overall_design.md).\
+For model saving work, we use the same mechanism. After completing all the training/evaluation/prediction tasks, master will generate a SaveModel task and insert it into the todo task queue. The first worker pulling this task will execute the model saving process. Please check the diagram below:
```
terrytangyuan: May need to add more details on what will happen when one of the tasks fail, how do we shard the tasks, etc.

docs/designs/serving_design.md
brightcoder01: Good idea, updated

docs/designs/serving_design.md
brightcoder01: Good catch, updated.

docs/designs/serving_design.md
brightcoder01: It means that the entire model (including dense variables and embedding table) are stored in the master.
It's the first version of implementation in ElasticDL.

docs/designs/serving_design.md
brightcoder01: Yes, you are right. Updated the description to be clearer.

docs/designs/serving_design.md
brightcoder01: Good catch, updated.

docs/designs/serving_design.md
brightcoder01: I think we don't need shard the SavedModel task. Because we only generate one task of this type. The first worker pulling the task will complete this work. 
If the worker is preempted while saving the model, the task will be resumed and assigned to other workers just like training/eval tasks.

docs/designs/serving_design.md
brightcoder01: If all the tasks are done, the master will generate a SavedModel task.
The first worker pulling this task will execute the task. Other workers are waiting.
If the worker executing the model saving process is preempted. The SaveModel task is assigned to other workers.
When the model save task is finished, all the workers will exit.
Added more description.

docs/designs/serving_design.md
brightcoder01: It seems that *cannot* is more popular. Updated. :)

docs/designs/serving_design.md
```diff
@@ -0,0 +1,101 @@
...
+
+|                            | Master Central Storage |  AllReduce  |            Parameter Server              |
+|:--------------------------:|:----------------------:|:-----------:|:----------------------------------------:|
+| Small or Medium Size Model |       SavedModel       |  SavedModel |               SavedModel                 |
+| Large Size Model           |          N/A           |     N/A     | Distributed Parameter Server for Serving |
+
+Small or medium size model\
+A single serving process can load the entire model into memory. No matter which training strategy to choose, we will export the model to the SavedModel format. For master central storage and AllReduce strategy, each worker has a replica of the entire model and then export it directly. For parameter server strategy, the model contains unsaveable elasticdl.layers.Embedding layer and the variables are stored in the PS. We will use the process in [the section below](#Export-the-model-with-elasticdl.layers.Embedding-to-SavedModel) to export the model. And then we can deploy it using the existed serving frameworks like TFServing. **We focus on this case in this article.**
```
mhaoli, line 19: Unsaveable ElasticDL embedding layer and PS strategy are two constraints that are independent. For small or medium models with embedding table, we may also use AllReduce strategy to update it.

docs/designs/serving_design.md
brightcoder01: Add the description of different training strategies mentioned.

docs/designs/serving_design.md
brightcoder01: Add the description of these training strategies :)

brightcoder01: Yes, in the doc, these two model size are handled the same way. But it seems that there are some gap between small and large size, readers may think it doesn't cover all the cases.

docs/designs/serving_design.md
```diff
@@ -0,0 +1,103 @@
...
+
+|                            | Master Central Storage |  AllReduce  |            Parameter Server              |
+|:--------------------------:|:----------------------:|:-----------:|:----------------------------------------:|
+| Small or Medium Size Model |       SavedModel       |  SavedModel |               SavedModel                 |
+| Large Size Model           |          N/A           |     N/A     | Distributed Parameter Server for Serving |
+
+We consider three main training strategies of ElasticDL. Master central storage strategy (Implemented): The entire model is loaded in the master node. Workers pull the model from the master while processing each minibatch of data. AllReduce strategy (Planning): It mainly support sync training. Each worker loads a replica of the entire model. AllReduce aggregates gradients across all the workers and publish them to each worker. Parameter server strategy (In Progress): We store the model variables on the parameter servers. All the workers pull the variables as needed from the parameter server and execute computation.
```
skydoorkai: Better to add (1) (2) (3) before the 3 strategies to make it look clearer.

skydoorkai: support -> supports

docs/designs/serving_design.md
```diff
@@ -0,0 +1,103 @@
...
+
+1. Guarantee consistency between training and serving for data input.
+2. Export the model with elasticdl.layers.Embedding to SavedModel.
+3. Execute task to save model with fault-tolerance.
+
+## Guarantee consistency between training and serving for data input
+
+ElasticDL is a distributed deep learning framework based on TensorFlow 2.0 eager execution. In ElasticDL, we use tf.data.Dataset to create input pipeline for training. The recommended way to preprocess data from tf.data.Dataset is to use [feature columns](https://www.tensorflow.org/tutorials/structured_data/feature_columns) in TensorFlow. What's more, tf.saved_model.save will save the defined feature columns with the model. So tf-serving will use the same preprocessing logic with training to make inference.
+
+Define a keras model with feature columns:
```
skydoorkai: keras - >Keras

docs/designs/serving_design.md
```diff
@@ -0,0 +1,103 @@
...
+
+## Export the model with elasticdl.layers.Embedding to SavedModel
+
+Using native Keras layers to define a model is more user-friendly than using custom layers in ElasticDL. However, it is inefficient to train a model with tf.keras.layers.Embedding. When the model executes the forward-pass computation for each mini-batch, it must get all embedding parameters from the parameter server (PS) even if the mini-batch only contains several embedding ids. So, the elastic.layers.Embedding is designed to improve the training efficiency in ElasticDL. Considering user-friendliness and training efficiency, we need to define a model with tf.keras.layers.Embedding and train the model with elasticdl.layers.Embedding. For the Sequential model and the Functional models, we can use tf.keras.models.clone_model to replace the tf.keras.layers.Embedding with elasticdl.layers.Embedding before training starts. For subclass model, we can replace the attribute of tf.keras.layers.Embedding type with elastic.layers.Embedding.
```
skydoorkai: Keras does not have embedding layers with sparse inputs and a combiner. 
How to deal with such a layer?

skydoorkai: So users have to define embedding layers as class attributes.

docs/designs/serving_design.md
brightcoder01: I think we only use ElasticDL.layers.Embedding in the PS strategy. 
For AllReduce, we will use the keras native embedding layer.

docs/designs/serving_design.md
brightcoder01: Good idea, updated.

docs/designs/serving_design.md
brightcoder01: Yes, I think it will be common practice that user defines the layers as attributes and calls it in the model.call method for subclass model.

docs/designs/serving_design.md
```diff
@@ -0,0 +1,103 @@
...
+        return keras.models.clone_model(model, clone_function=_clone_function)
+    else:
+    # replace embedding attribute for subclass model
+        for attr_name, attr_value in model.__dict__.items():
+            if type(attr_value) == keras.layers.Embedding:
+                setattr(model, attr_name, edl_Embedding(attr_value.output_dim))
+        return model
+```
+
+However, tf.saved_model.save cannot export the replaced model to SavedModel. Because ElasticDL.Embedding uses tf.py_function to invoke Rpc to interact with the parameter server. It is not mapped to any native TensorFlow op. As a result we choose to save the origin model with native keras embedding layer, replace the embedding parameters with the trained parameters of elasticdl.layers.embedding.
```
QiJune: Rpc --> RPC

docs/designs/serving_design.md
```diff
@@ -0,0 +1,103 @@
...
+
+```python
+def restore_keras_embedding_for_subclass(model):
+    for attr_name, attr_value in model.__dict__.items():
+        if type(attr_value) == elasticdl.layers.Embedding:
+            setattr(model, attr_name, keras.layers.Embedding(attr_value.output_dim))
+    return model
+
+def restore_keras_embedding_from_edl_embedding(model):
```
QiJune: Maybe `restore_keras_model_param`

docs/designs/serving_design.md
```diff
@@ -0,0 +1,103 @@
...
+        for attr_name, attr_value in model.__dict__.items():
+            if type(attr_value) == keras.layers.Embedding:
+                setattr(model, attr_name, edl_Embedding(attr_value.output_dim))
+        return model
+```
+
+However, tf.saved_model.save cannot export the replaced model to SavedModel. Because ElasticDL.Embedding uses tf.py_function to invoke Rpc to interact with the parameter server. It is not mapped to any native TensorFlow op. As a result we choose to save the origin model with native keras embedding layer, replace the embedding parameters with the trained parameters of elasticdl.layers.embedding.
+
+```python
+def restore_keras_embedding_for_subclass(model):
```
QiJune: Maybe `restore_keras_model_def`

docs/designs/serving_design.md
brightcoder01: Good catch, updated.

docs/designs/serving_design.md
brightcoder01: Good catch, updated.

docs/designs/serving_design.md
brightcoder01: That's a good question. 
At the current stage, we use feature_column.embedding_column for sparse input.
But it can't leverage the benefit from the embedding sharding mechanism of  parameter server. We need figure out a way in a separate topic.

docs/designs/serving_design.md
brightcoder01: Good catch, updated.

docs/designs/serving_design.md
brightcoder01: More details are added :)

docs/designs/serving_design.md
```diff
@@ -0,0 +1,103 @@
...
+
+|                            | Master Central Storage |  AllReduce  |            Parameter Server              |
+|:--------------------------:|:----------------------:|:-----------:|:----------------------------------------:|
+| Small or Medium Size Model |       SavedModel       |  SavedModel |               SavedModel                 |
+| Large Size Model           |          N/A           |     N/A     | Distributed Parameter Server for Serving |
+
+We consider three main training strategies of ElasticDL. 1) Master central storage strategy (Implemented): The master node loads the entire model. Workers pull the model from the master while processing each minibatch of data. 2) AllReduce strategy (Planning): It mainly supports synchronized training. Each worker loads a replica of the entire model. AllReduce aggregates gradients across all the workers and publish them to each worker. 3) Parameter server strategy (In Progress): We store the model variables on the parameter servers. All the workers pull the variables as needed from the parameter server and execute computation.
```
ywskycn: `Implemented` -> `implemented`, `Planning` -> `planning`, `In Progress` -> 'in progress`.
`publish them to each worker` -> `publishes them to each worker`.
`All the workers` ->  `All workers`. We always overuse `the` :)

docs/designs/serving_design.md
```diff
@@ -0,0 +1,103 @@
...
+
+def custom_model(feature_columns):
+    input_layers = get_input_layer()
+    dense_feature = tf.keras.layers.DenseFeatures(feature_columns=feat_cols)(input_layers)
+    dense = tf.keras.layers.Dense(10, activation='relu')(dense_feature)
+    dense = tf.keras.layers.Dense(1, activation='sigmoid')(dense)
+    return tf.keras.models.Model(inputs=input_layers, outputs=dense)
+```
+
+Although all feature columns in TensorFlow can be used in ElasticDL, tf.feature_column.embedding_column is not recommended in ElasticDL. Because the embedding_column has a variable containing a large embedding table. In eager execution the model must get all the embedding parameters to train. It will bring a large inter-process communication overhead.
```
ywskycn: It looks like we miss some sentences at the end. We don't recommend users to use tf.feature_column.embedding_column, based on some reasons. So users should use ..? We can add one more sentence to connect with the next section. Or we just remove this paragraph, as I saw you put a detailed explanation at the beginning of the next section.

docs/designs/serving_design.md
mhaoli: Yes, you are right.

docs/designs/serving_design.md
workingloong: When a categorical column has many possible values, tf.feature_column.embedding_column will be used to wrap categorical columns for tf.keras.layers.Densefeature.  But, the embedding layer cannot do it for categorical columns. Now, we don't find a solution to feed the keras model with the output of a categorical column when a categorical column has many possible values. Here, I just want to remind users of this issue.

docs/designs/serving_design.md
workingloong: Done

docs/designs/serving_design.md
workingloong: I approve of the name.

skydoorkai (approved):

Issue #1331, opened 2019-10-22
Failed to reshape embedding outputs when ids shape is (None, None)


workingloong:
The tf.reshape in the Embedding `call` will failed when the  shape of input layer is (None, None) for elasticdl.layers.Embedding and the error is 
TypeError: Failed to convert object of type <class 'tuple'> to Tensor. Contents: (-1, None, 4). Consider casting elements to a supported type

The error can reproduced in tf.keras.models.clone_model:
```
import tensorflow as tf
from elasticdl.python.elasticdl.layers.embedding import Embedding as edl_Embedding
def clone_function(layer):
    if type(layer) == tf.keras.layers.Embedding:
        edl_layer = edl_Embedding(layer.output_dim)
        return edl_layer
    return layer

model = tf.keras.Sequential()
model.add(tf.keras.layers.Embedding(10,4)),
model.add(tf.keras.layers.Dense(4, activation='relu'))
model.add(tf.keras.layers.Dense(1, activation='sigmoid'))
new_model = tf.keras.models.clone_model(model, clone_function=clone_function)
```

workingloong:
We have resolved the issue in  #1332, so I close it.

PR #1332, opened 2019-10-22
set input_dim=-1 when  shape[1] of ids is None

workingloong:
When ids.get_shape()[1] is None, we should use -1 for tf.reshape.  #1331

### DIFF
--- a/elasticdl/python/elasticdl/layers/embedding.py
+++ b/elasticdl/python/elasticdl/layers/embedding.py
@@ -207,7 +207,8 @@ class Embedding(tf.keras.layers.Layer):
         if ids.get_shape().rank == 2:
             input_length = ids.get_shape()[1]
             if input_length is None:
-                return tf.reshape(outputs, (-1, self.output_dim))
+                outputs.set_shape(shape=(None, None, self.output_dim))
+                return outputs
             output_shape = (-1, input_length, self.output_dim)
         else:
             output_shape = ids.get_shape().concatenate(self.output_dim)
--- a/elasticdl/python/elasticdl/layers/embedding.py
+++ b/elasticdl/python/elasticdl/layers/embedding.py
@@ -205,8 +205,10 @@ class Embedding(tf.keras.layers.Layer):
         outputs = tf.gather(batch_embedding, idx)
         # tf.reshape does not support shape with None. Replace None with -1.
         if ids.get_shape().rank == 2:
-            input_dim = 1 if ids.get_shape()[1] is None else ids.get_shape()[1]
-            output_shape = (-1, input_dim, self.output_dim)
+            input_length = (
+                1 if ids.get_shape()[1] is None else ids.get_shape()[1]
+            )
+            output_shape = (-1, input_length, self.output_dim)
         else:
             output_shape = ids.get_shape().concatenate(self.output_dim)
         outputs = tf.reshape(outputs, output_shape)
--- a/elasticdl/python/elasticdl/layers/embedding.py
+++ b/elasticdl/python/elasticdl/layers/embedding.py
@@ -205,9 +205,9 @@ class Embedding(tf.keras.layers.Layer):
         outputs = tf.gather(batch_embedding, idx)
         # tf.reshape does not support shape with None. Replace None with -1.
         if ids.get_shape().rank == 2:
-            input_length = (
-                1 if ids.get_shape()[1] is None else ids.get_shape()[1]
-            )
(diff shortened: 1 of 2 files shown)


elasticdl/python/elasticdl/layers/embedding.py
```diff
@@ -205,7 +205,10 @@ def call(self, input):
         outputs = tf.gather(batch_embedding, idx)
         # tf.reshape does not support shape with None. Replace None with -1.
         if ids.get_shape().rank == 2:
-            output_shape = (-1, ids.get_shape()[1], self.output_dim)
+            input_dim = (
+                -1 if ids.get_shape()[1] is None else ids.get_shape()[1]
+            )
+            output_shape = (-1, input_dim, self.output_dim)
```
skydoorkai: TensorFlow does not support shape containing two `-1` for reshape.

workingloong: yeah, I have fixed the default value to `1`.

terrytangyuan:
Could you add a test to cover this case?

elasticdl/python/elasticdl/layers/embedding.py
brightcoder01: I think input_dim in Embedding Layer means the size of the vocabulary. It's used to make a constraint that the id < input_dim, other wise it's out of range.

ids.get_shape()[1] means the input_length

brightcoder01: And I'm wondering in which scenario that the input_length(aka ids.get_shape()[1]) will be none

workingloong: I think I have mixed the input_dim and input_length. I will rename the input_dim to input_length like:
```python
input_length = (
 1 if ids.get_shape()[1] is None else ids.get_shape()[1]
)
output_shape = (-1, input_length, self.output_dim)
```

workingloong: When we use tf.keras.model.clone_model to clone a new sequential model with elasticdl.Embedding, the`clone_model` will call `call` of elasticdl.Embedding with inputs And the shape of inputs is (None, None).

brightcoder01: > When we use tf.keras.model.clone_model to clone a new sequential model with elasticdl.Embedding, the`clone_model` will call `call` of elasticdl.Embedding with inputs And the shape of inputs is (None, None).

When we are building the Squential model with native embedding layer, it invoke  keras.layers.embedding.call with input(None, None) at the first time. The output is a tensor with shape (None, None, output_dim) without any value inside. Could we keep the same behavior and don't do the reshape?

skydoorkai (approved):

elasticdl/python/elasticdl/layers/embedding.py
workingloong: Besides the `clone_model`, input(None, None) is supported in Keras to build model like the demo:
```
inputs = Input(shape=(None,))
embedding = Embedding(4,2)(inputs)
model = tf.keras.Model(inputs=[inputs], outputs=[embedding])
```

Do you mean that we return the outputs without reshaping when ids.get_shape()[1] is None?

brightcoder01: I think input (None, None) is not only for clone model. It's also invoked in the Squential Api. I think keras use this way to build the layer and initialize the variables. Not for real calculation.

workingloong: I agree with it. So we also should support building the elasticdl.Embedding layer with input(None, None).

elasticdl/python/tests/layer_test.py
```diff
@@ -373,6 +373,15 @@ def test_embedding_layer_gradient_with_sparse_inputs(self):
...
                     )
 
+    def test_embedding_layer_with_none_shape_input(self):
+        output_dim = 8
+        embedding_size = 16
+        layer = create_embedding_layer(embedding_size, output_dim)
+        input = tf.keras.layers.Input(shape=(None,))
+        embedding_output = layer(input)
+        embedding_output_shape = embedding_output.get_shape().as_list()
+        assert embedding_output_shape == [None, 1, 8]
```
terrytangyuan: You can also use self.assertEquals for consistency with other tests

workingloong: OK, do it.

PR #1347, opened 2019-10-23
Add EmbeddingTable Data Structure

### LINKED ISSUE #1305: embedding table Python data structure

QiJune:
Fix #1305


elasticdl/python/tests/embedding_table_test.py
```diff
@@ -0,0 +1,69 @@
...
+import numpy as np
+
+from elasticdl.proto.elasticdl_pb2 import EmbeddingTableInfo
+from elasticdl.python.pserver.embedding_table import (
+    EmbeddingTable,
+    create_embedding_table,
+)
+
+
+def prod(iterable):
```
skydoorkai: No need to define a method, just use `np.prod(iterable)`.

elasticdl/python/pserver/embedding_table.py
```diff
@@ -0,0 +1,39 @@
+import numpy as np
+import tensorflow as tf
+
+
+class EmbeddingTable(object):
```
skydoorkai: May add some comments to explain the class.

elasticdl/python/pserver/embedding_table.py
```diff
@@ -0,0 +1,39 @@
...
+        self.dim = dim
+        self.initializer = initializer
+        self.embedding_vectors = {}
+
+    def get(self, indices):
+        if len(indices) == 0:
+            return None
+        values = []
+        for i in indices:
+            if i not in self.embedding_vectors:
```
LiMinghao1994: Maybe better change the code snippet 1 to code snippet 2 to query dictionary only once if `i in embedding_vectors`. Because querying dictionary is not fast when there are many embedding vectors.

code snippet 1:
```python
if i not in self.embedding_vectors:
    ...
else:
   value = self.embedding_vectors[i]
``` 

code snippet 2:
```python
value = self.embedding_vectors.get(i, None)
if value is None:
    ....
```

elasticdl/python/pserver/embedding_table.py
```diff
@@ -0,0 +1,39 @@
...
+                init = tf.keras.initializers.get(self.initializer)
+                value = init(shape=self.dim).numpy()
+                self.embedding_vectors[i] = value
```
mhaoli: If we are using synchronous SGD, we may need to add a lock when we write back to `embedding_vectors`. 

It is a little complex. Maybe leave a ToDo here and add this lock when we are implementing parameters updating process?

elasticdl/python/pserver/embedding_table.py
```diff
@@ -0,0 +1,39 @@
...
+            else:
+                value = self.embedding_vectors[i]
```
mhaoli: Maybe better change the code snippet 1 to code snippet 2 to query dictionary only once if `i in embedding_vectors`. Because querying dictionary is not fast when there are many embedding vectors.

code snippet 1:
```python
if i not in self.embedding_vectors:
    ...
else:
   value = self.embedding_vectors[i]
``` 

code snippet 2:
```python
value = self.embedding_vectors.get(i, None)
if value is None:
    ....
```

QiJune: Great idea, Done

elasticdl/python/pserver/embedding_table.py
QiJune: Done, add a TODO here.

elasticdl/python/tests/embedding_table_test.py
QiJune: Done

elasticdl/python/pserver/embedding_table.py
QiJune: Done

mhaoli (approved):
LGTM. @skydoorkai Please double check.

elasticdl/python/pserver/embedding_table.py
skydoorkai: We may add `Parameters` class in the same file. If so, we should rename this file, such as `datatypes.py` or `ps_datatype.py`.

Also, I used the directory name `ps` in another PR. We can use `pserver` in the proto service definition, and use a same directory.

QiJune: It's better to keep each module short and clear. The `Parameters` class will be defined in another file `parameters.py`

And maybe `ps` directory is too short? `pserver` is a little more clear, I could move the main.py to pserver directory, to make all related file in the same directory.

skydoorkai: Using one file or two files for datatypes, both are fine in this case. 

For the directory name, as long as the meaning is clear, short name is fine.  Also consistent with design doc.

QiJune: Done. make it consistent with the design doc

PR #1349, opened 2019-10-23
create `ps.main` for PS main process and parse arguments

### LINKED ISSUE #1339: [PS-1] create `ps/main.py` for PS main process and parse required arguments
Parse arguments required by PS, including to add a new argument `num_ps_pods`, with default value as 0.

num_ps_pods = 0 for ElasticlDL to use legacy PS (PS with master + Redis). We will change the default after we have implemented the new PS.

skydoorkai:
fix #1339 

Also, add a number argument `num_ps_pods` for the number of PS pods.

### DIFF
--- a/elasticdl/python/common/args.py
+++ b/elasticdl/python/common/args.py
@@ -520,6 +520,7 @@ def parse_master_args(master_args=None):
 
     return args
 
+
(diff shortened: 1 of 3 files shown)


QiJune (approved):
Excellent

Issue #1222, opened 2019-09-20
Summarize the current data loading pipeline


wangkuiyi:
An example [command-line](https://github.com/sql-machine-learning/elasticdl/blob/develop/doc/tutorials/elasticdl_local.md#summit-a-training-job) that submits a training job.

```bash
elasticdl train \
  --training_data_dir=/data/mnist/train \
  --evaluation_data_dir=/data/mnist/test \
```

TODO: renmae `--evaluation_data_dir` into `--validation_data_dir`, so to follow the terminology [here](https://en.wikipedia.org/wiki/Training,_validation,_and_test_sets).

Parsing the [training command-line](https://github.com/sql-machine-learning/elasticdl/blob/0b07d1e783f349174f18d89376202040e815b939/elasticdl/python/elasticdl/client.py#L12)

Call function [`elasticdl.train`](https://github.com/sql-machine-learning/elasticdl/blob/0b07d1e783f349174f18d89376202040e815b939/elasticdl/python/elasticdl/api.py#L14).

terrytangyuan:
Keras uses `validation_data` in its `fit()` function as well so I agree the terminology should be reconsidered. 

However, one concern is that based on [this link you gave above](https://en.wikipedia.org/wiki/Training,_validation,_and_test_sets), "validation data" is better for tasks with mixed training + evaluation. If it's evaluation-only job, e.g. via `elasticdl evaluate`, it seems like "test data" is more suitable (a dataset used to provide an unbiased evaluation of a final model fit on the training dataset). Would it be better to keep the current "evaluation_data_dir" so it's suitable for both cases? In my opinion, `evaluation_data_dir` is pretty intuitive since it's the data directory used for **evaluation** tasks (no matter whether this evaluation happens with training or not).

brightcoder01:
The summary has been already added in this PR.

PR #1354, opened 2019-10-24
Modify evaluation_data to validation_data

workingloong:
Rename "evaluation_data" with "validation_data". For the task to calculate the metrics of the trained model on  "validation_data", we should call the task "evaluation task". Because it is a process to evaluate the model performance on  "validation_data".  #1355


terrytangyuan (changes requested):
Should we hold off on changing this until a major release? Also see discussion in https://github.com/sql-machine-learning/elasticdl/issues/1222

ywskycn (approved):
@terrytangyuan should be fine here, as the PR will be merged into `develop` branch. We just need to do a big release for next version.

terrytangyuan (approved):

PR #1358, opened 2019-10-24
Refactor the main function of master

### LINKED ISSUE #1220: The master's main function is too long
We need to split it into multiple function calls, so the function names give a brief overview of what the `main` function does.

Currently, it is hundreds of lines of code:

https://github.com/sql-machine-learning/elasticdl/blob/0b07d1e783f349174f18d89376202040e815b939/elasticdl/python/master/main.py#L61-L312

brightcoder01:
Fix #1220 
In the current implementation, the main function is very long. The component initialization and main loop are put together.

Extract a master component, it mainly contains the following Api：
1. __init__: create the components(such as worker manager and so on) as master's submodule.
2. prepare: start the components one by one, make sure the whole cluster is in the state of ready-to-run.
3. run: run the main loop until all the tasks are completed just like Windows Message Pump.
4. _stop: stop the components such as the RPC server and so on, make sure that whole cluster is shut down.


elasticdl/python/master/main.py
```diff
@@ -1,315 +1,12 @@
...
-                time.sleep(10)
-            else:
-                logger.warning(
-                    "Unable to keep TensorBoard running. "
-                    "It has already terminated"
-                )
-                break
-    logger.info("Master stopped")
+    master = Master(args)
+    master.start()
```
skydoorkai: Why first `start` then `run`? Probably just `run` and put `start` at the beginning of `run`.

brightcoder01: In my opinion, start() means start and setup the whole cluster (including the master RPC service and other sub modules), make sure that the cluster is in the right state and ready to run. 
run() means that we are entering a event loop just like Message Pump in Windows. :)

skydoorkai: Then `prepare` is better than `start` for this method name?

brightcoder01: Good suggestion, *prepare* is more appropriate.

elasticdl/python/master/k8s_worker_manager.py
```diff
@@ -84,6 +84,7 @@ def update_status(self, status):
         )
 
     def start_workers(self):
+        logger.info("Launching %d workers", self._num_workers)
```
terrytangyuan: This is unnecessary since we are already printing out all the parsed arguments

elasticdl/python/master/master.py
```diff
@@ -0,0 +1,401 @@
...
+        ) = self._create_embedding_service(args)
+
+        # Initialize master service
+        self.master_servicer, self.server = self._create_master_service(args)
+
+        # Initialize worker manager
+        self.worker_manager = self._create_worker_manager(args)
+
+    def prepare(self):
+        '''
```
terrytangyuan: Use `"""` - same in other places

elasticdl/python/master/master.py
```diff
@@ -0,0 +1,401 @@
...
+                    if self.checkpoint_output_path:
+                        self.master_servicer.save_latest_checkpoint(
+                            self.checkpoint_output_path
+                        )
+                    break
+                time.sleep(30)
+        except KeyboardInterrupt:
+            self.logger.warning("Server stopping")
+
+        self._stop()
```
terrytangyuan: This isn't guaranteed to be called. Perhaps move to a `finally` block.

elasticdl/python/master/master.py
```diff
@@ -0,0 +1,401 @@
...
+
+    def _stop(self):
+        '''
+        Stop all the components.
+        Make sure that the entire cluster is shut down.
```
terrytangyuan: This is not shutting down the cluster. Maybe change to shut down all the created services and components

elasticdl/python/master/master.py
```diff
@@ -0,0 +1,401 @@
...
+            self.logger.info(
+                "Create TensorBoard service with log directory %s",
+                tensorboard_log_dir,
+            )
+            # Start TensorBoard CLI
+            tb_service = TensorboardService(tensorboard_log_dir, master_ip)
+
+        return tb_service
+
+    def _create_checkpoint_service(self, args):
```
terrytangyuan, line 249: This and other functions below all use `args`. This will make the requirements of args for each function unclear. Perhaps make then more explicit?

elasticdl/python/master/master.py
```diff
@@ -0,0 +1,401 @@
...
+                % str(endpoint)
+            )
+            embedding_dims = dict(
+                [(layer.name, layer.output_dim) for layer in layers]
+            )
+
+        return endpoint, embedding_dims
+
+    def _create_master_service(self, args):
+        # The master service
```
terrytangyuan: Unnecessary comment

elasticdl/python/master/master.py
```diff
@@ -0,0 +1,401 @@
...
+        self.logger.info("Create master service")
```
terrytangyuan: Create -> Creating

elasticdl/python/master/master.py
brightcoder01: Good catch, updated.

elasticdl/python/master/master.py
brightcoder01: Good catch, updated.

elasticdl/python/master/master.py
brightcoder01: Good catch, updated.

elasticdl/python/master/master.py
brightcoder01: Good suggestion, updated.

elasticdl/python/master/master.py
brightcoder01: Good suggestion, updated the similar cases.

elasticdl/python/master/k8s_worker_manager.py
brightcoder01: Good suggestion, removed.

elasticdl/python/master/master.py
brightcoder01: These sub methods almost use the same parameters with the internal Api call. And for _create_worker_manager, it contains almost all the K-V in args. So at the present, I choose pass args as the private method params.

The sub components will be re-designed in the following development. Will refine the parameters at that time.

skydoorkai (approved):

PR #1359, opened 2019-10-24
Add Parameters Data Structure

### LINKED ISSUE #1342: [PS-1] Create `Parameters` and `EmbeddingTable` class definition
According to [PS design](https://github.com/sql-machine-learning/elasticdl/blob/develop/docs/designs/ps_design.md#data-structure).

ps.main also create a `Parameters` instance to store model parameters

Two steps:

- [x] add `EmbeddingTable` #1305 
- [x] add `Parameters` #1359

QiJune:
Fix #1342

### DIFF
--- a/elasticdl/python/ps/parameters.py
+++ b/elasticdl/python/ps/parameters.py
@@ -10,12 +10,6 @@ class Parameters(object):
         self.non_embedding_params = {}
         self.embedding_params = {}
 
-    def get_non_embedding_params(self):
-        return self.non_embedding_params
-
-    def set_non_embedding_params(self, variables):
-        self.non_embedding_params = variables
-
     def get_embedding_param(self, name, indices):
         if name not in self.embedding_params:
             raise ValueError(
--- a/elasticdl/python/ps/parameters.py
+++ b/elasticdl/python/ps/parameters.py
@@ -1,0 +1,44 @@
+from elasticdl.python.ps.embedding_table import create_embedding_table
+
+
+class Parameters(object):
+    def __init__(self):
+        self.init_status = False
+        self.non_embedding_params = {}
+        self.embedding_params = {}
+
+    def get_embedding_param(self, name, indices):
+        if name not in self.embedding_params:
+            raise ValueError(
+                "Please initialize embedding param %s first!", name
+            )
+        return self.embedding_params[name].get(indices)
+
+    def set_embedding_param(self, name, indices, values):
+        if name not in self.embedding_params:
+            raise ValueError(
+                "Please initialize embedding param %s first!", name
+            )
+        self.embedding_params[name].set(indices, values)
+
+    def set_non_embedding_params(self, variables):
+        self.non_embedding_params = variables
+
+    def get_non_embedding_params(self):
+        return self.non_embedding_params
+
+    def init_from_model_pb(self, model_pb):
+        # TODO(qijun) waiting for Tensor/Model proto message definition
+        pass
+
+    def _init_non_embedding_param(self, variables_pb):
+        # TODO(qijun) waiting for Tensor/Model proto message definition
+        pass
+
+    def _init_embedding_param(self, embedding_table_info):
+        table = create_embedding_table(embedding_table_info)
+        self.embedding_params[table.name] = table
+
+    def clear(self):
+        self.non_embedding_params.clear()
+        self.embedding_params.clear()
--- a/elasticdl/python/ps/parameters.py
+++ b/elasticdl/python/ps/parameters.py
@@ -1,3 +1,6 @@
+import tensorflow as tf
+
+from elasticdl.python.common.ndarray import tensor_to_ndarray
 from elasticdl.python.ps.embedding_table import create_embedding_table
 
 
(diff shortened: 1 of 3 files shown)


elasticdl/python/ps/parameters.py
```diff
@@ -0,0 +1,49 @@
...
+            arr = tensor_to_ndarray(pb)
+            var = tf.Variable(name=name, initial_value=arr, trainable=True)
+            self.non_embedding_params[name] = var
+
+    def _init_embedding_params(self, embeddings_pb):
+        for pb in embeddings_pb:
+            table = create_embedding_table(pb)
+            self.embedding_params[table.name] = table
+
+    def clear(self):
```
skydoorkai: When ` clear` will be used besides tests?

elasticdl/python/ps/parameters.py
```diff
@@ -0,0 +1,49 @@
...
+        self.embedding_params = {}
+
+    def get_embedding_param(self, name, indices):
+        if name not in self.embedding_params:
+            raise ValueError(
+                "Please initialize embedding param %s first!", name
+            )
+        return self.embedding_params[name].get(indices)
+
+    def set_embedding_param(self, name, indices, values):
```
mhaoli, line 32: In my opinion, we could also provide a `set_embedding_param` interface that takes a tensor, so no need to use three arguments and looks cleaner. 

But I'm not sure whether we need to support two interfaces, one takes a tensor, another takes these three arguments. The interface which takes a tensor is more convenient for RPC servicer to call (when sync replica or read from checkpoint file). The interface which takes these three arguments is more convenient for optimizer to call.

Or support these two kinds of arguments in one interface?

elasticdl/python/ps/parameters.py
```diff
@@ -0,0 +1,49 @@
...
+        if name not in self.embedding_params:
+            raise ValueError(
+                "Please initialize embedding param %s first!", name
+            )
+        self.embedding_params[name].set(indices, values)
+
+    def init_from_model_pb(self, model_pb):
```
mhaoli, line 39: Would it be better to assume that interface `init_from_model_pb` takes a tensor list and a EmbeddingInfo list? If so, we can put the logic of parsing protocol buffer into RPC servicer.

elasticdl/python/ps/parameters.py
```diff
@@ -0,0 +1,49 @@
...
+
+from elasticdl.python.common.ndarray import tensor_to_ndarray
+from elasticdl.python.ps.embedding_table import create_embedding_table
+
+
+class Parameters(object):
+    def __init__(self):
+        self.init_status = False
+        self.non_embedding_params = {}
+        self.embedding_params = {}
```
terrytangyuan, line 23: Is this ElasticDL embedding params or Keras native embedding params?

elasticdl/python/ps/parameters.py
```diff
@@ -0,0 +1,49 @@
...
+    def _init_non_embedding_params(self, tensors_pb):
+        for pb in tensors_pb:
+            name = pb.name
+            arr = tensor_to_ndarray(pb)
+            var = tf.Variable(name=name, initial_value=arr, trainable=True)
+            self.non_embedding_params[name] = var
+
+    def _init_embedding_params(self, embeddings_pb):
+        for pb in embeddings_pb:
+            table = create_embedding_table(pb)
```
terrytangyuan: Not necessary to create a separate variable for this

elasticdl/python/ps/parameters.py
QiJune: Since `Tensor` has these three fields, we could reuse the same interface.

```python
a = Tensor()
p = Parameters()
p.set_embedding_param(a.name, a,indices, a.values)
```

If we really need an interface which takes `Tensor` as an argument, we could add it later.

elasticdl/python/ps/parameters.py
QiJune: It's ElasticDL embedding params. The key is an embedding layer name, the value is a n`EmbeddingTable` object. I will add some comments in the class `Parameters`.

elasticdl/python/ps/parameters.py
QiJune: Done

elasticdl/python/ps/parameters.py
QiJune: Currently, it is just used in unit tests. I will remove it.

elasticdl/python/ps/parameters.py
QiJune: I think that `Parameters` is actually an in-memory data structure of `proto::Model`.  We could deserialize from a model_pb, and create a parameters object. We could also serialize a parameters object directly to a model_pb file, for the next step serving in C++. 

So it makes sense that we have an interface which could init a parameters object directly from model_pb.

elasticdl/python/ps/parameters.py
mhaoli: By using `Tensor` data structure, we can make the interface cleaner, and this is why I think we nedd a `Tensor` data structure. Otherwise, we add a `Tensor` data structure and change current code from
```python
# current code
# RPC servicer
tensor_pb = get_tensor_pb_from_workers()
name, values, indices = tensor_pb_to_ndarray(tensor_pb)
parameter_instance.set(name, values, indices)
```
to something like
```python
tensor_pb = get_tensor_pb_from_workers()
tensor = tensor_pb_to_tensor(tensor_pb)
parameter_instance_set(tensor.name, tensor.values, tensor.indices)
```

QiJune: Make sense. Since we have `Tensor` data structure, we could use it directly.

mhaoli (approved):
Modify interface `set_embedding_param` to take an argument of type `Tensor` after PR #1361 .

Issue #1360, opened 2019-10-24
Tensor Python data structure


mhaoli:
Implement a `Tensor` python data structure as shown in [design doc](https://github.com/sql-machine-learning/elasticdl/blob/develop/docs/designs/ps_design.md#data-structure). Some util functions can be implemented with this data structure, including 
* convert `Tensor` to protobuf message
* parse `Tensor` from protobuf message
* convert `Tensor` to TensorFlow tensor

This issue includes two steps:
- [x] add `Tensor` data structure and util functions
- [x] use `Tensor` and these util functions in master and worker, remove `elasticdl/python/common/ndarray.py`

mhaoli:
Closing this issue because #1386 fixed it.

PR #1361, opened 2019-10-24
Implement Tensor python data structure and related util functions

mhaoli:
Part of #1360 . This PR is step 1.

This PR implements `Tensor` python data structure and related util functions. Note that they are slightly differ from design doc. And we may need to modify them in the step 2 of #1360. We will modify design doc when finish step 2 of #1360 .

### DIFF
--- a/elasticdl/python/common/tensor.py
+++ b/elasticdl/python/common/tensor.py
@@ -25,7 +25,7 @@ def serialize_tensor(tensor, tensor_pb):
 
 
 def deserialize_tensor_pb(tensor_pb, tensor):
-    """Create an Tensor instance from Tensor proto message.
+    """Create a Tensor instance from Tensor proto message.
 
     Note: upon return, the input tensor message is reset and underlying buffer
     passed to the returned ndarray.
@@ -64,9 +64,9 @@ class Tensor(object):
     def __init__(self, values=None, indices=None, name=None):
         """
         `Tensor` can save dense tensors and sparse tensors.
-        To save dense tensors, `values` should be `numpy.ndarray` and `indices`
-            should be None.
-        There are two ways to save sparse tensors:
+        To pass in a dense tensor, `values` should be `numpy.ndarray` and
+            `indices` should be None.
+        There are two ways to pass in a sparse tensor:
             * `values` is a `numpy.ndarray` and `indices` is a `numpy.ndarray`.
             * `values` is a `TensorFlow.IndexedSlices` and `indices` is None.
 
@@ -74,7 +74,8 @@ class Tensor(object):
             values: A `numpy.ndarray` or `TensorFlow.IndexedSlices`.
                 If `values` is a `TensorFlow.IndexedSlices`, `indices` should
                 be None.
-            indices: `indices` of `TensorFlow`
+            indices: A `numpy.ndarray` or None.
+            name: A python string.
         """
         self.set(values, indices, name)
 
--- a/elasticdl/python/common/tensor.py
+++ b/elasticdl/python/common/tensor.py
@@ -27,8 +27,8 @@ def serialize_tensor(tensor, tensor_pb):
 def deserialize_tensor_pb(tensor_pb, tensor):
     """Create a Tensor instance from Tensor proto message.
 
-    Note: upon return, the input tensor message is reset and underlying buffer
-    passed to the returned ndarray.
+    Note that the input tensor message is reset and underlying buffer is passed
+    to the returned ndarray.
     """
 
     if not tensor_pb.dim:
@@ -47,7 +47,7 @@ def deserialize_tensor_pb(tensor_pb, tensor):
         )
     tensor.set(
         np.ndarray(shape=tensor_pb.dim, dtype=dtype, buffer=tensor_pb.content),
-        tensor_pb.indices,
+        np.array(tensor_pb.indices),
         tensor_pb.name,
     )
     tensor_pb.Clear()
@@ -83,8 +83,8 @@ class Tensor(object):
         if isinstance(values, tf.IndexedSlices):
             if indices is not None:
                 raise ValueError(
-                    "When creating a Tensor instance with values of type "
-                    "tf.IndexedSlices, indices should be None."
+                    "When creating a Tensor object with values of type "
+                    "tf.IndexedSlices, indices must be None."
                 )
             self.values = values.values.numpy()
             self.indices = values.indices.numpy()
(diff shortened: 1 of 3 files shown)


elasticdl/python/tests/tensor_test.py
```diff
@@ -0,0 +1,186 @@
...
+        def verify(values, name=None, indices=None):
+            tensor = Tensor(values, indices, name)
+            pb = elasticdl_pb2.Tensor()
+            serialize_tensor(tensor, pb)
+            deserialize_tensor_pb(pb, tensor)
+            np.testing.assert_array_equal(values, tensor.values)
+            if indices is not None:
+                np.testing.assert_array_equal(indices, tensor.indices)
+            if name:
+                assert name == tensor.name
```
terrytangyuan: Use `self.assertEqual()`

elasticdl/python/common/tensor.py
```diff
@@ -0,0 +1,116 @@
...
+            name: A python string.
+        """
+        self.set(values, indices, name)
+
+    def set(self, values=None, indices=None, name=None):
+        self.name = name
+        if isinstance(values, tf.IndexedSlices):
+            if indices is not None:
+                raise ValueError(
+                    "When creating a Tensor instance with values of type "
```
terrytangyuan: instance -> object

elasticdl/python/common/tensor.py
```diff
@@ -0,0 +1,116 @@
...
+                    "tf.IndexedSlices, indices should be None."
```
terrytangyuan: should be -> must be

elasticdl/python/common/tensor.py
```diff
@@ -0,0 +1,116 @@
...
+            tensor_pb.dim,
+            len(tensor_pb.content),
+        )
+    tensor = Tensor()
+    tensor.set(
+        np.ndarray(shape=tensor_pb.dim, dtype=dtype, buffer=tensor_pb.content),
+        tensor_pb.indices,
+        tensor_pb.name,
+    )
+    tensor_pb.Clear()
```
terrytangyuan, line 53: Do we need to move this to a finally block?

elasticdl/python/common/tensor.py
```diff
@@ -0,0 +1,116 @@
...
+    if tensor.is_indexed_slices():
+        tensor_pb.indices.extend(tensor.indices)
+    if tensor.name:
+        tensor_pb.name = tensor.name
+
+
+def deserialize_tensor_pb(tensor_pb, tensor):
+    """Create a Tensor instance from Tensor proto message.
+
+    Note: upon return, the input tensor message is reset and underlying buffer
```
terrytangyuan: return -> returning

elasticdl/python/common/tensor.py
```diff
@@ -0,0 +1,116 @@
...
+    passed to the returned ndarray.
```
terrytangyuan: Or move "upon return" completely
passed -> is passed

elasticdl/python/common/tensor.py
QiJune: I think that we should move this line.
tensor_pb is the input of this function. The caller usually not except the tensor_pb is cleared when using this function. The caller is responsible for clearing tensor_pb at some place.

elasticdl/python/common/tensor.py
```diff
@@ -0,0 +1,116 @@
...
+    size = dtype.itemsize
+    for d in tensor_pb.dim:
+        size *= d
+    if size != len(tensor_pb.content):
+        raise ValueError(
+            "Tensor PB size mismatch, dim: %s, len(content): %d",
+            tensor_pb.dim,
+            len(tensor_pb.content),
+        )
+    tensor = Tensor()
```
QiJune: We assume that the tensor object has been created already. Here we should not construct a new Tensor object again.

```python
deserialize_tensor_pb(tensor_pb, tensor):
...
tensor.set(...) # just set directly
```

elasticdl/python/common/tensor.py
mhaoli: @QiJune Good point. In fact, I clear `tensor_pb` because `common.ndarray.tensor_to_ndarray` does, and I made some mistakes because I forget this function will clear `tensor_pb`... 

For me, clearing `tensor_pb` in this function is confusing. Thus I agree to move clearing tensor_pb out. What do you think @terrytangyuan ?

mhaoli: By the way, do we usually need to clean tensor_pb **manually** after parsing it? @QiJune @terrytangyuan

elasticdl/python/common/tensor.py
```diff
@@ -0,0 +1,116 @@
...
+    tensor.set(
+        np.ndarray(shape=tensor_pb.dim, dtype=dtype, buffer=tensor_pb.content),
+        tensor_pb.indices,
```
QiJune: Here, we directly set `tensor_pb.indices` to `tensor.indices`. So the indices type of tensor is actually a Repeated Proto type. This is not we expected. The indices type of tensor should be a 1-D numpy ndarry.

```python
indices = np.array([i, for i in tensor_pb.indices])
tensor.set(..., indices,...)
```

elasticdl/python/tests/tensor_test.py
mhaoli: Done

elasticdl/python/common/tensor.py
mhaoli: Done

elasticdl/python/common/tensor.py
mhaoli: Done

elasticdl/python/common/tensor.py
mhaoli: Done

elasticdl/python/common/tensor.py
mhaoli: Done

elasticdl/python/common/tensor.py
mhaoli: Done

elasticdl/python/common/tensor.py
mhaoli: Good point. And it seems that `indices = np.array(tensor_pb.indices)` works.

mhaoli: Modified. Also added a unit test about the type of `tensor.incides` and `tensor.values`.

elasticdl/python/common/tensor.py
QiJune: In common, tensor_pb is a local variable, it will be deleted by Python Interpreter at some time automatically.
If we really want to clear it right now, we should call `tensor_pb.Clear()` explicitly at some place, but not put it inside a deserialize function.

For example:

```python
deserialize_tensor_pb(tensor_pb, tensor)
tensor_pb.Clear()
```

Anyway, this line should be removed from this function.

mhaoli: Yes, I agree with moving this line out.

terrytangyuan (approved):

Issue #1338, opened 2019-10-23
[PS-1]Add new RPC services definition in elasticdl.proto according to PS design


skydoorkai:
[PS design](https://github.com/sql-machine-learning/elasticdl/blob/develop/docs/designs/ps_design.md#rpc-definition) adds some new RPC services.

PR #1369, opened 2019-10-25
Init PS RPC servicer

### LINKED ISSUE #1341: [PS-1] Initialize PS RPC servicer
Create a Servicer class, and `ps.main` start the RPC service.

skydoorkai:
fix #1341

### DIFF
--- a/elasticdl/proto/elasticdl.proto
+++ b/elasticdl/proto/elasticdl.proto
@@ -142,3 +142,4 @@ service Pserver {
     rpc pull_embedding_vector(google.protobuf.Empty) returns (google.protobuf.Empty);
     rpc push_model(google.protobuf.Empty) returns (google.protobuf.Empty);
     rpc push_gradient(google.protobuf.Empty) returns (google.protobuf.Empty);
+}
--- a/elasticdl/proto/elasticdl.proto
+++ b/elasticdl/proto/elasticdl.proto
@@ -134,3 +134,11 @@ service Master {
     rpc ReportEvaluationMetrics(ReportEvaluationMetricsRequest) returns (ReportEvaluationMetricsResponse);
     rpc ReportTaskResult(ReportTaskResultRequest) returns (google.protobuf.Empty);
 }
+
+// PS service
+// TODO: define messages and replace `google.protobuf.Empty` with correct message.
+service Pserver {
+    rpc pull_variable(google.protobuf.Empty) returns (google.protobuf.Empty);
+    rpc pull_embedding_vector(google.protobuf.Empty) returns (google.protobuf.Empty);
+    rpc push_model(google.protobuf.Empty) returns (google.protobuf.Empty);
+    rpc push_gradient(google.protobuf.Empty) returns (google.protobuf.Empty);
--- a/elasticdl/python/ps/main.py
+++ b/elasticdl/python/ps/main.py
@@ -1,9 +1,18 @@
+import time
+from concurrent import futures
+
+import grpc
+
+from elasticdl.proto import elasticdl_pb2_grpc
 from elasticdl.python.common.args import parse_ps_args
+from elasticdl.python.common.constants import GRPC
 from elasticdl.python.common.log_utils import get_logger
 from elasticdl.python.common.model_utils import (
     get_module_file_path,
     load_module,
 )
+from elasticdl.python.ps.parameters import Parameters
+from elasticdl.python.ps.servicer import PserverServicer
 
 
 def main():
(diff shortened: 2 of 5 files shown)


elasticdl/proto/elasticdl.proto
```diff
@@ -134,3 +134,12 @@ service Master {
     rpc ReportEvaluationMetrics(ReportEvaluationMetricsRequest) returns (ReportEvaluationMetricsResponse);
     rpc ReportTaskResult(ReportTaskResultRequest) returns (google.protobuf.Empty);
 }
+
+// PS service
+// TODO: define messages and replace `google.protobuf.Empty` with correct message.
+service Pserver {
```
terrytangyuan, line 140: How about ParameterServerService?

elasticdl/python/tests/pserver_servicer_test.py
```diff
@@ -0,0 +1,100 @@
...
+        )
+        server.add_insecure_port("[::]:{}".format(port))
+        return server
+
+    @staticmethod
+    def create_stub(port):
+        addr = "localhost:%d" % port
+        channel = grpc.insecure_channel(
+            addr,
+            options=[
```
terrytangyuan: Maybe move this to a constant in this file so it's reused

elasticdl/proto/elasticdl.proto
skydoorkai: proto will generate a function `elasticdl_pb2_grpc.add_PserverServicer_to_server` if named `Pserver`. 
If changed to `ParameterServerService`, it will be `elasticdl_pb2_grpc.add_ParameterServerServiceServicer_to_server`.  That is too long and reduntant.

elasticdl/python/tests/pserver_servicer_test.py
skydoorkai: Moved to class attributes for reuse.

elasticdl/python/ps/main.py
```diff
@@ -16,4 +25,40 @@ def main():
     optimizer = model_module[args.optimizer]()
 
     logger.info("Starting PS pod with optimizer as %s", optimizer._name)
-    # TODO: create `Parameters` class instance, and start PS RPC servicer.
+
```
mhaoli: Would it be better if the main function keeps the same style as `master.main`?

elasticdl/python/tests/pserver_servicer_test.py
```diff
@@ -0,0 +1,100 @@
...
+                ("grpc.max_send_message_length", GRPC.MAX_SEND_MESSAGE_LENGTH),
+                (
+                    "grpc.max_receive_message_length",
+                    GRPC.MAX_RECEIVE_MESSAGE_LENGTH,
+                ),
+            ],
+        )
+
+    def create_server(
```
mhaoli: Maybe move this function out of class `PserverServicerTest`?

elasticdl/proto/elasticdl.proto
chunyang-wen: I think xxxxService is better than `Pserver`.

elasticdl/python/ps/servicer.py
```diff
@@ -0,0 +1,38 @@
...
+    ):
+        self._parameters = parameters
+        self._grads_to_wait = grads_to_wait
+        self._optimizer = optimizer
+        self._lr_staleness_modulation = lr_staleness_modulation
+        self._use_async = use_async
+        self._version = 0
+
+    def pull_variable(self, request, _):
+        # TODO: implement this RPC service
```
chunyang-wen, line 25: What does `_` mean here?

chunyang-wen: It seems that there are too many `push_*` and `pull_*` functions here. Can we wrap them and use a parameter like `data_type` to differentiate them.

elasticdl/python/tests/pserver_servicer_test.py
```diff
@@ -0,0 +1,100 @@
...
+            res = stub.pull_embedding_vector(req)
+            self.assertEqual(res, req)
+            res = stub.push_model(req)
+            self.assertEqual(res, req)
+            res = stub.push_gradient(req)
+            self.assertEqual(res, req)
+        except Exception:
+            self.assertTrue(False)
+
+        server.stop(0)
```
chunyang-wen: Cleaning up operations can be done in the `tearDwon` function which is similar to `setUp`

skydoorkai: `server` is not created in setUp, but inside tests, as the init arguments may be different in different tests.

skydoorkai: assign server into class attribute, and `teatDown` can be used.

elasticdl/python/ps/servicer.py
skydoorkai: This is for rpc service. Need to consistent with the names defined in proto file.

elasticdl/python/ps/main.py
skydoorkai: Re-org `ps.main` to be consistent with `master.main`.

elasticdl/python/ps/servicer.py
skydoorkai: `_` should be `context`. As it is not used or even implemented in Python GRPC, `_` is used.

elasticdl/proto/elasticdl.proto
skydoorkai: [Another issue](https://github.com/sql-machine-learning/elasticdl/issues/1338) is responsible for adding message and rpc services in the proto file.
We can modify it if needed in that PR.

mhaoli (approved):

Issue #1374, opened 2019-10-25
A ValueError will be raised when the input dtype of ` elasticdl.Embedding call` is tf.float32.


workingloong:
ValueError: Tensor conversion requested dtype int64 for Tensor with dtype float32: <tf.Tensor 'embedding_2/Reshape:0' shape=(640,) dtype=float32>
![image](https://user-images.githubusercontent.com/18071380/67551113-b185a380-f73a-11e9-992d-8b644c5df943.png)

PR #1375, opened 2019-10-25
Cast input to int64 for embedding `call`

workingloong:
We should check the input dtype for elasticdl.layers.Embedding `call` and cast to int64. #1374

### DIFF
--- a/elasticdl/python/elasticdl/layers/embedding.py
+++ b/elasticdl/python/elasticdl/layers/embedding.py
@@ -186,6 +186,9 @@ class Embedding(tf.keras.layers.Layer):
         return batch_embedding
 
     def call(self, input):
+        if input.dtype is not tf.int64:
+            input = tf.cast(input, tf.int64)
+
         if isinstance(input, tf.SparseTensor):
             return self._sparse_input_call(input)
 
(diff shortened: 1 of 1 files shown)


elasticdl/python/elasticdl/layers/embedding.py
```diff
@@ -186,6 +186,9 @@ def _record_gradients(self, batch_embedding, ids):
         return batch_embedding
 
     def call(self, input):
+        if input.dtype is not tf.int64:
```
skydoorkai: No need to use `if`, just `tf.cast`?

workingloong: Nonuse of `tf.cast` may be better when input dtype is `tf.int64`. The operation in Keras is https://github.com/tensorflow/tensorflow/blob/1cf0898dd4331baf93fe77205550f2c2e6c90ee5/tensorflow/python/keras/layers/embeddings.py#L181-L182 

I wonder whether we should cast tf.int32 to tf.int64.

workingloong: Only `tf.int64` works because we defined batch_ids with `tf.int64`. So, just `tf.cast`.

terrytangyuan (approved):

PR #1380, opened 2019-10-25
Do not save ElasticDL embedding's variables in the master

### LINKED ISSUE #1379: Do not save ElasticDL embedding's variables in the master
In current code, ElasticDL embedding layer create trainable variables and use them only for training under graph mode (i.e. under tf.function decorator). These variables are used for training and thus should only be created in workers. 

However, now ElasticDL embedding layer creates these variables in `__init__` function. The master (or PS) creates these variables when it creates a model instance, and then save these variables in `self._model`. 

I propose that ElasticDL embedding layer should not create these variables in `__init__` function. Instead, creating them before the training process only in worker is enough.

mhaoli:
Fixes #1379.

### DIFF
--- a/elasticdl/python/elasticdl/layers/embedding.py
+++ b/elasticdl/python/elasticdl/layers/embedding.py
@@ -66,17 +66,19 @@ class Embedding(tf.keras.layers.Layer):
         # `tf.Variable` requires initial value if shape has `None` dimension.
         self._embedding_and_ids_graph = []
 
-    def init_for_graph_mode(self):
+    def _init_for_graph_mode(self):
         self._embedding_and_ids_graph = [
             EmbeddingAndIds(
                 batch_embedding=tf.Variable(
-                    initial_value=tf.zeros((1, self.output_dim)),
+                    # In some cases, `tf.Variable` requires that initial value
+                    # is callable.
+                    initial_value=lambda: tf.zeros((1, self.output_dim)),
                     shape=tf.TensorShape((None, self.output_dim)),
                     dtype=tf.float32,
                     trainable=True,
                 ),
                 batch_ids=tf.Variable(
-                    initial_value=tf.zeros((1, 1), dtype=tf.int64),
+                    initial_value=lambda: tf.zeros((1, 1), dtype=tf.int64),
                     shape=tf.TensorShape(None),
                     dtype=tf.int64,
                     trainable=False,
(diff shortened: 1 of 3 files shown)


terrytangyuan (approved):

elasticdl/python/elasticdl/layers/embedding.py
```diff
@@ -186,6 +189,17 @@ def _record_gradients(self, batch_embedding, ids):
         return batch_embedding
 
     def call(self, input):
+        if (
+            self.tape
+            and not tf.executing_eagerly()
+            and not self._embedding_and_ids_graph
+        ):
+            raise RuntimeError(
+                "Call `init_for_graph_mode` for each ElasticDL embedding"
```
terrytangyuan: Maybe just init_for_graph_mode()

elasticdl/python/elasticdl/layers/embedding.py
```diff
@@ -186,6 +189,17 @@ def _record_gradients(self, batch_embedding, ids):
...
+                "layer before training model in graph mode.(i.e. under"
```
terrytangyuan: Move the period in "mode." to the end of this sentence or just remove this period

elasticdl/python/elasticdl/layers/embedding.py
```diff
@@ -64,6 +64,9 @@ def __init__(
         # `None` dimension. This is because they have different shapes in
         # different iterations.
         # `tf.Variable` requires initial value if shape has `None` dimension.
+        self._embedding_and_ids_graph = []
+
+    def init_for_graph_mode(self):
```
terrytangyuan: Should we detect whether it's graph mode and then just run this initializing function implicitly so we don't have to call `init_for_graph_mode()` manually?

terrytangyuan (changes requested):
See my comments above

elasticdl/python/elasticdl/layers/embedding.py
ywskycn: Agree... a result of writing a lot of documents... :D

ywskycn (approved):

elasticdl/python/elasticdl/layers/embedding.py
mhaoli: Good idea. Done

terrytangyuan (approved):

PR #1382, opened 2019-10-28
Add model handler to process model for different distribution strategies.

workingloong:
Create the model handler to process model to train in different distribution strategies. For example, ParameterServerModelHandler will replace the Keras Embedding layer with an ElasticDL Embedding layer in ParameterServerStrategy.  #1392


elasticdl/python/common/args.py
```diff
@@ -455,6 +455,14 @@ def add_common_args_between_master_and_worker(parser):
         help="The data reader parameters in a string separated by semi-colon "
         'used to instantiate the data reader, e.g. "param1=1; param2=2"',
     )
+    parser.add_argument(
+        "--distribution_strategy",
+        type=str,
+        default="",
```
chunyang-wen: a better default value?

elasticdl/python/common/model_handler.py
```diff
@@ -0,0 +1,111 @@
+import abc
+
+import tensorflow as tf
+
+from elasticdl.python.common.log_utils import default_logger as logger
+from elasticdl.python.elasticdl.layers.embedding import Embedding
+
+
+class ModelHander(metaclass=abc.ABCMeta):
```
terrytangyuan: Handler

workingloong: Done.

elasticdl/python/common/args.py
workingloong: "ParameterServerStrategy"? The only supported distribution strategy for the moment.

skydoorkai: yes, need a working default value.

brightcoder01: Maybe default is master center storage?

elasticdl/python/common/model_handler.py
```diff
@@ -0,0 +1,111 @@
...
+        Replace the tf.keras.layers.Embedding layer in the model with
+        an elasticdl.layers.Embedding layer in ParameterServerStrategy.
+        """
+        if type(model) == tf.keras.Sequential or model._is_graph_network:
+            model = self._clone_model_for_sequential_and_functional(model)
+        else:
+            model = self._replace_embedding_attribute_for_subclass(model)
+        return model
+
+    def get_saved_model_from_trained_model(self, model, dataset):
```
brightcoder01: get_model_to_export_from_trained_model

elasticdl/python/common/model_handler.py
```diff
@@ -0,0 +1,111 @@
...
+                    input_length=layer.input_length,
+                )
+                return edl_embedding_layer
+            return layer
+
+        return tf.keras.models.clone_model(
+            model, clone_function=_clone_function
+        )
+
+    def _replace_embedding_attribute_for_subclass(self, model):
```
brightcoder01: _replace_embedding_attribute_for_subclass => _replace_embedding_attributes_for_subclass

elasticdl/python/common/model_handler.py
```diff
@@ -0,0 +1,111 @@
...
+        """
+
+        def _clone_function(layer):
+            if type(layer) == tf.keras.layers.Embedding:
+                logger.info(
+                    "Replace Keras Embedding with \
+                            ElasticDL Embedding"
+                )
+                edl_embedding_layer = Embedding(
+                    output_dim=layer.output_dim,
```
brightcoder01, line 89: We need add input_dim in edl_embedding_layer. If user define the model using native keras embedding layer, user expect the vocab size = input_dim. edl_embedding_layer should return error if the id > input_dim.

elasticdl/python/common/args.py
mhaoli: Agree with master center storage because parameter server strategy is not ready. Also should check consistency with argument "num_ps_pods".

elasticdl/python/common/model_handler.py
```diff
@@ -0,0 +1,111 @@
...
+        """
+        pass
+
+    @classmethod
+    def get_model_handler(cls, distribution_strategy=None):
+        """
+        Create a model handler to process the model for the
+        distributed strategy.
+        """
+        if distribution_strategy == "ParameterServerStrategy":
```
mhaoli: Maybe better to move string "ParameterServerStrategy" to a constant.

elasticdl/python/common/model_handler.py
mhaoli: `get_model_to_export_from_trained_model` seems a little strange. Would `get_model_for_tf_save_model` be better?
Also we can rename `get_train_model_for_elasticdl` to `get_model_for_elasticdl_train`.

elasticdl/python/common/args.py
workingloong: "master center storage" may not be accurate. The embedding parameters will locate in Parameter Servers. The difference is that the parameters of keras.Embedding table will all locate a single pod and the parameters of elasticdl.Embedding table will distribute all PS pods.  Maybe the default value is "small_embedding"?

elasticdl/python/common/model_handler.py
```diff
@@ -0,0 +1,111 @@
...
+        Clone a new model with elasticdl.layers.Embedding for
+        Sequential and functional API model.
+        """
+
+        def _clone_function(layer):
+            if type(layer) == tf.keras.layers.Embedding:
+                logger.info(
+                    "Replace Keras Embedding with \
+                            ElasticDL Embedding"
+                )
```
mhaoli: The output of this log info will be

```
Replace Keras Embedding with               ElasticDL Embedding             
```

It looks strange. Better to use following code:
```python
logger.info(
    "Replace Keras Embedding with "
    "ElasticDL Embedding"
)
```

elasticdl/python/common/model_handler.py
```diff
@@ -0,0 +1,111 @@
...
+        model.
+        """
+        # TODO
+        pass
+
+    def _clone_model_for_sequential_and_functional(self, model):
+        """
+        Clone a new model with elasticdl.layers.Embedding for
+        Sequential and functional API model.
+        """
```
mhaoli, line 83: If user writes model in functional api, here needs to use twice the size of the model. Maybe add a note in the docstring of this function.

elasticdl/python/tests/test_module.py
```diff
@@ -13,6 +13,33 @@ def custom_model():
...
+
+def feature_columns_fn():
+    age = tf.feature_column.numeric_column("age", dtype=tf.int64)
+    education = tf.feature_column.categorical_column_with_hash_bucket(
+        "education", hash_bucket_size=4
+    )
+    education_one_hot = tf.feature_column.indicator_column(education)
+    return [age, education_one_hot]
+
+
```
mhaoli, line 42: Because we may have many models only for unit tests. Up to now, there are two model files `test_module.py` and `embedding_test_module.py`. 

In my opinion, we can save each model in a separate file, so we can create a new file for these functions.

Or maybe you can take a look at `embedding_test_module.py` to see whether it is convenient to use them.

elasticdl/python/worker/worker.py
```diff
@@ -181,9 +188,10 @@ def get_model(self, version, method):
 
         # Assumes all trainable variables exist in model.param.
         for tensor_pb in model.param:
-            self._non_embed_vars[tensor_pb.name].assign(
-                tensor_to_ndarray(tensor_pb)
-            )
+            if tensor_pb.name in self._non_embed_vars:
+                self._non_embed_vars[tensor_pb.name].assign(
+                    tensor_to_ndarray(tensor_pb)
+                )
```
mhaoli: This line has a conflict with the above comment. And just ignore the variables that are not in `worker._non_embed_vars` without doing anything seems a little dangerous.

In fact, it seems that this has already be fixed in #1380 (this PR is merged after you submit your PR). So maybe remove this `if`?

elasticdl/python/common/model_handler.py
workingloong: Except the embedding layer, other layers are shared between the original model and the cloned model.

elasticdl/python/common/model_handler.py
workingloong: Further simplicity, `get_model_to_train`  and `get_model_to_export` ?

elasticdl/python/common/model_handler.py
mhaoli: Got it!

elasticdl/python/common/args.py
skydoorkai: If we want to have "master center storage" after the new PS implementation, we can create a [ParameterService](https://github.com/sql-machine-learning/elasticdl/blob/9bb018f3756ac3f707b44524d4fadc64e15080e2/elasticdl/python/ps/parameter_server.py#L17) instance in master.
Argument "num_ps_pods" can be used for ParameterServerStrategy only.

workingloong: > Agree with master center storage because parameter server strategy is not ready. Also should check consistency with argument "num_ps_pods".

Now, a single parameter server is on the master pod which is also the parameter server strategy.

elasticdl/python/elasticdl/layers/embedding.py
```diff
@@ -174,6 +179,15 @@ def lookup_embedding(self, unique_ids):
         embedding_vectors = np.concatenate(embedding_vectors, axis=0)
         return embedding_vectors.reshape((len(keys), self.output_dim))
 
+    def _check_id_valid(self, ids):
+        if self.input_dim is not None and np.sum(ids >= self.input_dim) > 0:
```
brightcoder01: We can try more efficient way instead of np.sum(ids >= self.input_dim) and ids[ids >= self.input_dim][0]

workingloong: Done,  use np.argmax(ids >= self.input_dim).

brightcoder01 (approved):

PR #1383, opened 2019-10-28
create legacy ps related services only when num_ps_pods==0

### LINKED ISSUE #1370: Master uses legacy PS only when `num_ps_pods==0`
When `num_ps_pods == 0`, the master uses the legacy PS to create `MasterServicer`, `embedding_service`, `evaluation_service`, `checkpoint_service`.

For `num_ps_pods > 0 > 0`, the master will use the new PS design and create PS pods.

The old `MasterServicer` and `embedding_service` will be implemented in PS pods. 
`evaluation_service` and `checkpoint_service` will be redesigned in other projects.

Note that the legacy PS codes will be removed after we have implemented the new PS.

skydoorkai:
fix #1370

### DIFF
--- a/elasticdl/python/master/master.py
+++ b/elasticdl/python/master/master.py
@@ -105,12 +105,12 @@ class Master(object):
         )
         self.optimizer = self.model_module[args.optimizer]()
 
-        # checkpoint_service, evaluation_service and embedding_service will be
-        # redesigned after distributed PS is implemented
+        # TODO: checkpoint_service, evaluation_service and embedding_service
+        #       will be redesigned after distributed PS is implemented
         if self.num_ps_pods:
             self.checkpoint_service = None
             self.evaluation_service = None
-            self.self.embedding_service_endpoint = None
+            self.embedding_service_endpoint = None
             self.embedding_dims = None
         else:
             # Initialize checkpoint service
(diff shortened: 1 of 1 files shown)


elasticdl/python/master/master.py
```diff
@@ -104,21 +105,33 @@ def __init__(self, args):
         )
         self.optimizer = self.model_module[args.optimizer]()
 
-        # Initialize checkpoint service
-        self.checkpoint_service = self._create_checkpoint_service(args)
+        # checkpoint_service, evaluation_service and embedding_service will be
```
mhaoli: Do we need to mark this line as a TODO?

elasticdl/python/master/master.py
```diff
@@ -104,21 +105,33 @@ def __init__(self, args):
...
+        # redesigned after distributed PS is implemented
+        if self.num_ps_pods:
+            self.checkpoint_service = None
+            self.evaluation_service = None
+            self.self.embedding_service_endpoint = None
```
mhaoli: duplicated `self`

elasticdl/python/master/master.py
skydoorkai: added.

elasticdl/python/master/master.py
skydoorkai: fixed.

mhaoli (approved):

PR #1388, opened 2019-10-29
Create and dispatch the SaveModel task after all the training/evaluation/prediction task are completed

### LINKED ISSUE #1346: [Serving] Create and dispatch the SaveModel Task after all training/evaluation task are done.
1. Leverage the existed task dispatch&recover mechanism
2. Create SavedModel with a shard of data (build the model structure to export)
3. Only one worker do this work, other worker keep waiting

brightcoder01:
Fix #1346 
Add a callback list to invoke while the todo and doing list are finished. Pop and invoke a callback to try to supplement more tasks and dispatch them.
The master exit while the task list in the dispatcher are finished and the callback list are all popped and invoked.

### DIFF
--- a/elasticdl/python/master/task_dispatcher.py
+++ b/elasticdl/python/master/task_dispatcher.py
@@ -167,10 +167,10 @@ class _TaskDispatcher(object):
         self._todo.append(save_model_task)
 
     def invoke_task_list_done_callback(self):
-        '''
+        """
         Pop a callback from the list and invoke it.
         If the callback list is empty, return False directly.
-        '''
+        """
         if not self._task_list_done_callbacks:
             return False
 
--- a/elasticdl/python/master/task_dispatcher.py
+++ b/elasticdl/python/master/task_dispatcher.py
@@ -144,7 +144,7 @@ class _TaskDispatcher(object):
         return tasks
 
     def _create_save_model_task(self):
-        # For SavedModel Task, we need a shard of data to build the model
+        # For the SavedModel Task, we need a shard of data to build the model
         shards = self._training_shards
         assert shards is not None
 
--- a/elasticdl/python/master/task_dispatcher.py
+++ b/elasticdl/python/master/task_dispatcher.py
@@ -168,8 +168,8 @@ class _TaskDispatcher(object):
 
     def invoke_task_list_done_callback(self):
         '''
-        Pop a callback and invoke it from the list.
-        If the callback list is empty, return False.
+        Pop a callback from the list and invoke it.
+        If the callback list is empty, return False directly.
         '''
         if not self._task_list_done_callbacks:
             return False
--- a/elasticdl/python/master/task_dispatcher.py
+++ b/elasticdl/python/master/task_dispatcher.py
@@ -173,9 +173,13 @@ class _TaskDispatcher(object):
         if not self._task_list_done_callbacks:
             return False
 
-        callback = self._task_list_done_callbacks.pop()
-        callback()
-        return True
+        with self._lock:
+            if not self._task_list_done_callbacks:
+                return False
+
+            callback = self._task_list_done_callbacks.pop()
+            callback()
+            return True
 
     def get(self, worker_id):
         """Return next (task_id, Task) tuple"""
--- a/elasticdl/python/master/task_dispatcher.py
+++ b/elasticdl/python/master/task_dispatcher.py
@@ -143,7 +143,12 @@ class _TaskDispatcher(object):
         return tasks
 
     def _create_save_model_task(self):
-        # For the SavedModel Task, we need a shard of data to build the model
+        """
+        Build one instance of SaveModel task and add it to todo list.
+        Because we need create a dataset to build the model,
+        we include a shard of data in this task.
+        """
+
         shards = self._training_shards
         assert shards is not None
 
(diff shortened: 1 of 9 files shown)


elasticdl/python/master/task_dispatcher.py
```diff
@@ -49,6 +50,8 @@ def __init__(
             records_per_task: The number of records per task.
             num_epochs: The total number of epochs for the tasks where
                 an epoch is a complete iteration over the shards.
+            need_save_model: Save model or not after all the
+                training/evaluation/prediction task are completed.
```
ywskycn: `Whether to save model after all tasks complete`

elasticdl/python/master/task_dispatcher.py
```diff
@@ -65,6 +68,12 @@ def __init__(
         self._task_id = 0
         self._evaluation_service = None
 
+        # the callback list to invoke while
+        # all the task are completed.
```
ywskycn: `Callbacks to invoke after all tasks complete`

elasticdl/python/master/task_dispatcher.py
```diff
@@ -134,6 +143,41 @@ def create_tasks(self, task_type, model_version=-1):
...
+
+        # Use the first shard of data to do the SavedModel work
+        save_model_task = _Task(
+            shard_name=shard_name,
+            start=start_ind_this_task,
+            end=end_ind_this_task,
+            type=elasticdl_pb2.SAVE_MODEL,
+        )
+
+        self._todo.append(save_model_task)
```
ywskycn, line 171: The `save_model_task` should be invoked at the end of the job, right? But if we insert `save_model_task` into `_todo` at the beginning, it means the task will be executed before training? Do I miss anything here?

brightcoder01: Yes, it will be executed before training if the task is inserted at the head before the training starts. But this case won't happen because SaveModel task is created after all tasks complete.
In this PR, we add a callback to create save model task in _task_list_done_callbacks. The callback will only be invoked when all the training/evaluation/prediction tasks complete

elasticdl/python/master/task_dispatcher.py
brightcoder01: Good suggestion, updated.

elasticdl/python/master/task_dispatcher.py
brightcoder01: Good suggestion, updated.

elasticdl/python/master/task_dispatcher.py
```diff
@@ -134,6 +142,41 @@ def create_tasks(self, task_type, model_version=-1):
...
 
+    def _create_save_model_task(self):
+        # For the SavedModel Task, we need a shard of data to build the model
+        shards = self._training_shards
+        assert shards is not None
+
+        (shard_name, (start_ind_this_shard, num_records_this_shard)) = next(
+            iter(shards.items())
+        )
+        start_ind_this_task = start_ind_this_shard
```
skydoorkai, line 158: No need to assign `start_ind_this_shard` to `start_ind_this_task`, just use it directly in _Task init.

elasticdl/python/master/task_dispatcher.py
```diff
@@ -134,6 +142,41 @@ def create_tasks(self, task_type, model_version=-1):
...
+        save_model_task = _Task(
+            shard_name=shard_name,
+            start=start_ind_this_task,
+            end=end_ind_this_task,
+            type=elasticdl_pb2.SAVE_MODEL,
+        )
+
+        self._todo.append(save_model_task)
+
+    def invoke_task_list_done_callback(self):
```
skydoorkai, line 173: A lock is needed here, as there may be multiple concurrent GetTask rpc calls.

brightcoder01: Good catch, fixed :)
I would choose using the double check *if not self._task_list_done_callbacks* inside and outside the lock for performance consideration.

elasticdl/python/master/task_dispatcher.py
brightcoder01: We also use this variable in 
end_ind_this_task = start_ind_this_shard + min(
            self._records_per_task, num_records_this_shard
        )
I prefer keeping it to make the code more readable.

elasticdl/python/master/task_dispatcher.py
```diff
@@ -134,6 +142,45 @@ def create_tasks(self, task_type, model_version=-1):
                 self._todo.extend(tasks)
         return tasks
 
+    def _create_save_model_task(self):
+        # For the SavedModel Task, we need a shard of data to build the model
```
workingloong: Add a docstring and place the note in the docstring.
"we need a shard of data to build the model" -> "we need create a dataset with a shard of data to build the model"

elasticdl/python/master/task_dispatcher.py
```diff
@@ -49,6 +50,8 @@ def __init__(
             records_per_task: The number of records per task.
             num_epochs: The total number of epochs for the tasks where
                 an epoch is a complete iteration over the shards.
+            need_save_model: Whether to save model after all tasks
+                are completed.
```
workingloong, line 54: "all tasks" -> "all training tasks". Save model task is also a task.

elasticdl/python/master/task_dispatcher.py
brightcoder01: Good suggestion, updated.

workingloong (approved):

PR #1391, opened 2019-10-30
Refact the process on worker to execute training, evaluation and prediction tasks.

### LINKED ISSUE #1390: Separate evaluation tasks from todo list.
If the master inserts evaluation tasks into todo list, all evaluation tasks will be got by a single worker. So the master creates the eval_todo list to save the evaluation tasks when an evaluation job starts. The worker will query an evaluation task for training every mini-batch. The detail is in [Design for ElasticDL Evaluation during Training](https://github.com/sql-machine-learning/elasticdl/blob/develop/docs/designs/evaluation_design.md#the-existing-evaluation-design)

workingloong:
Fix #1390
1. The master creates the eval_todo list to save evaluation tasks when an evaluation job starts.
2. The worker only gets training tasks in tf.data.Dataset generator thread.
3. The worker queries an evaluation task for training every mini-batch and will starts to evaluate if the evaluation task isn't None.


elasticdl/python/worker/worker.py
```diff
@@ -559,47 +563,48 @@ def run(self):
...
+                dataset,
+                Mode.TRAINING,
+                self._task_data_service.data_reader.metadata,
             )
             dataset = dataset.batch(self._minibatch_size).prefetch(1)
             for dataset_batch in dataset:
                 if self._job_type == JobType.TRAINING_WITH_EVALUATION:
-                    if self._process_eval_task_if_needed():
-                        last_minibatch_do_evaluation = True
+                    self._evaluate_only()
```
skydoorkai: Add a comment explaining that this is to give the worker a chance to process an evaluation task during training if the task exists.

workingloong: Yeah!

elasticdl/python/master/task_dispatcher.py
```diff
@@ -131,9 +132,19 @@ def create_tasks(self, task_type, model_version=-1):
             self._todo.extend(tasks)
         else:
             with self._lock:
-                self._todo.extend(tasks)
+                self._eval_todo.extend(tasks)
```
mhaoli: It seems a little confused that we don't need a lock for `_todo`, but we need one for `_eval_todo`. Do you have any idea about this difference?

elasticdl/python/worker/worker.py
```diff
@@ -170,6 +170,16 @@ def get_task(self):
 
         return self._stub.GetTask(req)
 
+    def get_eval_task(self):
```
mhaoli: Can we reuse `get_task()` but add an argument to it?

elasticdl/python/worker/worker.py
```diff
@@ -559,47 +563,50 @@ def run(self):
...
             dataset = self._dataset_fn(
-                dataset, mode, self._task_data_service.data_reader.metadata
+                dataset,
+                Mode.TRAINING,
+                self._task_data_service.data_reader.metadata,
             )
             dataset = dataset.batch(self._minibatch_size).prefetch(1)
             for dataset_batch in dataset:
                 if self._job_type == JobType.TRAINING_WITH_EVALUATION:
-                    if self._process_eval_task_if_needed():
```
mhaoli, line 584: Please leave this logic, because if a worker does some evaluation tasks in the training process, it should pull model instead of training with local model parameters.

workingloong: Yeah, we can get a status from `_evaluate_only`.

elasticdl/python/master/task_dispatcher.py
workingloong: I also don't know the reason for the difference. But, I think we also need a lock to operate `_todo`.

elasticdl/python/worker/worker.py
```diff
@@ -572,47 +567,51 @@ def run(self):
...
-                        last_minibatch_do_evaluation = True
+                    # Give the worker a chance to process an evaluation task
+                    # during training if the task exists
+                    evaluation_task_executed = self._evaluate_only()
```
mhaoli: Two places can modify `evaluation_task_executed`, need to use 
```
evaluation_task_executed = True if self._evaluate_only() else evaluation_task_executed
```

elasticdl/python/master/task_dispatcher.py
mhaoli: Okay, we can take a look at this lock later after this PR.

elasticdl/proto/elasticdl.proto
```diff
@@ -77,6 +77,7 @@ message Model {
 
 message GetTaskRequest {
     int32 worker_id = 1;
+    int32 task_type = 2;
```
brightcoder01: We can use TaskType instead of int32

elasticdl/python/master/task_dispatcher.py
```diff
@@ -207,7 +219,10 @@ def report(self, task_id, success):
                 logger.warning("Unknown task_id: %d" % task_id)
             elif not success:
                 # TODO: keep count of retries.
-                self._todo.append(task)
+                if task.type == elasticdl_pb2.TRAINING:
```
brightcoder01, line 221: I think we need check task.type == elasticdl_pb2.EVALUATION, otherwise we will add the task into _todo.
Because we treat TRAINING/PREDICTION/SAVE_MODEL task in the same way.

elasticdl/python/worker/worker.py
```diff
@@ -544,21 +545,15 @@ def _process_minibatch_and_report(
...
-        elif self._job_type == JobType.EVALUATION_ONLY:
-            mode = Mode.EVALUATION
-        else:
-            mode = Mode.TRAINING
 
         # The worker needs to get model from PS if
         # `train_with_local_model=False`. This happens when:
         #     processing first minibatch
-        #     last minibatch is evaluation task
+        #     any evaluation task has been executed before the minibatch
```
brightcoder01: executed before the minibatch => executed just before this minibatch

elasticdl/python/master/task_dispatcher.py
workingloong: We cannot use lock here because the lock has been used by `get` or `get_eval_task` in TaskDispatcher. A deadlock will happen if the lock is used for `_todo`. And I think we can remove the lock for `_eval_to_do` too.

terrytangyuan (approved):

PR #1393, opened 2019-10-30
pull_variable RPC implementation

### LINKED ISSUE #1344: [PS-1] RPC service `pull_variable` implementation

skydoorkai:
fix #1344

### DIFF
--- a/elasticdl/python/ps/servicer.py
+++ b/elasticdl/python/ps/servicer.py
@@ -3,6 +3,7 @@ import threading
 from google.protobuf import empty_pb2
 
 from elasticdl.proto import elasticdl_pb2, elasticdl_pb2_grpc
+from elasticdl.python.common.dtypes import dtype_numpy_to_tensor
 
 
 class PserverServicer(elasticdl_pb2_grpc.PserverServicer):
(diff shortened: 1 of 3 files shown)


elasticdl/python/ps/servicer.py
```diff
@@ -25,8 +26,31 @@ def __init__(
...
+        """
+        Response with all non-embedding parameters if initialized.
+        """
+        res = elasticdl_pb2.PullVariableResponse()
+        if not self._parameters.init_status:
+            res.model_init_status = False
+            return res
+
+        # Only sync-SGD needs lock
+        # TODO: use a read-write lock to support multiple concurrent reads
```
mhaoli, line 38: What does this TODO means? There is already a lock when using sync-SGD.

elasticdl/python/ps/servicer.py
```diff
@@ -25,8 +26,31 @@ def __init__(
...
+        if not self._use_async:
+            self._lock.acquire()
+        res.model.version = self._parameters.version
+        for name in self._parameters.non_embedding_params:
+            tensor = res.model.param.add()
+            tensor.name = name
+            var = self._parameters.non_embedding_params[name]
```
mhaoli: How about using `for name, var in self._parameters.non_embedding_params.items()`. Through this way, it does not need to query the python dictionary and hence it is faster.

elasticdl/python/ps/servicer.py
skydoorkai: In sync-SGD, if PS is updating parameters with gradients, it should hold on the pull_variable from workers until the updates are done. Otherwise, workers may get inconsistent model variables. In no updates happening, we should support concurrent reads.

elasticdl/python/ps/servicer.py
skydoorkai: Fixed.

elasticdl/python/ps/servicer.py
mhaoli: Got it. How about modify the TODO to "If no updates happening, do not lock and support concurrent reads"?

skydoorkai: If no update happening, the lock is still needed to prevent updates in the middle of this read service.
A read-write lock is needed. If no write, multiple reads. If write needed, all new reads are blocked,  wait for all existing reads done, then write.

mhaoli (approved):

PR #1394, opened 2019-10-30
Add fail task tolerance

chunyang-wen:
Add support for report error if bad data appears in batch.

### DIFF
--- a/elasticdl/python/common/args.py
+++ b/elasticdl/python/common/args.py
@@ -478,9 +478,9 @@ def add_common_args_between_master_and_worker(parser):
         'e.g. "ParameterServerStrategy"',
     )
     parser.add_argument(
-        "--record_failure_tolerance",
-        type=int,
-        default=0,
+        "--record_failure_tolerance_percentage",
+        type=float,
+        default=0.0,
         help="The percentage of records which is allowed to fail before"
         " failing the job, e.g. 10, means at most 10% failure is allowed.",
     )
--- a/elasticdl/python/common/args.py
+++ b/elasticdl/python/common/args.py
@@ -477,13 +477,6 @@ def add_common_args_between_master_and_worker(parser):
         "according to the distributed strategy, "
         'e.g. "ParameterServerStrategy"',
     )
-    parser.add_argument(
-        "--record_failure_tolerance_percentage",
-        type=float,
-        default=0.0,
-        help="The percentage of records which is allowed to fail before"
-        " failing the job, e.g. 10, means at most 10% failure is allowed.",
-    )
 
 
 def parse_master_args(master_args=None):
--- a/elasticdl/python/common/args.py
+++ b/elasticdl/python/common/args.py
@@ -476,6 +476,12 @@ def add_common_args_between_master_and_worker(parser):
         help="Master will use a distribution policy on a list of devices "
         "according to the distributed strategy, "
         'e.g. "ParameterServerStrategy"',
+    parser.add_argument(
+        "--record_failure_tolerance",
+        type=int,
+        default=0,
+        help="The percentage of records which is allowed to fail before"
+        " failing the job, e.g. 10, means at most 10% failure is allowed.",
     )
 
 
--- a/elasticdl/python/common/args.py
+++ b/elasticdl/python/common/args.py
@@ -476,6 +476,7 @@ def add_common_args_between_master_and_worker(parser):
         help="Master will use a distribution policy on a list of devices "
         "according to the distributed strategy, "
         'e.g. "ParameterServerStrategy"',
+    )
     parser.add_argument(
         "--record_failure_tolerance",
         type=int,
--- a/elasticdl/python/worker/task_data_service.py
+++ b/elasticdl/python/worker/task_data_service.py
@@ -12,11 +12,7 @@ from elasticdl.python.data.dataset_utils import create_dataset_from_tasks
 
 class TaskDataService(object):
     def __init__(
-        self,
-        worker,
-        training_with_evaluation,
-        data_reader_params=None,
-        record_failure_tolerance_percentage=0.0,
+        self, worker, training_with_evaluation, data_reader_params=None,
     ):
         self._worker = worker
         self._training_with_evaluation = training_with_evaluation
@@ -32,9 +28,6 @@ class TaskDataService(object):
             self.data_reader = create_data_reader(data_origin=None)
         self._warm_up_task = None
         self._has_warmed_up = False
-        self._record_failure_tolerance_percentage = (
-            record_failure_tolerance_percentage
-        )
 
     def _reset(self):
         """
--- a/elasticdl/python/worker/task_data_service.py
+++ b/elasticdl/python/worker/task_data_service.py
@@ -59,7 +59,7 @@ class TaskDataService(object):
         if err_msg:
             self._failed_record_count += count
             logger.warning(
-                "Process records of task contains " "error: %s", err_msg
+                "Process records of task contains error: %s", err_msg
             )
         if self._pending_tasks:
             task = self._pending_tasks[0]
@@ -79,6 +79,7 @@ class TaskDataService(object):
                         t=total_record_num,
                     )
                     self.report_task_done(task.task_id, msg)
+                    return
             elif self._failed_record_count != 0:
                 self.report_task_done(task.task_id, err_msg)
             with self._lock:
--- a/elasticdl/python/worker/task_data_service.py
+++ b/elasticdl/python/worker/task_data_service.py
@@ -1,4 +1,5 @@
 import threading
+from collections import deque
 
 import tensorflow as tf
 
@@ -36,10 +37,10 @@ class TaskDataService(object):
         """
         Reset pending tasks and record counts
         """
-        self._record_count = 0
+        self._cur_record_count = 0
         self._reported_record_count = 0
         self._failed_record_count = 0
-        self._pending_tasks_with_counts = []
+        self._pending_tasks = deque()
         self._current_task = None
 
     def get_current_task(self):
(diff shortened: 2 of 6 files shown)


skydoorkai:
If processed on each worker, the percentage calculation is not correct, as each worker only processes a portion of the data.

mhaoli:
> If processed on each worker, the percentage calculation is not correct, as each worker only processes a portion of the data.

Agree. Better that workers report failed record number to the master, and the master calculates fail rate.

elasticdl/python/worker/task_data_service.py
```diff
@@ -27,44 +32,67 @@ def __init__(
...
         self._reported_record_count += count
-        if (
-            len(self._pending_tasks_with_counts)
-            and self._reported_record_count
-            >= self._pending_tasks_with_counts[0][1]
-        ):
+        if err_msg:
+            self._failed_record_count += count
+            logger.warning(
+                "Process records of task contains " "error: %s", err_msg
```
brightcoder01: The error msg is "Process records of task contains " "error: %s"

Maybe unnecessary " " inside?

elasticdl/python/worker/task_data_service.py
```diff
@@ -27,44 +32,67 @@ def __init__(
...
+                if (
+                    self._pending_tasks
+                    and self._reported_record_count >= total_record_num
                 ):
-                    task, _ = self._pending_tasks_with_counts.pop(0)
-                    self._worker.report_task_result(task.task_id, err_msg)
-                if len(self._pending_tasks_with_counts):
-                    self._current_task = self._pending_tasks_with_counts[0][0]
+                    task = self._pending_tasks.popleft()
+                    self.report_task_done(task.task_id, err_msg)
```
brightcoder01: Will we report twice for the same task in this method?

chunyang-wen:
@LiMinghao1994 @skydoorkai The failure rate = failure_records_in_task / records_in_task

It determines whether this task is successfully handled. The master should have its strategy to decide whether the job is successful by the rate of successful tasks.

Usually, there should be no failures or only a small number of failures due to bad data.

elasticdl/python/worker/task_data_service.py
chunyang-wen: Fixed

elasticdl/python/worker/task_data_service.py
chunyang-wen: Fixed

elasticdl/python/common/args.py
```diff
@@ -463,6 +463,13 @@ def add_common_args_between_master_and_worker(parser):
         "according to the distributed strategy, "
         'e.g. "ParameterServerStrategy"',
     )
+    parser.add_argument(
+        "--record_failure_tolerance",
```
terrytangyuan: If this is a percentage, then maybe rename this to "failure_tolerance_percentage"?

elasticdl/python/common/args.py
```diff
@@ -463,6 +463,13 @@ def add_common_args_between_master_and_worker(parser):
...
+        type=int,
```
terrytangyuan: Should this be float, e.g. 10.1?

elasticdl/python/worker/task_data_service.py
```diff
@@ -159,11 +188,9 @@ def _gen(self):
                     self._pending_save_model_task = task
                     continue
 
-                self._record_count += task.end - task.start
-                self._pending_tasks_with_counts.append(
-                    (task, self._record_count)
-                )
-                if len(self._pending_tasks_with_counts) == 1:
+                self._cur_record_count = task.end - task.start
```
skydoorkai: When getting the second task, self._cur_record_count will be overwritten and results in error later.

chunyang-wen: Tasks are handled sequentially?

elasticdl/python/worker/task_data_service.py
```diff
@@ -27,44 +32,68 @@ def __init__(
...
+            elif self._failed_record_count != 0:
+                self.report_task_done(task.task_id, err_msg)
             with self._lock:
-                while (
-                    len(self._pending_tasks_with_counts)
-                    and self._reported_record_count
-                    >= self._pending_tasks_with_counts[0][1]
+                if (
+                    self._pending_tasks
+                    and self._reported_record_count >= total_record_num
```
skydoorkai: reported record count includes all reported records. total  record number only has one task record.  this logic is incorrect.

chunyang-wen: Oops! It seems that the original logic has a similar problem. Because the `_reported_record_count` contains successful batches from different workers.

skydoorkai: The original version `self._pending_tasks_with_counts[0][1]` has all record counts from 1st task. Only the comments are not correct in the original version.

elasticdl/python/common/args.py
chunyang-wen: Fixed

elasticdl/python/common/args.py
chunyang-wen: Fixed

elasticdl/python/worker/task_data_service.py
chunyang-wen: Got it. I have updated the code.

elasticdl/python/worker/task_data_service.py
```diff
@@ -27,44 +32,77 @@ def __init__(
...
-            >= self._pending_tasks_with_counts[0][1]
-        ):
-            with self._lock:
-                while (
-                    len(self._pending_tasks_with_counts)
-                    and self._reported_record_count
-                    >= self._pending_tasks_with_counts[0][1]
+        if self._pending_tasks:
+            task = self._pending_tasks[0]
+            self._reported_record_count[task.task_id] += count
```
skydoorkai: No need to define a dict for `_reported_record_count`. Just define an attribute `self._ _reported_record_count_in_pending_tasks`, and subtract the finished pending task record count in  `handle_next_task`.

chunyang-wen: Fixed

elasticdl/python/worker/task_data_service.py
```diff
@@ -27,44 +32,77 @@ def __init__(
...
+                    > self._record_failure_tolerance_percentage
                 ):
-                    task, _ = self._pending_tasks_with_counts.pop(0)
-                    self._worker.report_task_result(task.task_id, err_msg)
-                if len(self._pending_tasks_with_counts):
-                    self._current_task = self._pending_tasks_with_counts[0][0]
+                    should_handle_next = True
+            elif self._failed_record_count != 0:
+                should_handle_next = True
+            if self._reported_record_count >= total_record_num:
```
skydoorkai: `self._reported_record_count` is total records of all tasks, but `total_record_num` is only for the current pending tasks?
task A with 8 records reports finished 8 records: `self._reported_record_count = 8`, `total_record_num=8`.
Then task B with 8 records reports finished 4 records: `self._reported_record_count = 12`, `total_record_num= 8.
Task B is finished under this condition.

chunyang-wen: The previous design is not correct. We can not use `self._reported_record_count`(current total report) > sum_of_all_record_num. Because we have to make sure that the records of a specific task are all correctly processed. Currently, there can be only one task being processed at the same time. So it is safe to directly use `self._reported_record_count` > `total_record_num` (which means the total num of the task being processed).

skydoorkai: In the previous design,  for pending tasks, if uses `while` loop to determine:
`self._reported_record_count`(current total report) > sum_of_all_record_num_till_this pending_task. 
So it can handle any situation for task size and batch size combination.

In current PR,  `self._reported_record_count` is reset to 0 after a task done, and `if` instead of `while` is used, so it may get wrong when:
1. `self._reported_record_count = 20` after adding the count, and pending_task's record number is 16. Then `self._reported_record_count` is reset to 0, but it should be 4 instead.
2. If task's record number is small and the minibatch number is larger, for example, each task has 32 records, minibatch=64, one minibatch will finish 2 tasks. So `while` loop should be used instead of `if`.

elasticdl/python/worker/task_data_service.py
```diff
@@ -26,44 +32,90 @@ def __init__(
...
-                while (
-                    len(self._pending_tasks_with_counts)
-                    and self._reported_record_count
-                    >= self._pending_tasks_with_counts[0][1]
-                ):
-                    task, _ = self._pending_tasks_with_counts.pop(0)
-                    self._worker.report_task_result(task.task_id, err_msg)
-                if len(self._pending_tasks_with_counts):
-                    self._current_task = self._pending_tasks_with_counts[0][0]
+        if self._pending_tasks:
```
skydoorkai: This methdod `report_record_done` can be simplified:
```
def report_record_done(self, count, err_msg=""):
    self._reported_record_count += count
    if err_msg:
            self._failed_record_count += count

    pending_task_finished = False
    if self._pending_tasks:
         task=  self._pending_tasks[0]
         if self._reported_record_count >= task.end - task.start:
             pending_task_finished = True
         if pending_task_finished:
             with self._lock:
                 while  self._pending_tasks and self._reported_record_count >= self._pending_tasks[0].end - self._pending_tasks[0].start:
                  task = self._pending_tasks.pop(0)
                  self._reported_record_count -= task.end - task.start
                  then report_task_done with some err_msg/exec_counters
       
```

skydoorkai (dismissed):

elasticdl/python/worker/task_data_service.py
chunyang-wen: Updated.

skydoorkai (approved):

PR #1399, opened 2019-10-31
Pass save model task parameter from client CLI to the worker execution

brightcoder01:
1. For the task object, we add extended_config to store some configurations besides the data shards and model versions. We can put various extended parameters in this field. For SaveModelTask, we store saved_model_path.
2. Add a Api to register create_save_model_task callback instead of add it in the constructor of TaskDispatcher.
3. We export the Model to SavedModel instead of store the latest checkpoint after completing all the tasks.


elasticdl/python/master/master.py
```diff
@@ -97,6 +98,10 @@ def __init__(self, args):
             args.num_epochs,
         )
 
+        saved_model_path = args.output
```
workingloong, line 101: Maybe `args.export_dir` better ?

The args of `tf.saved_model.save` is
```
tf.saved_model.save(
    obj,
    export_dir,
    signatures=None,
    options=None
)

File: elasticdl/python/master/master.py [hunk 2]
```diff
@@ -184,10 +189,6 @@ def run(self):
                         self.worker_manager.update_status(
                             WorkerManagerStatus.FINISHED
                         )
-                    if self.checkpoint_output_path:
```
terrytangyuan, line 187: So we save the saved model but nothing else, e.g. checkpoint?

elasticdl/python/master/task_dispatcher.py
```diff
@@ -10,12 +10,15 @@
...
-    def __init__(self, *, shard_name, start, end, type, model_version=-1):
+    def __init__(
+        self, shard_name, start, end, type, model_version=-1, **kwargs
+    ):
         self.shard_name = shard_name
         self.start = start
         self.end = end
         self.type = type
         self.model_version = model_version
+        self.extended_config = kwargs
```
terrytangyuan, line 21: Maybe `"extra_config"` is better?

elasticdl/proto/elasticdl.proto
```diff
@@ -41,6 +47,9 @@ message Task {
 
     // Whether this is training or evaluation task.
     TaskType type = 7;
+
+    // Extended task configuration as a key value list
```
terrytangyuan: "a key value list" -> "a list of key-value pairs"

elasticdl/python/worker/worker.py
```diff
@@ -513,9 +518,17 @@ def _process_save_model_task_if_needed(self):
...
+        if (task is not None) and (dataset is not None):
             # TODO: Implement the save model execution process
-            return
+            config_dict = {
+                pair.key: pair.value for pair in task.extended_config
+            }
+            saved_model_path = config_dict.get(
+                SaveModelConfig.SAVED_MODEL_PATH
+            )
+            logger.info("The saved model path is {}".format(saved_model_path))
```
terrytangyuan: -> The path to saved model is

elasticdl/python/worker/worker.py
```diff
@@ -513,9 +518,17 @@ def _process_save_model_task_if_needed(self):
         task, dataset = (
             self._task_data_service.get_save_model_task_and_dataset()
         )
-        if task and dataset:
+        if (task is not None) and (dataset is not None):
```
terrytangyuan: The parenthesis are not needed

elasticdl/python/worker/worker.py
```diff
@@ -513,9 +518,17 @@ def _process_save_model_task_if_needed(self):
...
-            return
+            config_dict = {
+                pair.key: pair.value for pair in task.extended_config
+            }
+            saved_model_path = config_dict.get(
+                SaveModelConfig.SAVED_MODEL_PATH
+            )
+            logger.info("The saved model path is {}".format(saved_model_path))
+
+            self.report_task_result(task_id=task.task_id, err_msg="")
```
terrytangyuan, line 527: Is `err_msg=""` intentional? If so, maybe add a comment here explaining

elasticdl/proto/elasticdl.proto
```diff
@@ -41,6 +47,9 @@ message Task {
...
+    repeated Pair extended_config = 8;
```
mhaoli: Add comment here about the info that are stored in `extended_config`?

elasticdl/python/master/master.py
```diff
@@ -68,8 +68,9 @@ def _maybe_create_shards(data_origin):
 class Master(object):
     def __init__(self, args):
         self.logger = get_logger("master", level=args.log_level.upper())
-        self.checkpoint_output_path = args.output
+
         self.num_ps_pods = args.num_ps_pods
+        self.checkpoint_output_path = args.checkpoint_dir
```
mhaoli, line 73: It seems a little confused that we save model in two formats, one is `checkpoint`, another is `tf.saved_model`. Would it be convenient if we change the save checkpoint file in the same format as `tf.saved_model`?

elasticdl/python/worker/worker.py
mhaoli: `if task and not dataset` is more concise.

elasticdl/python/master/master.py
brightcoder01: The checkpoint is saved regularly. At the last step, we export to the SavedModel as the final result.

elasticdl/proto/elasticdl.proto
brightcoder01: Good suggestion, updated.

elasticdl/python/worker/worker.py
brightcoder01: Good suggestion, update it to "The path to export model"

elasticdl/python/master/master.py
brightcoder01: Checkpoint is necessary during training, but it only stores the parameter value including optimizer slot value, it's used to restore the training process if preempted.

SavedModel contains both variable and model graph definition. It's used for serving and doesn't contain the optimizer state. What's more, it's a heavier operation.

mhaoli: Saving checkpoint now ignores many optimizer states. 

Yes, I known there are differences between checkpoint and SavedModel. But it seems no need to use two logic and format to save them. If we can save them in the same format, it would reuse many code and make our code base more clean. For example, save checkpoint in the same logic and format as used for savedModel, but only save variables and do not save graph definition. Any way, you can finish the project of serving first and then consider cleaning code.

elasticdl/python/worker/worker.py
brightcoder01: For SaveModel task, the dataset is necessary to build the model, so we need both task and dataset.

elasticdl/python/worker/worker.py
brightcoder01: In current implementation, err_msg = "" means the task is completed.

elasticdl/python/worker/worker.py
brightcoder01: Good idea, updated :)

elasticdl/python/master/task_dispatcher.py
brightcoder01: I think 'extra_config' and 'extended_config' are both fine. I prefer to keep the current name. :)

elasticdl/python/master/master.py
brightcoder01: Good idea, I prefer renaming it in the next PR. The 'output' command name already exist in the document, we need update it in all the documents. So that we can focus on the function implementation in this PR.

elasticdl/proto/elasticdl.proto
brightcoder01: Good suggestion, added.

workingloong (approved):

PR #1400, opened 2019-10-31
Restore keras model and parameters to export.

### LINKED ISSUE #1357: [Serving] Export the model with elasticdl.layers.Embedding to SavedModel
1. Restore the Keras model def and params by trained params of the elasticdl.layers.Embedding layer.
2. Add inputs and outputs to the model and save the model using SavedModel.

workingloong:
Resolve #1357
Implement the `get_model_to_export` in ModelHandler.


elasticdl/python/common/model_handler.py
```diff
@@ -61,57 +66,93 @@ def get_model_to_train(self, model):
...
         return model
 
-    def get_model_to_export(self, model, dataset):
+    def get_model_to_export(self, model, trained_params, dataset):
         """
-        To export model for tf-serving by tf.saved_model.save:
-        1. Add inputs and outputs to the model.
-        2. Restore Keras model and replace embedding parameters with trained
-        model.
+        Get the model which can be exported a SavedModel by
```
terrytangyuan: exported a -> exported to a

elasticdl/python/common/model_handler.py
```diff
@@ -61,57 +66,93 @@ def get_model_to_train(self, model):
...
+        tf.saved_model.save.
```
terrytangyuan: Put tilts around code (same in other places)

elasticdl/python/common/model_handler.py
```diff
@@ -61,57 +66,93 @@ def get_model_to_train(self, model):
...
+        for var in model.trainable_variables:
+            var.assign(trained_params[var.name])
+        return model
 
-    def _clone_model_for_sequential_and_functional(self, model):
+    def _restore_keras_model_def(self, model):
         """
-        Clone a new model with elasticdl.layers.Embedding for
-        Sequential and functional API model.
+        Restore Keras model definition by replacing elasticdl.layers.Embedding
```
terrytangyuan: Put tilts around code

elasticdl/python/common/model_handler.py
```diff
@@ -61,57 +66,93 @@ def get_model_to_train(self, model):
...
-        2. Restore Keras model and replace embedding parameters with trained
-        model.
+        Get the model which can be exported a SavedModel by
+        tf.saved_model.save.
         """
-        # TODO
-        pass
+        model = self._restore_keras_model_def(model)
+        if not model.inputs:
+            # build model to add inputs and outputs for tf-serving
```
terrytangyuan: for -> that can be consumed by

elasticdl/python/common/model_handler.py
```diff
@@ -61,57 +66,93 @@ def get_model_to_train(self, model):
...
+        Clone a new model by cloning model and replace the
+        src_embedding_class layer with a dst_embedding_class.
         """
 
         def _clone_function(layer):
-            if type(layer) == tf.keras.layers.Embedding:
-                logger.info("Replace Keras Embedding with ElasticDL Embedding")
-                edl_embedding_layer = Embedding(
+            if type(layer) == src_embedding_class:
+                logger.info(
```
terrytangyuan: maybe debug level log?

elasticdl/python/common/model_handler.py
```diff
@@ -61,57 +66,93 @@ def get_model_to_train(self, model):
...
-        for attr_name, attr_value in model.__dict__.items():
-            if type(attr_value) == tf.keras.layers.Embedding:
-                edl_embedding_layer = Embedding(
-                    output_dim=attr_value.output_dim,
-                    input_dim=attr_value.input_dim,
-                    embedding_initializer=attr_value.embeddings_initializer,
-                    mask_zero=attr_value.mask_zero,
-                    input_length=attr_value.input_length,
+        for name, value in model.__dict__.items():
+            if type(value) == src_embedding_class:
```
terrytangyuan, line 150: use `isinstance` instead of string comparison?

elasticdl/python/common/model_handler.py
```diff
@@ -61,57 +66,93 @@ def get_model_to_train(self, model):
...
+                embedding_layer = dst_embedding_class(
+                    output_dim=value.output_dim,
+                    input_dim=value.input_dim,
+                    embeddings_initializer=value.embeddings_initializer,
+                    mask_zero=value.mask_zero,
+                    input_length=value.input_length,
                 )
-                setattr(model, attr_name, edl_embedding_layer)
+                del value
```
terrytangyuan: Why this has to be deleted?

elasticdl/python/common/model_handler.py
workingloong: `isinstance(value, class)` returns True if the value is a subclass instance of the class. Here, we only want to replace `tf.keras.layers.Embedding` and don't replace any custom layer extended `tf.keras.layers.Embedding`. So, I use `type` and not `isinstance`.

elasticdl/python/common/model_handler.py
workingloong: I want to delete it to free the memory if the embedding layer instance may have been initialized. However, the embedding layer in a subclass model is initialized until the `call` of the model is called. So, we don't need to delete it and I have removed the line in the last commit.

elasticdl/python/common/model_handler.py
workingloong: Use `tf.saved_model.save`  in docstring?

elasticdl/python/common/model_handler.py
workingloong: Yes, users don't need to care about the log.

terrytangyuan (approved):

PR #1402, opened 2019-10-31
Design overview for ElasticDL on SQLFlow

terrytangyuan:
This is a brief summary of the current design, implementation, and outline of next steps.

cc @tonyyang-svail 

### DIFF
--- a/docs/designs/sqlflow_integration_design.md
+++ b/docs/designs/sqlflow_integration_design.md
@@ -1,0 +1,237 @@
+# Design Overview for ElasticDL on SQLFlow
+
+## Overview
+
+This is a design doc on integration with [SQLFlow](https://github.com/sql-machine-learning/sqlflow).
+
+### User Interface
+
+#### Training Job Submission
+
+```sql
+SELECT
+    c1, c2, c3, c4, c5 as class
+FROM training_data
+TRAIN ElasticDLKerasClassifier
+WITH
+    model.optimizer = "optimizer",
+    model.loss = "loss",
+    model.eval_metrics_fn = "eval_metrics_fn",
+    model.num_classes = 3,
+    model.dataset_fn = "dataset_fn",
+    train.shuffle = 120,
+    train.epoch = 2,
+    train.grads_to_wait = 2,
+    train.tensorboard_log_dir = "",
+    train.checkpoint_steps = 0,
+    train.checkpoint_dir = "",
+    train.keep_checkpoint_max = 0,
+    eval.steps = 0,
+    eval.start_delay_secs = 100,
+    eval.throttle_secs = 0,
+    eval.checkpoint_filename_for_init = "",
+    engine.master_resource_request = "cpu=400m,memory=1024Mi",
+    engine.master_resource_limit = "cpu=1,memory=2048Mi",
+    engine.worker_resource_request = "cpu=400m,memory=2048Mi",
+    engine.worker_resource_limit = "cpu=1,memory=3072Mi",
+    engine.num_workers = 2,
+    engine.volume = "",
+    engine.image_pull_policy = "Never",
+    engine.restart_policy = "Never",
+    engine.extra_pypi_index = "",
+    engine.namespace = "default",
+    engine.minibatch_size = 64,
+    engine.master_pod_priority = "",
+    engine.cluster_spec = "",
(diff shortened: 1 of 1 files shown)


docs/designs/sqlflow_integration_design.md
```diff
@@ -0,0 +1,237 @@
...
+--minibatch_size=64 \
+--num_minibatches_per_task=10 \
+--num_workers=2 \
+--checkpoint_steps=10 \
+--evaluation_steps=15 \
+--grads_to_wait=2 \
+--job_name=test-iris \
+--log_level=INFO \
+--image_pull_policy=Never \
+--output=<model-output> \
```
tonyyang-svail, line 200: Looks like we save the trained model at `--output`. Is `--output` determined by the `INTO` clause in the SQL? If so, please be aware that `TrainIR` doesn't contain the `INTO` clause due to [this reason](https://github.com/sql-machine-learning/sqlflow/blob/1fc8e5d31261c16d74416922dac5689c42f83f9f/pkg/sql/codegen/intermediate_representation.go#L64-L66).

docs/designs/sqlflow_integration_design.md
```diff
@@ -0,0 +1,237 @@
...
+LABEL class
+INTO trained_elasticdl_keras_classifier;
+```
+
+#### Prediction Job Submission
+
+```sql
+SELECT
+    c1, c2, c3, c4
+FROM prediction_data
```
tonyyang-svail, line 63: How to make sure the prediction has the same input name as the training?

docs/designs/sqlflow_integration_design.md
```diff
@@ -0,0 +1,237 @@
...
+                    record, label_col_ind, features_shape
+                ),
+                labels,
+            )
+        {{end}}
+        return tf.reshape(record, features_shape)
+
+    dataset = dataset.map(_parse_data)
+
+    {{if .IsTraining}}
```
wangkuiyi, line 156: Let us remove Go template parameters from the Python code. Instead, let us try passing in `{{.IsTraining}}` as a Python function parameter of `dataset_fn` or `_parse_data`. In this way, most part of the submitter template would be runnable Python code, which could have corresponding unit tests.

Ultimately, the template should be extremely simple -- a single funtion call, which parameters are template paramters. This function call in turn calls other functions, like `dataset_fn`.

wangkuiyi (approved):

docs/designs/sqlflow_integration_design.md
terrytangyuan: "The sql package will save the output files of a generated Python program." -> how are these files saved? So where else would I obtain this `INTO` clause if it's not in `TrainIR`?

docs/designs/sqlflow_integration_design.md
terrytangyuan: Yes, I agree that we should make the generated code as concise as possible and it would be a good next step. This doc is a summary of the current design, implementation, and brief outline of next steps.

docs/designs/sqlflow_integration_design.md
skydoorkai: > How to make sure the prediction has the same input name as the training?

Good question. Seems that SQLFlow needs to know the model input names so that it can generate `dataset_fn` with correct input names.

terrytangyuan: @tonyyang-svail How does SQLFlow currently handle this? Are feature names saved somewhere as part of training?

docs/designs/sqlflow_integration_design.md
tonyyang-svail: > how are these files saved?

In a training job, SQLFlow runs the EDL submitter program in a temporary directory. When training is finished, SQLFlow serialized the directory via zip, and dump the zip file to an ODPS table specified by the `INTO` clause.

> So where else would I obtain this `INTO` clause if it't not in `TrainIR`?

So, I believe EDL shouldn't know `INTO` clause.

docs/designs/sqlflow_integration_design.md
tonyyang-svail: At the end of the training, SQLFlow saves the train SQL along with the model files. In a prediction job, the train SQL will be used to determine the input schema.

PR #1406, opened 2019-11-04
Async push gradients

### LINKED ISSUE #1367: [PS-1] RPC service `push_gradient` async-SGD implementation
Async apply gradients to non-embedding parameters. Updating embedding parameters will be supported in issue #1306 .

mhaoli:
Fixes #1367.

### DIFF
--- a/elasticdl/python/tests/pserver_servicer_test.py
+++ b/elasticdl/python/tests/pserver_servicer_test.py
@@ -3,6 +3,7 @@ import unittest
 
 import grpc
 import numpy as np
+import tensorflow as tf
 from google.protobuf import empty_pb2
 
 from elasticdl.proto import elasticdl_pb2, elasticdl_pb2_grpc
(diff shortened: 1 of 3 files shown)


elasticdl/python/tests/pserver_servicer_test.py
```diff
@@ -78,6 +79,8 @@ def create_server_and_stub(
         self._server = pserver.server
         self._stub = elasticdl_pb2_grpc.PserverStub(self._channel)
 
+        self._lr = 0.1
```
QiJune, line 82: We set self._lr = 0.1 directly here. Actually, the value is set at https://github.com/sql-machine-learning/elasticdl/blob/develop/elasticdl/python/tests/test_module.py#L60

This is a little confusing. We could refine it later.

QiJune (approved):
Excellent!

elasticdl/python/tests/pserver_servicer_test.py
mhaoli: Yes, but we can't get servicer in pserver_servicer_test. Hence, if we want to get *lr*, we need to get model module, and code will be a little tedious. So I set `lr=0.1` directly. 

May set servicer to `parameter_server`?

QiJune: I think it's Okay to add `servicer` as a member of `ParameterServer`.

PR #1409, opened 2019-11-04
add sync update logic in push_gradient of pserver servicer

### LINKED ISSUE #1368: [PS-1] RPC service `push_gradient` sync-SGD implementation

QiJune:
Fix #1368


elasticdl/python/ps/parameters.py
```diff
@@ -40,6 +40,26 @@ def set_embedding_param(self, name, indices, values):
...
+        name = grad.name
+        if name in self.non_embedding_params:
+            param_shape = tuple(
+                self.non_embedding_params[name].get_shape().as_list()
+            )
+            if grad.is_indexed_slices():
+                dim0 = tf.math.reduce_max(grad.indices).numpy()
+                dim1 = grad.values.shape[1]
+                if dim0 > param_shape[0] or dim1 != param_shape[1]:
+                    raise ValueError("Gradient %s shape is incompatible", name)
```
terrytangyuan: Shape of the gradient %s is incompatible

terrytangyuan: It would also be better if we could print out what the shapes are that are different

elasticdl/python/tests/parameters_test.py
```diff
@@ -106,3 +106,23 @@ def test_set_embedding_param(self):
 
         with self.assertRaises(ValueError):
             self.params.set_embedding_param("tom", [0, 1, 2], values)
+
+    def test_check_grad(self):
+        self._clear_params()
+        self.params.init_from_model_pb(self.model_pb)
+
+        grad0 = Tensor(name="z")
+        with self.assertRaises(ValueError):
```
terrytangyuan: For all the `assertRaises` checks, it would be better to check a regex of the error message so we know which exception we are testing against

elasticdl/python/ps/servicer.py
```diff
@@ -70,29 +73,62 @@ def push_model(self, request, _):
...
+        res = elasticdl_pb2.PushGradientResponse()
         if self._use_async:
             grad_vars = []
             for pb in request.gradients:
-                tensor = Tensor.from_tensor_pb(pb)
-                var = self._parameters.get_non_embedding_param(tensor.name)
+                grad = Tensor.from_tensor_pb(pb)
+                self._parameters.check_grad(grad)
+                var = self._parameters.get_non_embedding_param(grad.name)
                 if var is None:
```
mhaoli: Do not need to check it because this has been checked in `parameters.check_grad`.

elasticdl/python/ps/servicer.py
```diff
@@ -70,29 +73,62 @@ def push_model(self, request, _):
...
     def push_gradient(self, request, _):
+        res = elasticdl_pb2.PushGradientResponse()
         if self._use_async:
             grad_vars = []
             for pb in request.gradients:
-                tensor = Tensor.from_tensor_pb(pb)
-                var = self._parameters.get_non_embedding_param(tensor.name)
+                grad = Tensor.from_tensor_pb(pb)
+                self._parameters.check_grad(grad)
+                var = self._parameters.get_non_embedding_param(grad.name)
```
mhaoli, line 81: `check_grad` queries the dictionary `non_embedding_param`, and `get_non_embedding_param` also queries it. Can we rename `check_grad` to `check_grad_and_get_var`, just return the `var` in this function?

elasticdl/python/ps/parameters.py
QiJune: Done

QiJune: Done

elasticdl/python/tests/parameters_test.py
QiJune: Done

elasticdl/python/ps/servicer.py
QiJune: Done

elasticdl/python/ps/servicer.py
QiJune: There is also embedding_param. We could decide to rename it or not after the embedding param feature is supported.

mhaoli: Agree

mhaoli (approved):
Excellent!

PR #1413, opened 2019-11-05
Rename variables in k8s_worker_manager

ywskycn:
To be prepared for adding ps_pod management worker there.


elasticdl/python/master/k8s_worker_manager.py
```diff
@@ -10,26 +10,26 @@ class WorkerManager(object):
...
-        command,
-        args,
         num_workers=1,
+        worker_command=None,
+        worker_args=None,
         worker_resource_request="cpu=1,memory=4096Mi",
         worker_resource_limit="cpu=1,memory=4096Mi",
-        pod_priority=None,
+        worker_pod_priority=None,
         volume=None,
```
terrytangyuan, line 19: The rest of the args seem to be specific to worker pod as well. Should we rename them as well? Or should we remove all "worker_" prefixes and assume all these args apply to worker pods? Docstring would help clarify this too.

ywskycn: For `volume`, image_pull_policy`, `restart_policy`, `envs`, current workers and ps pods can share, thus I keep them there.

terrytangyuan (approved):

PR #1414, opened 2019-11-05
partition gradients in worker

### LINKED ISSUE #1378: Worker partitions gradients and `push_gradient` to corresponding PS pods

QiJune:
Fix #1378 

I am thinking about how to add unit tests.


elasticdl/python/worker/worker.py
```diff
@@ -198,6 +209,22 @@ def get_model(self, version, method):
...
 
+    def get_model_from_ps(self, version, method):
+        for stub in self._ps_stubs:
+            req = empty_pb2.Empty()
+            res = stub.pull_variable(req)
+
+            for tensor_pb in res.model.param:
+                tensor = Tensor.from_tensor_pb(tensor_pb)
+                self._non_embed_vars[tensor.name].assign(tensor.to_ndarray())
+            self._model_version = res.model.version
```
skydoorkai: Each ps pod may have a different version. We can assign self._model_version as max(versions from ps pods).

elasticdl/python/worker/worker.py
```diff
@@ -207,21 +234,45 @@ def report_task_result(self, task_id, err_msg):
...
             emplace_tensor_pb_from_ndarray(
                 req.variable, v.numpy(), name=v.name
             )
         self._stub.ReportVariable(req)
 
-    def report_gradient(self, grads):
-        """
-        report gradient to ps, return (accepted, model_version) from rpc call.
-        """
+    def _hash_var_to_ps(self, name):
```
skydoorkai: Move hash/partition related methods to common directory, as ps checkpoint loading may need these methods.

elasticdl/python/worker/worker.py
```diff
@@ -207,21 +234,45 @@ def report_task_result(self, task_id, err_msg):
...
+        h = hashlib.sha256(name.encode("utf-8"))
+        return int(h.hexdigest(), base=32) % len(self._ps_stubs)
+
+    def report_variable_to_ps(self):
+        ps_vars = {}
```
mhaoli, line 249: Just save ps_vars to `self.ps_vars` and no need to calculate it every time.

mhaoli (commented):
It seems that there are no unit tests for worker's gRPC call?

QiJune:
> It seems that there is no unit tests for worker's gRPC call?

@LiMinghao1994 Yes, we lack unit tests for the worker module. I am thinking about how to add it.

elasticdl/python/worker/worker.py
QiJune: Done

elasticdl/python/worker/worker.py
QiJune: Since `report_variable_to_ps` will be called only when PS is not initialized, it's okay to use a temporary dictionary here.

elasticdl/python/worker/worker.py
QiJune: Done

elasticdl/python/worker/worker.py
```diff
@@ -198,6 +209,25 @@ def get_model(self, version, method):
             self._non_embed_vars[tensor.name].assign(tensor.to_ndarray())
         self._model_version = model.version
 
+    def get_model_from_ps(self, version, method):
```
workingloong, line 212: Add a docstring.

elasticdl/python/worker/worker.py
```diff
@@ -198,6 +209,25 @@ def get_model(self, version, method):
...
+                self._non_embed_vars[tensor.name].assign(tensor.to_ndarray())
+
+            model_version = max(model_version, res.model.version)
+        self._model_version = model_version
+
+    def get_model(self, version, method):
+        if self._use_multi_ps:
+            self.get_model_from_ps(version, method)
+        else:
+            self.get_model_from_master(version, method)
```
workingloong, line 229: Maybe `get_non_embed_vars_from_ps` and `get_non_embed_vars_from_master` ? And the tensor assignment can be replicated in `get_model`.
```python
def get_model(self, version, metho):
     ...
     for tensor_pb in params:
         tensor = Tensor.from_tensor_pb(tensor_pb)
         self._non_embed_vars[tensor.name].assign(tensor.to_ndarray())
```
I also wonder whether we need to abstract clients to communicate with the Master or the PS. For example, the worker gets tasks by the master client and gets model pararms by PS client. Now, the communication operators are all in worker.py which may be not very clean.

elasticdl/python/worker/worker.py
QiJune: This is only a temporary method. After the development of the new PS is ready, we will move `master-worker` architect to `master-pserver-worker` architect. And unused method will be cleaned. We will add detailed docstring later.

elasticdl/python/worker/worker.py
QiJune: The same reason, `get_model_from_master` will be removed later. We keep compatibility with master-worker architect now.  So, I just keep the master related code unchanged.

For the communication client abstraction, I think it does make sense. A worker has two clients:
- pull parameters/push gradients communication with PS
- get task/report metrics communication with master

We could consider cleaning worker code in next PRs.

elasticdl/python/tests/worker_mnist_test.py
```diff
@@ -0,0 +1,147 @@
...
+class WorkerMNISTTest(unittest.TestCase):
+    def setUp(self):
+        ports = [12345, 12346]
+        self._pserver, self._channel = self._create_pserver_and_channel(ports)
+
+    def tearDown(self):
+        for pserver in self._pserver:
+            pserver.server.stop(0)
+
+    def _create_pserver_and_channel(self, ports):
```
mhaoli, line 64: More than one test files have `_create_pserver_and_channel`. Better to remove it as a function in `test_utils.py`. (This can be done in another PR.)

mhaoli (approved):

Issue #1306, opened 2019-10-14
Implement optimizer wrapper in PS module


mhaoli:
This can be done in two steps:

* [x] add slot parameters in `PS.Paramters`, each slot variable can be represented as a embedding table.
* [x] use optimizer wrapper to do parameters updating.

PR #1418, opened 2019-11-06
Add slot table in PS.Parameters

mhaoli:
Part of #1306 , step 1.

### DIFF
--- a/elasticdl/python/ps/embedding_table.py
+++ b/elasticdl/python/ps/embedding_table.py
@@ -16,6 +16,16 @@ class EmbeddingTable(object):
     """
 
     def __init__(self, name, dim=None, initializer=None, is_slot=False):
+        """
+        Args:
+            name: The embedding table name.
+            dim: The dimension of embeddings in this embedding table.
+            initializer: The initializer to initialize new embeddings. If this
+                embedding table is for slots, `initializer` is a float and this
+                table will initialize with constant initializer. Otherwise
+                `initializer` is the name of Keras initializer.
+            is_slot: A bool. True for storing slot variable, otherwise false.
+        """
         self.name = name
         self.dim = dim
         if is_slot:
--- a/elasticdl/python/ps/embedding_table.py
+++ b/elasticdl/python/ps/embedding_table.py
@@ -15,10 +15,15 @@ class EmbeddingTable(object):
     dictionary, the corresponding value will be initialized.
     """
 
-    def __init__(self, name, dim=None, initializer=None):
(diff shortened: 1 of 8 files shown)


elasticdl/python/ps/embedding_table.py
```diff
@@ -15,10 +15,15 @@ class EmbeddingTable(object):
     dictionary, the corresponding value will be initialized.
     """
 
-    def __init__(self, name, dim=None, initializer=None):
+    def __init__(self, name, dim=None, initializer=None, is_slot=False):
```
terrytangyuan, line 18: Add some docs for each arg

elasticdl/python/ps/parameters.py
```diff
@@ -99,6 +103,8 @@ def init_from_model_pb(self, model_pb):
             self._init_embedding_params(embeddings_pb)
             self.version = model_pb.version
             self.init_status = True
+            return True
```
terrytangyuan, line 117: Combine the two return statements to one return `self.init_status` instead?

mhaoli: > Combine the two return statements to one return `self.init_status` instead?

The return value does not equal to `init_status`, it means whether `init_from_model_pb` accepts this model pb. I will add a docs here.

elasticdl/python/ps/parameters.py
```diff
@@ -110,3 +116,22 @@ def _init_non_embedding_params(self, tensors_pb):
...
         for pb in embeddings_pb:
             self.embedding_params[pb.name] = create_embedding_table(pb)
+
+    def has_embedding_params(self):
+        return len(self.embedding_params) > 0
+
+    def create_slot_params(self, slot_names, init_values):
+        embed_layer_names = list(self.embedding_params.keys())
+        for layer_name in embed_layer_names:
+            for slot_name in slot_names:
```
QiJune, line 137: One question, are the slots of all the embedding layers the same?

elasticdl/python/ps/embedding_table.py
mhaoli: Done

elasticdl/python/ps/parameters.py
mhaoli: Yes, it is decided by the optimizer type. Adam creates slots "m" and "v" for each variable, and  Adagrad creates slots "accumulator" for each variable. The user-defined arguments of initial values are also in optimizer.

QiJune (approved):
Excellent!

PR #1420, opened 2019-11-06
Add design doc for allreduce-based distributed training

terrytangyuan:


terrytangyuan:
This should be ready for review now. Also cc'ing Caicloud folks for feedback @zw0610 @pendoragon @gaocegege.

docs/designs/allreduce.md
```diff
@@ -1,14 +1,198 @@
...
+This document describes the design for supporting allreduce-based distributed training in ElasticDL.
+
+## Motivation
+
+While distributed training based on parameter servers can support training very large models and datasets by adding
+more workers and parameter servers as described in [parameter server design doc](parameter_server.md), there are additional
+challenges involved in order to optimize the performance, for example:
+
+* It is not easy to identify the right ratio of the number of workers to the number of parameter servers. For example,
+if only a small number of parameter servers are used, there will likely be network communication overheads. If many parameter
```
ywskycn: `if only a small number of parameter servers are used, network communication will likely become training bottleneck.`

docs/designs/allreduce.md
```diff
@@ -1,14 +1,198 @@
...
+to use in certain use cases. There are many existing technologies available that provide implementations for these collective
+communication primitives and please head over to the section on [existing collective communication technologies](#existing-collective-communication-technologies)
+for details if interested. Allreduce-based distributed training could address many of the challenges mentioned above, for example:
+
+* Each worker stores a complete set of model parameters. In other words, no parameter server is needed so it's straightforward
+to add more workers when necessary.
+* Failures among the workers can be recovered easily by restarting the failed workers and then load the current model sent from
+any of the existing workers. Model does not need to be replicated to support fault-tolerance, which avoids the waste of resources.
+* The model can be updated more efficiently by fully leveraging the network structure and collective communication algorithms.
+For example, in ring-allreduce algorithm, each of *N* workers only needs to communicate with two of its peer workers *2 * (N − 1)* times.
```
ywskycn: `2*(N-1) times to get all parameter updates`?

docs/designs/allreduce.md
```diff
@@ -1,14 +1,198 @@
...
+
+If any of the gradients averaging operation fails while the workers are still active and healthy, we simply report the
+failure and continue training on the next batch.
+
+If any of the workers fails, e.g. the pod is accidentally killed, the following steps will be performed:
+
+1. The task that the failed worker was handling will be added back to the task queue.
+1. Since the `AllreduceCommunicator` is aware of the failure. It will reconstruct the communicator and re-assign ranks among
+the existing active workers. The existing workers will then continue to run on the tasks at hand.
+1. The master pod will try to create a new worker pod after a specified restart delay period.
```
ywskycn, line 170: 👍  `restart delay`

docs/designs/allreduce.md
```diff
@@ -1,14 +1,198 @@
...
+    1. Initialize the model on the new worker pod using the current model that master pod just received.
+    1. All workers continue to operate and perform allreduce operations to update the model.
+
+Note that since the existing active workers have the exact same copy of the model after the previous allreduce operation completes,
+we can guarantee that the new worker will have the same copy of the model as the ones on other workers once the next
+allreduce operation completes.
+
+#### Training with Evaluation
+
+If the worker encounters any evaluation tasks in the above process, it will evaluate the model directly on master once the
```
ywskycn: Need to revise this sentence... `it will evaluate the model directly on master` --> looks like the master will evaluate the model, but actually workers use the model received from the master..

docs/designs/allreduce.md
```diff
@@ -1,14 +1,198 @@
...
+failed worker pod which requires reconstruction of the communicator again once the pod becomes active. 
+* ``--distribution_strategy``: In addition to the existing "ParameterServerStrategy" that we have, we add a new strategy
+called "AllreduceStrategy".
+* ``--num_ps_pods`` will be ignored if "AllreduceStrategy" is used and only ``--num_workers`` will be taken into account.
+* ``--use_async`` and ``--lr_staleness_modulation`` will be ignored if "AllreduceStrategy" is used.
+* Only ``--grads_to_wait = 1`` will be supported if "AllreduceStrategy" is used.
+
+The following new CLI arguments are added:
+
+* ``--restart_delay_secs``: The number of seconds to delay before restarting the failed pods. This could be useful when
```
ywskycn: Better to use `--relaunch_delay_secs`?

docs/designs/allreduce.md
```diff
@@ -1,14 +1,198 @@
...
+num_batches = 10
+data_loader = DataLoader(num_batches)
+communicator = AllreduceCommunicator()
+
+for _ in range(num_epochs):
+    for features, labels in data_loader:
+        outputs = model(features)
+        loss = loss_fn(outputs, labels)
+        grads = calculate_gradients(loss, model)
+        res, averaged_grads = communicator.average_gradients(grads)
```
zw0610: While we are conducting test in this style, compared with horovod, using `average gradients` explicitly may ask users to have an entry-level understanding of distributed training.

So it could remain as a potential improvement in the future whether we want to adopt the horovod style, where `average gradients` API is merged into optimizers.

terrytangyuan: Yes I agree that higher level APIs like Horovod’s DistributedOptimizer might be great addition. These low-level APIs are necessary for integration with ElasticDL though. Users of ElasticDL will not need to understand these since this kind of details are hidden from users.

docs/designs/allreduce.md
```diff
@@ -1,14 +1,198 @@
...
+
+* It is not easy to identify the right ratio of the number of workers to the number of parameter servers. For example,
+if only a small number of parameter servers are used, there will likely be network communication overheads. If many parameter
+servers are used, the communication may saturate network interconnects.
+* If the model could fit within the computational resources on each worker, additional maintenance and communication overheads are
+introduced when the model is partitioned to multiple parameter servers.
+* Additional computational resources for replicating models of each parameter server are needed to support fault-tolerance.
+
+In contrast, distributed training based on collective communication primitives such as [allreduce](https://mpitutorial.com/tutorials/mpi-reduce-and-allreduce/)
+and [ring-allreduce](http://research.baidu.com/bringing-hpc-techniques-deep-learning/) could be more efficient and easier
```
skydoorkai: `ring-allreduce` is an implementation of the collective communication primitive `allreduce`.

skydoorkai (commented):
Also 2 questions:
1. For allreduce, every worker must have training data in order to generate gradients. In ElasticDL's task-based system, how to distribute the training data evenly among workers so that every worker will generate gradients when the training is not finished.
2. For elastic, the number of workers may vary in the training. This will affect the batch size (batch_size = sum(minibatch on each worker)). The model training accuracy is affected by the batch size. How to deal with this problem?

docs/designs/allreduce.md
```diff
@@ -1,14 +1,198 @@
...
+
+We are collaborating with [Caicloud](https://github.com/caicloud/) on building an API that provides implementations of
+fault-tolerant allreduce. The initial implementation will contain an experimental Python binding for NCCL that is
+fault-tolerant and Kubernetes-native. This will include but not limited to the following objectives (more details to be disclosed later once
+the implementation has been open-sourced):
+
+* Fault-tolerant: if any of the worker pod fails, the [NCCL Communicator](https://docs.nvidia.com/deeplearning/sdk/nccl-developer-guide/docs/usage/communicators.html)
+can be reconstructed. The allreduce operation continues as long as there's at least one healthy worker pod.
+* Elastic: the number of worker pods can be dynamically added if there are enough computational resources available.
+Ranks can be re-assigned as the number of worker pods changes.
```
QiJune: Do we need to reconstruct a new NCCL Communicator here?

docs/designs/allreduce.md
```diff
@@ -1,14 +1,198 @@
...
+
+### Allreduce-based Training in ElasticDL
+
+#### Gradients Averaging and Model Updating
+
+For training based on parameter servers, gradients calculation and model updating include the following steps:
+
+1. Send the gradients calculated locally from each worker to master.
+1. Calculate the average of all the received gradients on master.
+1. Update the model in master.
```
QiJune: Since we have implemented new PS, the gradients are pushed to PS.

docs/designs/allreduce.md
```diff
@@ -1,14 +1,198 @@
...
+Ranks can be re-assigned as the number of worker pods changes.
+
+The interface would look like the following:
+
+```python
+num_epochs = 2
+num_batches = 10
+data_loader = DataLoader(num_batches)
+communicator = AllreduceCommunicator()
+
```
QiJune, line 69: Do we need a BroadcastCommunicator here? We choose one worker to initialize the parameters first, and then it broadcasts to other workers. This will make parameters on each worker the same.

```python
b_communicator = BroadcastCommunicator()
init_model(worker_0)
b_communicator.broadcast(all_workers, worker_0.params)
```

docs/designs/allreduce.md
```diff
@@ -1,14 +1,198 @@
...
+1. The task that the failed worker was handling will be added back to the task queue.
+1. Since the `AllreduceCommunicator` is aware of the failure. It will reconstruct the communicator and re-assign ranks among
+the existing active workers. The existing workers will then continue to run on the tasks at hand.
+1. The master pod will try to create a new worker pod after a specified restart delay period.
+1. Once the worker pod becomes active and `AllreduceCommunicator` is aware of it, we perform the following steps:
+    1. Lock all existing worker pods.
+    1. Send the current model from one of the worker pods to master pod.
+    1. Initialize the model on the new worker pod using the current model that master pod just received.
+    1. All workers continue to operate and perform allreduce operations to update the model.
```
QiJune: This is actually a broadcast operation. If a new worker is added, we broadcast the parameters from one worker which has parameters. The logic here is the same with initialization before the training starts.

We could make the logic of workers initialization and workers joining the same.

docs/designs/allreduce.md
QiJune: Under allreduce strategy, the models of each worker are the same. I am not sure why we need to store the model on master. Or another solution, we choose worker 0 to be the master. The priority of worker 0 pod is high than other worker pods.

docs/designs/allreduce.md
```diff
@@ -1,14 +1,198 @@
...
+under-going allreduce-based gradients averaging has completed and the model has been updated. The behavior is the same as
+what's described in [model evaluation design doc](model_evaluation.md) except that we are evaluating the model on workers
+instead of on parameter servers. Once an evaluation completes, we send the evaluation result to master for TensorBoard
+service to consume.
+
+#### ElasticDL Embedding Layer
+
+ElasticDL embedding layer is not supported under allreduce-based training since each worker has the exact same copy of
+the model that must fit within each worker's specified resources. Any layers in the model defined via
+``elasticdl.layers.embedding`` will be replaced by `tf.keras.layers.Embedding`.
```
QiJune: Do we have the plan to support SparseAllreduce here? Since the gradients of the embedding layer are row-sparse, SparseAllreduce will decrease communication expenses.

docs/designs/allreduce.md
```diff
@@ -1,14 +1,198 @@
...
+* ``--restart_delay_secs``: The number of seconds to delay before restarting the failed pods. This could be useful when
+
+
+## Potential Future Optimizations
+
+* We can potentially overlap the backward computations and gradient optimizations. More discussions on this can be found
+in [this Github issue](https://github.com/tensorflow/tensorflow/issues/33274).
+* For models with a large amount of tensors, such as ResNet, many small allreduce operations are needed. In this case,
+we could fuse multiple small tensors together before performing allreduce operations to maximize performance since
+allreduce utilizes the network in an optimal way if the tensors are large enough.
```
QiJune, line 252: Another one: SparseAllreduce support for embedding layers.

zw0610:
> 2. For elastic, the number of workers may vary in the training. This will affect the batch size (batch_size = sum(minibatch on each worker)). The model training accuracy is affected by the batch size. How to deal with this problem?

I haven't discussed this question with ElasticDL, but did think about it before. One possible approach is to make the learning rate scheduler aware of the change of overall batch_size. So that the lr scheduler can tweak the learning rate to remedy the bumping bs. The algorithm such tweak action follows might remain as a task for researchers who know better with numeric stability.

skydoorkai:
> > 2. For elastic, the number of workers may vary in the training. This will affect the batch size (batch_size = sum(minibatch on each worker)). The model training accuracy is affected by the batch size. How to deal with this problem?
> 
> I haven't discussed this question with ElasticDL, but did think about it before. One possible approach is to make the learning rate scheduler aware of the change of overall batch_size. So that the lr scheduler can tweak the learning rate to remedy the bumping bs. The algorithm such tweak action follows might remain as a task for researchers who know better with numeric stability.

Yes, ElasticDL needs to support customized learning rate scheduler, which will take epoch/batch_size/metrics into account. 
Also, support for LARS (Layer-wise Adaptive Rate Scaling) so that we can use a large batch size.

docs/designs/allreduce.md
terrytangyuan: Added

docs/designs/allreduce.md
terrytangyuan: Yes. Added.

docs/designs/allreduce.md
terrytangyuan: I think `restart_delay_secs` might be better to be consistent with existing args like `restart_policy`. We can revisit the args together later when needed.

docs/designs/allreduce.md
terrytangyuan: Thanks. Edited.

docs/designs/allreduce.md
terrytangyuan: Thanks. Edited.

docs/designs/allreduce.md
terrytangyuan: Thanks. Fixed.

docs/designs/allreduce.md
terrytangyuan: `AllreduceCommunicator` includes both reduce and broadcast so we don't need to do this separately.

docs/designs/allreduce.md
terrytangyuan: Yes, updated the steps.

docs/designs/allreduce.md
terrytangyuan: Thanks for the reminder. Yes we do need to expose broadcast API. Just added.

docs/designs/allreduce.md
terrytangyuan: @ywskycn This is a typo. I changed it to "it will evaluate the model directly on the **worker**". 

@QiJune Since we currently rely on master to perform TensorBoard and start new workers, etc. We still need to send the evaluation results to master for TensorBoard visualization purposes. ElasticDL needs to support master election in the future so we can reuse one of the worker pods as master pod. I'll open a separate issue for that. See https://github.com/sql-machine-learning/elasticdl/issues/1440

terrytangyuan:
> 2. For elastic, the number of workers may vary in the training. This will affect the batch size (batch_size = sum(minibatch on each worker)). The model training accuracy is affected by the batch size. How to deal with this problem?

@zw0610 @skydoorkai Yes, I agree that this should be adjusted by researchers via solutions like customized learning rate scheduler and LARS. I've added this to "Potential Future Optimizations" section.

docs/designs/allreduce.md
```diff
@@ -1,14 +1,215 @@
...
+
+### Allreduce-based Training in ElasticDL
+
+#### Gradients Averaging and Model Updating
+
+For training based on parameter servers, gradients calculation and model updating include the following steps:
+
+1. Send the gradients calculated locally from each worker to parameter servers.
+1. Calculate the average of all the received gradients on master.
+1. Update the model in master.
```
QiJune: master --> parameter servers

docs/designs/allreduce.md
```diff
@@ -1,14 +1,215 @@
...
+failure and continue training on the next batch.
+
+If any of the workers fails, e.g. the pod is accidentally killed, the following steps will be performed:
+
+1. The task that the failed worker was handling will be added back to the task queue.
+1. Since the `AllreduceCommunicator` is aware of the failure. It will reconstruct the communicator and re-assign ranks among
+the existing active workers. The existing workers will then continue to run on the tasks at hand.
+1. The master pod will try to create a new worker pod after a specified restart delay period.
+1. Once the worker pod becomes active and `AllreduceCommunicator` is aware of it, we perform the following steps:
+    1. Lock all existing worker pods.
```
QiJune: A detailed question, how to lock all existing worker pods? It's actually a global barrier, do we resue some primitive like MPI barrier? Or the master sends a message to all workers?

docs/designs/allreduce.md
```diff
@@ -1,14 +1,215 @@
...
+
+In contrast, distributed training based on collective communication primitives such as [allreduce](https://mpitutorial.com/tutorials/mpi-reduce-and-allreduce/)
+could be more efficient and easier to use in certain use cases. There are many existing technologies available that provide
+implementations for these collective communication primitives and please head over to the section on
+[existing collective communication technologies](#existing-collective-communication-technologies)
+for details if interested. Allreduce-based distributed training could address many of the challenges mentioned above, for example:
+
+* Each worker stores a complete set of model parameters. In other words, no parameter server is needed so it's straightforward
+to add more workers when necessary.
+* Failures among the workers can be recovered easily by restarting the failed workers and then load the current model sent from
```
brightcoder01: IMO, the model are replicated in all the workers. Compared with parameter server, the resource is not saved.

docs/designs/allreduce.md
```diff
@@ -1,14 +1,215 @@
...
-This document describes the design for supporting Allreduce in ElasticDL. Note that this is still a work-in-progress.
+This document describes the design for supporting allreduce-based distributed training in ElasticDL.
+
+## Motivation
+
+While distributed training based on parameter servers can support training very large models and datasets by adding
+more workers and parameter servers as described in [parameter server design doc](parameter_server.md), there are additional
+challenges involved in order to optimize the performance, for example:
+
+* It is not easy to identify the right ratio of the number of workers to the number of parameter servers. For example,
```
brightcoder01, line 11: What's more, how to set the memory quota of worker and parameter server is a fine tuning work. We need avoid OOM as well as wasting memories.

docs/designs/allreduce.md
```diff
@@ -1,14 +1,215 @@
...
+The next two sections will illustrate the details of these components.
+
+### Fault-tolerant Allreduce Implementation
+
+We are collaborating with [Caicloud](https://github.com/caicloud/) on building an API that provides implementations of
+fault-tolerant allreduce. The initial implementation will contain an experimental Python binding for NCCL that is
+fault-tolerant and Kubernetes-native. This will include but not limited to the following objectives (more details to be disclosed later once
+the implementation has been open-sourced):
+
+* Fault-tolerant: if any of the worker pod fails, the [NCCL communicator](https://docs.nvidia.com/deeplearning/sdk/nccl-developer-guide/docs/usage/communicators.html)
```
brightcoder01, line 57: Should we restart the other worker pods while reconstructing the NCCL communicator?

docs/designs/allreduce.md
```diff
@@ -1,14 +1,215 @@
...
+num_epochs = 2
+num_batches = 10
+data_loader = DataLoader(num_batches)
+communicator = AllreduceCommunicator()
+
+for _ in range(num_epochs):
+    for features, labels in data_loader:
+        outputs = model(features)
+        loss = loss_fn(outputs, labels)
+        grads = calculate_gradients(loss, model)
```
brightcoder01, line 74: Should we calculate the gradients of all the parameters of the model. And then we can execute AllReduce?

docs/designs/allreduce.md
```diff
@@ -1,14 +1,215 @@
...
+        res, averaged_grads = communicator.average_gradients(grads)
+        if res == SUCCESS:
+            update_model(averaged_grads)
+        elif res == FAILED:
+            continue
+```
+
+We also provide API to broadcast all model parameters from one of the worker pods to all other active worker pods.
```
brightcoder01, line 82: If one worker pod restart, which worker will call the broadcast Api?
Is broadcast Api sync or async?

docs/designs/allreduce.md
brightcoder01: How to implement the lock of worker pods?
"Select one of the worker pods to broadcast" -> Master select the target pod or?

docs/designs/allreduce.md
```diff
@@ -1,14 +1,215 @@
...
+    1. Release the lock that each worker pod holds.
+    1. All worker pods continue to process training tasks at hand and perform allreduce operations.
+
+Note that since the existing active workers have the exact same copy of the model after the previous allreduce operation completes,
+we can guarantee that the new worker will have the same copy of the model as the ones on other workers once the next
+allreduce operation completes.
+
+#### Training with Evaluation
+
+If the worker encounters any evaluation tasks in the above process, it will evaluate the model directly on the worker once the
```
brightcoder01, line 201: If we have N worker pods, M worker pods receive the evaluation task. The other N-M worker are training, will these worker pods wait in "communicator.average_gradients(grads)" until the M worker pods complete evaluation tasks?

docs/designs/allreduce.md
```diff
@@ -1,14 +1,215 @@
...
+
+ElasticDL embedding layer is not supported under allreduce-based training since each worker has the exact same copy of
+the model that must fit within each worker's specified resources. Any layers in the model defined via
+``elasticdl.layers.embedding`` will be replaced by `tf.keras.layers.Embedding`.
+
+#### Relevant CLI Arguments
+
+The following CLI arguments are relevant and their behaviors might be changed under allreduce-based training:
+
+* ``--restart_policy``: The pod restart policy when pod crashed.
```
brightcoder01: Do we have a enum for restart_policy?

docs/designs/allreduce.md
```diff
@@ -1,14 +1,215 @@
...
+
+If any of the gradients averaging operation fails while the workers are still active and healthy, we simply report the
+failure and continue training on the next batch.
+
+If any of the workers fails, e.g. the pod is accidentally killed, the following steps will be performed:
+
+1. The task that the failed worker was handling will be added back to the task queue.
+1. Since the `AllreduceCommunicator` is aware of the failure. It will reconstruct the communicator and re-assign ranks among
+the existing active workers. The existing workers will then continue to run on the tasks at hand.
+1. The master pod will try to create a new worker pod after a specified restart delay period.
```
brightcoder01: What will happen if the worker pod create fails?
Will other worker pods continue the AllReduce with the N-1 typology?

docs/designs/allreduce.md
```diff
@@ -1,14 +1,215 @@
...
+While distributed training based on parameter servers can support training very large models and datasets by adding
+more workers and parameter servers as described in [parameter server design doc](parameter_server.md), there are additional
+challenges involved in order to optimize the performance, for example:
+
+* It is not easy to identify the right ratio of the number of workers to the number of parameter servers. For example,
+if only a small number of parameter servers are used, network communication will likely become the bottleneck for training.
+If many parameter servers are used, the communication may saturate network interconnects.
+* If the model could fit within the computational resources on each worker, additional maintenance and communication overheads are
+introduced when the model is partitioned to multiple parameter servers.
+* Additional computational resources for replicating models of each parameter server are needed to support fault-tolerance.
```
brightcoder01: Use active voice other than passive voice. :)
We will replicate the model in each parameter server to support fault-tolerance. It requires additional computation and storage resources.

docs/designs/allreduce.md
```diff
@@ -1,14 +1,215 @@
...
+        process_minibatch_and_report(dataset_batch, task)
+```
+
+If any of the gradients averaging operation fails while the workers are still active and healthy, we simply report the
+failure and continue training on the next batch.
+
+If any of the workers fails, e.g. the pod is accidentally killed, the following steps will be performed:
+
+1. The task that the failed worker was handling will be added back to the task queue.
+1. Since the `AllreduceCommunicator` is aware of the failure. It will reconstruct the communicator and re-assign ranks among
```
brightcoder01: If one worker pod is killed and restart, the communicator need be reconstructed twice?

docs/designs/allreduce.md
```diff
@@ -1,14 +1,215 @@
...
+that we mentioned in the previous section. The main differences are the following:
+
+1. Gradients from each worker are not sent to master.
+1. The average of gradients across all workers is calculated locally on each worker.
+1. The model is updated directly on each worker and each worker has the exact same copy of the model.
+
+Below is the pseudo-code for this process on each worker:
+
+```python
+communicator = AllreduceCommunicator()
```
brightcoder01, line 132: Should we pass the ips of other worker pods into the AllReduceCommunicator?

docs/designs/allreduce.md
terrytangyuan: Thanks!

docs/designs/allreduce.md
terrytangyuan: NCCL doesn't seem to provide barrier but it should be straightforward to get this implemented, e.g. allreduce 1 byte and then CUDA-synchronize. Reference implementation in PyTorch: https://github.com/pytorch/pytorch/pull/14142

@zw0610 What do you think?

docs/designs/allreduce.md
terrytangyuan: Thanks. Added!

docs/designs/allreduce.md
terrytangyuan: Thanks!

docs/designs/allreduce.md
terrytangyuan: Agreed. I removed "which avoids the waste of resources" in this sentence.

docs/designs/allreduce.md
terrytangyuan: Please see "Failure Handling during Training" section, which describes this in more details. ElasticDL handles the restart of worker pods instead. This NCCL allreduce binding is only responsible to reconstruct from the active worker pods.

I edited this sentence so it's clearer.

docs/designs/allreduce.md
terrytangyuan: Yes, master. Updated. The implementation is still being discussed. See https://github.com/sql-machine-learning/elasticdl/pull/1420#discussion_r345242492

docs/designs/allreduce.md
terrytangyuan: Master selects one worker to call broadcast API.

terrytangyuan: I added "Note that all NCCL API calls are asynchronous but our implementation of the communication collective primitives is synchronous
so it's easier for other frameworks like ElasticDL to use without worrying about failure handling."

cc @zw0610

docs/designs/allreduce.md
terrytangyuan: Yes this should be a blocking call until all gradients are calculated in the workers.

docs/designs/allreduce.md
terrytangyuan: Good question. I don't think we need to pass IPs for allreduce since the shared storage in the implementation would keep track of the list of IPs. I just added a note on this.

cc @zw0610 to double check.

docs/designs/allreduce.md
terrytangyuan: I don't think so since I see this sentence in in `model_evaluation.md`: "considering the complexity to launch and manage a new job to load checkpoint, we choose to evaluate the model with the current model in PS. For simplicity, we don't stop training tasks when an evaluation job starts". 

Pausing training can be either useful or a bottleneck that slows down training. This might be something worth discussing in a separate thread on model evaluation design since it's also related to PS.

docs/designs/allreduce.md
terrytangyuan: Not yet but this should be one of the supported values consistent with k8s's pod restart policy: https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#restart-policy

terrytangyuan: I added this in https://github.com/sql-machine-learning/elasticdl/pull/1451.

terrytangyuan:
> For allreduce, every worker must have training data in order to generate gradients. In ElasticDL's task-based system, how to distribute the training data evenly among workers so that every worker will generate gradients when the training is not finished.

@skydoorkai I have added a section "Data Distribution among Workers" that proposes a solution to this issue. Please take a look when you get a chance.

docs/designs/allreduce.md
zw0610: I didn't find a MPI-barrier-like operation in NCCL. The pytorch approach you mentioned is kind of popular since I believe it's spotted in Caffe too.

In my trial effort, the locking is achieved by aborting all NCCL operations in the pod that has become aware of the new worker. Since one pod has aborted the remaining NCCL operations, the NCCL operations in the rest pods will hang/fail, which has the equivalent effort to MPI Barrier. The questions remains whether we want to just abort gradient generated in this batch iteration or not.

terrytangyuan: @zw0610 I think it should be fine to abort just for this batch iteration. ElasticDL can just move this task back to the queue so later we can re-calculate the gradients.

docs/designs/allreduce.md
terrytangyuan: @brightcoder01 FYI I added this to "Potential Future Optimizations" section.

docs/designs/allreduce.md
zw0610: If there is just one pod demanding weights, it seems broadcast is unnecessary. Peer-to-peer communication can cover such occasion, which can take place when the communicator is aware of the new worker join.

Another strategy is append a compulsory broadcast ops after anytime a new communicator is (re-)created, which might use a bit more network bandwidth.

Since the lib is still under development, we can discuss which is preferred.

docs/designs/allreduce.md
zw0610: I checked MPI and NCCL: the _communicator_ in the sense of these communication libraries ask no IPs during the creation procedure nor the allreduce procedure.

The proto-implementation uses shared storage, keeping track of the list of IPs, which represent workers. So the `AllreduceCommunicator` in this context would need the IP of itself (instead of others) to register on the shared storage and build a consensus of all the workers.

Let me know if there is any other thoughts.

skydoorkai (approved):

docs/designs/allreduce.md
zw0610: While all workers are created equal, one of them will be considered as a 'master' if self.rank == 0. This worker is responsible for sending weights to the new comers.

docs/designs/allreduce.md
terrytangyuan: @zw0610 Thanks! The IP of the `AllreduceCommunicator` itself should be automatically detected by `AllreduceCommunicator` though so users of the lib would not need to worry about this.

ywskycn (approved):

PR #1432, opened 2019-11-08
Add func to optimizer wrapper

mhaoli:
Part of #1306, step 2.

### DIFF
--- a/elasticdl/python/ps/servicer.py
+++ b/elasticdl/python/ps/servicer.py
@@ -139,20 +139,23 @@ class PserverServicer(elasticdl_pb2_grpc.PserverServicer):
         # to support using Redis as distributed KV storage.
         embedding_dims = None
         embedding_service_endpoint = None
+
         def lookup_embedding_func(keys):
             embeddings = []
             for key in keys:
                 arrs = key.split("-")
-                layer_name = "-".join(arrs[:-1]),
+                layer_name = "-".join(arrs[:-1])
                 id = int(arrs[-1])
-                embedding = self._parameters.get_embedding_param(layer_name, [id])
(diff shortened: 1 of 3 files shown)


elasticdl/python/ps/servicer.py
```diff
@@ -140,9 +140,30 @@ def wrap_optimizer(self):
         embedding_dims = None
         embedding_service_endpoint = None
 
+        def lookup_embedding_func(keys):
```
QiJune, line 143: In our new PS design, we could directly pass name and ids. This will be more efficient.

```python
def lookup_embedding_func(name, ids):
    ...
    return embedding, []
```

We could set up an unit test which contains a model with an embedding layer, and then refactor the OptimizeWrapper class gradually. Test-drive development :)

QiJune (approved):

elasticdl/python/ps/servicer.py
mhaoli: > In our new PS design, we could directly pass name and ids. This will be more efficient.
> 
> ```python
> def lookup_embedding_func(name, ids):
>     ...
>     return embedding, []
> ```
> 
> We could set up an unit test which contains a model with an embedding layer, and then refactor the OptimizeWrapper class gradually. Test-drive development :)

Yes, I agree with the new interface. But this interface is not suitable for Redis now. If we use two different interfaces for Redis and PS, code would be complex. 

At the same time, there are also other logic that is needed for Redis and can be removed in the future. So, I would like to just keep the interface as before, modify them after we do not need to support Redis (as the above TODO states). 

Also, in the future, we do not need new unit tests and just modify `optimizer_wrapper_test.py`.

PR #1435, opened 2019-11-08
Embedding layer looks up embedding from PS

mhaoli:
Fixes #1377.

PR #1437, opened 2019-11-11
support pushing embedding table info and gradients in worker

### LINKED ISSUE #1436: push embedding table info and gradients in worker

QiJune:
Fix #1436

### DIFF
--- a/elasticdl/python/common/hash_utils.py
+++ b/elasticdl/python/common/hash_utils.py
@@ -4,3 +4,24 @@ import hashlib
 def string_to_id(name, num):
     h = hashlib.sha256(name.encode("utf-8"))
     return int(h.hexdigest(), base=32) % num
+
+
+def int_to_id(index, num):
+    return index % num
+
+
+def scatter_embedding_vector(values, indices, num):
+    ps_ids = {}
+    indices_list = indices.tolist()
+    for i, item_id in enumerate(indices_list):
+        ps_id = int_to_id(item_id, num)
+        if ps_id not in ps_ids:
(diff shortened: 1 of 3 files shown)


QiJune:
After #1435 and #1437 are merged, we could start writing a DeepFM unit test case.

elasticdl/python/common/hash_utils.py
```diff
@@ -4,3 +4,24 @@
 def string_to_id(name, num):
     h = hashlib.sha256(name.encode("utf-8"))
     return int(h.hexdigest(), base=32) % num
+
+
+def int_to_id(index, num):
+    return index % num
+
+
+def scatter_embedding_vector(values, indices, num):
```
mhaoli, line 13: It is a util function, better to add a docstring.

QiJune: Done

mhaoli (approved):
Excellent!

elasticdl/python/worker/worker.py
```diff
@@ -350,6 +363,47 @@ def report_gradient_to_ps(self, grads):
             req.model_version = self._model_version
             res = self._ps_stubs[ps_id].push_gradient(req)
 
+        if self._embedding_layers:
```
skydoorkai, line 388: Here we should combine non-embedding param gradients and embedding gradients in one `push_gradient` call.
This is because in ps `push_gradient` service, each `push_gradient` would increase the parameter version by 1. 
Two calls would update the version by 2.

QiJune: Done

PR #1446, opened 2019-11-12
add unit test for model_handler to process subclass model

### LINKED ISSUE #1444: add subclass test case for model handle test
We only define a functional model in test_module.
https://github.com/sql-machine-learning/elasticdl/blob/35d0647a737d93e6969764cabe541ac8078d13bf/elasticdl/python/tests/test_module.py#L16-L20

We should add subclass one to ensure ModelHandle logic correct.

workingloong:
fix #1444


QiJune (approved):
Excellent!

terrytangyuan (approved):

PR #1449, opened 2019-11-12
report_variable for ps init when needed

### LINKED ISSUE #1448: When `get_model` returns with `model_init_status==False`,  worker reports local variables to ps for ps initialization.

skydoorkai:
fix #1448 
For PS, lazy initialization is used. When a worker finds out that ps is not initialized by checking `model_init_status` in `pull_variable` call, it sends its variables to PS by `report_variable` for PS initialization.

### DIFF
--- a/elasticdl/python/worker/worker.py
+++ b/elasticdl/python/worker/worker.py
@@ -159,6 +159,8 @@ class Worker(object):
                 self._model, self._embedding_layers
             ):
                 self._non_embed_vars[var.name] = var
+            if self._use_multi_ps:
+                self.init_ps_var_partition()
 
     def _init_embedding_layer(self):
         """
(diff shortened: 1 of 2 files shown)


elasticdl/python/worker/worker.py
```diff
@@ -220,10 +222,17 @@ def get_model_from_master(self, version, method):
...
+        for ps_id, stub in enumerate(self._ps_stubs):
+            if ps_id not in self._ps_vars:
+                continue
             req = empty_pb2.Empty()
             res = stub.pull_variable(req)
 
+            if not res.model_init_status:
+                # push variable to ps for initialization
+                self.report_variable_to_ps(ps_id)
+                res = stub.pull_variable(req)
```
mhaoli, line 240: Do we need to check `res.model_init_status` here?

skydoorkai: Add raise exception and comment to indicate that we should handle these exceptions when dealing with ps fault-tolerance.

mhaoli (approved):
Excellent!

PR #1455, opened 2019-11-13
report embedding info before model.call

### LINKED ISSUE #1454: need to initialize embedding table info before run `model.call`
If a model has an elasticdl embedding layer, the `model.call` will call `pull_embedding_vector` inside `embedding.call`. This will cause errors, if we do not push embedding table info to PS before.

https://github.com/sql-machine-learning/elasticdl/blob/4a995fe7eaf91bc5a9d50181e9aaaa14d15c8a09/elasticdl/python/worker/worker.py#L505-L514

QiJune:
Fix #1454

### DIFF
--- a/elasticdl/python/worker/worker.py
+++ b/elasticdl/python/worker/worker.py
@@ -177,6 +177,7 @@ class Worker(object):
             layer.set_endpoint(self._embedding_service_endpoint)
             if self._use_multi_ps:
                 layer.set_lookup_embedding_func(self.pull_embedding_vector)
+                self.report_embedding_info()
 
         self._need_embedding_layer_check = (
             True if self._embedding_layers else False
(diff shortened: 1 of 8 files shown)


mhaoli:
This problem reminds us that we also need to pass embedding info in case a ps instance failed. It can pull embedding info from its replica.

elasticdl/python/worker/worker.py
```diff
@@ -509,6 +511,7 @@ def _run_model_call_before_training(self, features):
               more than once during one forward-pass.
         """
         if self._embedding_layers:
+            self.report_embedding_info()
```
mhaoli: Would it be better to put this line into function `_init_embedding_layer `? There are two reasons:
* `_run model_call_before_training` does some initialization that needs to do after `model.call` is called. However, reporting embedding info is a kind of initialization that is not dependent on `model.call`.
* If a user calls `worker.set_model()` to change a model in unit tests, he doesn't need to call `report_embedding_info` manually.

QiJune: Done

mhaoli (approved):
Excellent!

PR #1457, opened 2019-11-13
improvement: pass arguments to worker directly

### LINKED ISSUE #1456: clean arguments
There are different roles in ElasticDL. And different roles require different arguments from the command-line. 
All arguments except those passed to master are constructed by master. So if we need to pass a new argument to any role, we first add the argument in the master's argument list and then pass it on to the specific role. There is a main function for each role currently. In the main function, it will parse the argument and pass it to the constructor of the role.

The problem is we have to add an argument twice before we can use it.
+ In master's argument list
+ In the specific role's constructor

So we suggest passing directly the parsed argument object instead of unpacking the arguments and passing each one of them as the argument of the constructor.

chunyang-wen:
Fixes #1456 
Pass arguments object directly.

### DIFF
--- a/elasticdl/python/common/args.py
+++ b/elasticdl/python/common/args.py
@@ -491,7 +491,6 @@ def add_common_args_between_master_and_worker(parser):
     parser.add_argument(
         "--distribution_strategy",
         type=str,
-        default="ParameterServerStrategy",
         help="Master will use a distribution policy on a list of devices "
         "according to the distributed strategy, "
         'e.g. "ParameterServerStrategy"',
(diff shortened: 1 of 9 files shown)


elasticdl/python/common/args.py
```diff
@@ -491,7 +491,6 @@ def add_common_args_between_master_and_worker(parser):
     parser.add_argument(
         "--distribution_strategy",
         type=str,
-        default="ParameterServerStrategy",
```
skydoorkai, line 494: We can keep the default as `ParameterServerStrategy`. If there are other conflicts for default values, change all default to `ParameterServerStrategy`.

chunyang-wen: In `ParameterStrategy`,  `tf.keras.layers.Embedding` will forcibly replaced with `elasticdl.python.elasticdl.layers.Embedding`. This will fail certain test cases. Maybe a better default strategy should be setup.

chunyang-wen: A related issue is created. #1458

skydoorkai (approved):

PR #1460, opened 2019-11-14
add 2 workers and 2 pservers unit test

QiJune:


elasticdl/python/tests/worker_ps_interaction_test.py
```diff
@@ -31,38 +32,75 @@ def random_batch(batch_size):
...
+
+
+def get_frappe_dataset(batch_size):
+    home = str(Path.home())
+
+    class TmpArgs(object):
+        def __init__(self, data):
+            self.data = data
+
+    args = TmpArgs(data=home + "/.keras/datasets/")
```
terrytangyuan: You can use a namedtuple for this instead of a separate class.

elasticdl/python/tests/worker_ps_interaction_test.py
```diff
@@ -31,38 +32,75 @@ def random_batch(batch_size):
...
+    test_db = tf.data.Dataset.from_tensor_slices((x_test, y_test))
+    test_db = test_db.batch(batch_size)
+
+    return db, test_db
+
+
+def get_frappe_dataset(batch_size):
+    home = str(Path.home())
+
+    class TmpArgs(object):
```
terrytangyuan: You can use a namedtuple for this instead of a separate class.

elasticdl/python/tests/worker_ps_interaction_test.py
terrytangyuan: Use `path.join` instead of string concatenation would be better

elasticdl/python/tests/worker_ps_interaction_test.py
QiJune: Done

elasticdl/python/tests/worker_ps_interaction_test.py
QiJune: Done

elasticdl/python/tests/worker_ps_interaction_test.py
```diff
@@ -78,18 +92,96 @@ def _create_pserver_and_channel(self, ports):
...
+            self._channels.append(channel)
+
+        self._pservers = []
+        self._workers = []
+
+    def tearDown(self):
+        for pserver in self._pservers:
+            pserver.server.stop(0)
+
+    def _create_pserver(self, model_def):
```
skydoorkai: We can add `num_ps` argument in this method, and create an arbitrary number of ps. This will make future unittests with different number of ps/worker much easier.

QiJune: Done

skydoorkai (approved):

PR #1464, opened 2019-11-14
Add check for empty string for ODPS endpoint

terrytangyuan:
Added explicit check for empty string so users can be explicit about using public ODPS endpoint. Also included this in CI so we can test it as well as showing off all the available options in `scripts/client_test.sh`.


typhoonzero (approved):
Thanks!

PR #1465, opened 2019-11-15
Worker waits until ps pod is ready

skydoorkai:
When ps pod is not ready, the worker cannot get ps pod IP.  Need to wait a few seconds then try again.


QiJune (approved):
Excellent!

PR #1467, opened 2019-11-15
stop ps pod when master pod is stopped

QiJune:
I find that we must call `read_namespaced_pod `  inside the loop. Otherwise, we will get outdated pod status.

This PR will make ps pod stopped elegantly:

```
kubectl get pods
NAME                            READY   STATUS      RESTARTS   AGE
elasticdl-test-mnist-master     0/1     Completed   0          3m6s
elasticdl-test-mnist-ps-0       0/1     Completed   0          3m3s
elasticdl-test-mnist-worker-0   0/1     Completed   0          3m2s
```

### DIFF
--- a/elasticdl/python/ps/parameter_server.py
+++ b/elasticdl/python/ps/parameter_server.py
@@ -2,6 +2,7 @@ import time
 from concurrent import futures
 
 import grpc
+from kubernetes import client, config
 
 from elasticdl.proto import elasticdl_pb2_grpc
 from elasticdl.python.common.constants import GRPC
@@ -28,6 +29,9 @@ class ParameterServer(object):
         self.optimizer = model_module[args.optimizer]()
         # Create Parameters instance
         self.parameters = Parameters()
+
+        self.master_name = "elasticdl-%s-master" % args.job_name
+        self.namespace = args.namespace
 
     def prepare(self):
         server = grpc.server(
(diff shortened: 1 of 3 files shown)


elasticdl/python/ps/parameter_server.py
```diff
@@ -29,6 +30,9 @@ def __init__(self, args):
         # Create Parameters instance
         self.parameters = Parameters()
 
+        self.master_name = "elasticdl-%s-master" % args.job_name
```
brightcoder01: Can we put the naming formats in a common file. I think multiple component will use this.

terrytangyuan: You can reuse `k8s_client.Client.get_master_pod_name()`

QiJune: Yes, but only master has a k8s_client object.  I will make it as a common function.

elasticdl/python/ps/parameter_server.py
```diff
@@ -63,10 +68,16 @@ def prepare(self):
...
+        config.load_incluster_config()
+        api = client.CoreV1Api()
         try:
             while True:
-                # TODO: add loop break condition
                 time.sleep(30)
+                master_pod = api.read_namespaced_pod(
+                    namespace=self.namespace, name=self.master_name
+                )
+                if master_pod.status.phase == "Succeeded":
```
skydoorkai, line 79: If master pod status is `Failed`, ps also should stop.

QiJune: Done

skydoorkai (approved):

Issue #1466, opened 2019-11-15
Save checkpoints during training.


workingloong:
* Save the model variables to a checkpoint file.
1. Each PS instance will save its owner variables to a checkpoint file every `checkpoints_steps` steps. The checkpoints of all PS instances will be placed in the same folder named `version_{version_num}`.
2. When to save a checkpoint, each PS instance create a `Tensor` with only values for a `tf.Variable` or with values and indices for an embedding table. Then, a `Model` will be created to save those `Tensor`s and `EmbeddingTableInfo` which contains meta-information of embedding tables.
3. Save The `Model` to a file in the folder `version-{version_num}` and remove redundant checkpoints.

workingloong:
The issue has been resolved by #1503. So, I close it.

PR #1472, opened 2019-11-18
Export parameters on the parameter server to a model pb

workingloong:
Implement #1466

### DIFF
--- a/elasticdl/proto/elasticdl.proto
+++ b/elasticdl/proto/elasticdl.proto
@@ -66,7 +66,7 @@ message Tensor {
     bytes content = 3;
 
     // Indices will be tf.IndexedSlices.indices if the tensor is in the form
-    // of tf.IndexedSlices. Otherwise indices will be None. 
+    // of tf.IndexedSlices. Otherwise indices will be None.
     repeated int32 indices = 4;
 
     // Dtype of the tensor, e.g. "int64", "float32".
--- a/elasticdl/proto/elasticdl.proto
+++ b/elasticdl/proto/elasticdl.proto
@@ -66,7 +66,7 @@ message Tensor {
     bytes content = 3;
 
     // Indices will be tf.IndexedSlices.indices if the tensor is in the form
-    // of tf.IndexedSlices. Ohterwise indices will be None.
+    // of tf.IndexedSlices. Otherwise indices will be None. 
     repeated int32 indices = 4;
 
     // Dtype of the tensor, e.g. "int64", "float32".
--- a/elasticdl/python/ps/embedding_table.py
+++ b/elasticdl/python/ps/embedding_table.py
@@ -1,7 +1,8 @@
 import numpy as np
 import tensorflow as tf
-from elasticdl.python.common.tensor import Tensor
+
 from elasticdl.proto.elasticdl_pb2 import EmbeddingTableInfo
+from elasticdl.python.common.tensor import Tensor
 
 
 class EmbeddingTable(object):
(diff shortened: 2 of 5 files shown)


elasticdl/python/ps/parameters.py
```diff
@@ -157,3 +170,23 @@ def create_slot_params(self, slot_names, init_values):
...
+            )
+
+        for name, embedding_table in self.embedding_params.items():
+            embedding_table_tensor = embedding_table.to_tensor()
+            tensor_pb = pb_model.param.add()
+            serialize_tensor(embedding_table_tensor, tensor_pb)
+
+            embedding_info = embedding_table.to_embedding_table_info_pb()
+            pb_model.embedding_table_info.append(embedding_info)
+
```
QiJune, line 192: 1 pb_model --> model_pb
2 version should also be serialized, model_pb.version = self.version

elasticdl/proto/elasticdl.proto
```diff
@@ -66,7 +66,7 @@ message Tensor {
     bytes content = 3;
 
     // Indices will be tf.IndexedSlices.indices if the tensor is in the form
-    // of tf.IndexedSlices. Ohterwise indices will be None.
+    // of tf.IndexedSlices. Otherwise indices will be None.
```
mhaoli: Could you delete the trailing whitespace?

elasticdl/python/ps/embedding_table.py
```diff
@@ -28,6 +31,7 @@ def __init__(self, name, dim=None, initializer=None, is_slot=False):
         """
         self.name = name
         self.dim = dim
+        self.initializer_pb = initializer
```
mhaoli: Should check that `initializer` is a string because you add a function `to_embedding_info_pb` and it requires that `initializer` is a string.

I didn't take into consideration that `initializer` should be string if `is_slot=True`. In my impression, I passed in floating-point numbers in some unit tests... Or you can ignore this comment and I will fix this issue after your PR.

workingloong: I can cast `self.initializer_value` which has been renamed to string when I set it to `embedding_pb.initializer`.

terrytangyuan (approved):

elasticdl/python/ps/embedding_table.py
```diff
@@ -28,6 +31,7 @@ def __init__(self, name, dim=None, initializer=None, is_slot=False):
         """
         self.name = name
         self.dim = dim
+        self.initializer_value = initializer
         if is_slot:
             initializer = float(initializer)
             self.initializer = tf.keras.initializers.Constant(initializer)
```
terrytangyuan: This needs to be changed to `self.initializer_value` as well?

terrytangyuan (changes requested):

QiJune (approved):
Excellent!

QiJune:
cc to @terrytangyuan

elasticdl/python/ps/embedding_table.py
```diff
@@ -57,6 +64,27 @@ def set(self, indices, values):
...
+            indices=np.array(indices),
+            name=self.name,
+        )
+
+    def to_embedding_table_info_pb(self):
+        """Convert the embedding table information to a protobuf"""
+        embedding_pb = EmbeddingTableInfo()
+        embedding_pb.name = self.name
+        embedding_pb.dim = self.dim
+        embedding_pb.initializer = str(self.initializer_value)
```
brightcoder01, line 85: How do we differentiate whether the initializer_value is float or str while restoring from pb?

workingloong: In `__init__` of EmbeddingTable, we differentiate by `is_slot`.
https://github.com/sql-machine-learning/elasticdl/blob/b3fd10197d0f1184838e4e533865ca162570ccef/elasticdl/python/ps/embedding_table.py#L35-L42

brightcoder01 (approved):

terrytangyuan (approved):

PR #1473, opened 2019-11-18
add frappe dataset to Dockerfile

QiJune:
This PR aims to add a frappe dataset to the Dockerfile. Then, we could add a deepFM model test in travis CI,  which covers both embedding params and non-embedding params.

### DIFF
--- a/elasticdl/docker/Dockerfile.dev
+++ b/elasticdl/docker/Dockerfile.dev
@@ -24,5 +24,10 @@ RUN pip install -r /requirements-dev.txt --extra-index-url=${EXTRA_PYPI_INDEX}
 # rest part of ElasticDL.  The generated data will be in /data.
 COPY elasticdl/python/data/recordio_gen/image_label.py /var/image_label.py
 RUN python /var/image_label.py --dataset mnist --fraction 0.25 \
-	--records_per_shard 4096 /data && \
-    rm -rf /root/.keras/datasets
+	--records_per_shard 4096 /data
+
+# Copy frappe dataset
+COPY elasticdl/python/data/recordio_gen/frappe_recordio_gen.py /var/frappe_recordio_gen.py
+RUN python /var/frappe_recordio_gen.py --data /root/.keras/datasets --output_dir /data/frappe \
+    --fraction 0.25
+RUN rm -rf /root/.keras/datasets


skydoorkai:
What is the size of 25% frappe dataset?

mhaoli:
Can we only add it in Dockerfile.ci?

elasticdl/docker/Dockerfile.dev
```diff
@@ -24,5 +24,10 @@ RUN pip install -r /requirements-dev.txt --extra-index-url=${EXTRA_PYPI_INDEX}
...
 RUN python /var/image_label.py --dataset mnist --fraction 0.25 \
-	--records_per_shard 4096 /data && \
-    rm -rf /root/.keras/datasets
+	--records_per_shard 4096 /data
+
+# Copy frappe dataset
+COPY elasticdl/python/data/recordio_gen/frappe_recordio_gen.py /var/frappe_recordio_gen.py
+RUN python /var/frappe_recordio_gen.py --data /root/.keras/datasets --output_dir /data/frappe \
+    --fraction 0.25
+RUN rm -rf /root/.keras/datasets
```
terrytangyuan, line 33: Why removing this? I thought we added cache in travis.yml

QiJune:
@skydoorkai  It's the faction of the frappe dataset. For example, the dataset has 1000 rows, and the fraction is set to 0.25. As a result, 250 rows will be transformed into recordio file.

elasticdl/docker/Dockerfile.dev
QiJune: I just follow the former code here. If we remove the directory, the size of this image will be smaller. 

And for the travis, at the first time, the dataset will download inside docker. Since we mount the host directory to docker container, the dataset will also be write to host disk. Then, we cache the dataset in host. At the second time, the dataset will be used directly, which speed up the ci.

terrytangyuan: Okay. Maybe add a comment here since otherwise it's hard to understand the logistics here

QiJune:
@LiMinghao1994 Mnist dataset and frappe dataset are two representing datasets, one for CV and non-embedding model, the other for recommendation and embedding model. I think it's better to add these two datasets in the same place for conveniency.

QiJune:
Please refer to #1481

Issue #1398, opened 2019-10-31
Master handles the errMsg that worker reports


chunyang-wen:
Currently, when a worker reports an error message to the master, the message seems not correctly handled. 

If we support the failure of certain records, the master should not fail the job if the failure rate is acceptable.

If we do not support any failures, the job should be in the status of failure.

Now workers will report statistics to the master. Master can make its decision based on this.

PR #1478, opened 2019-11-18
Add counters for an ElasticDL job

chunyang-wen:
Related to #1398 
This PR adds counters in `_TaskDispatcher`.

### DIFF
--- a/elasticdl/python/master/main.py
+++ b/elasticdl/python/master/main.py
@@ -1,3 +1,5 @@
+import sys
+
 from elasticdl.python.common.args import parse_master_args
 from elasticdl.python.master.master import Master
 
@@ -6,8 +8,8 @@ def main():
     args = parse_master_args()
     master = Master(args)
     master.prepare()
-    master.run()
+    return master.run()
 
 
 if __name__ == "__main__":
-    main()
+    sys.exit(main())
--- a/elasticdl/python/master/main.py
+++ b/elasticdl/python/master/main.py
@@ -1,5 +1,3 @@
-import sys
-
 from elasticdl.python.common.args import parse_master_args
 from elasticdl.python.master.master import Master
 
@@ -12,4 +10,4 @@ def main():
 
 
 if __name__ == "__main__":
-    sys.exit(main())
+    main()
--- a/elasticdl/python/master/main.py
+++ b/elasticdl/python/master/main.py
@@ -1,3 +1,5 @@
+import sys
+
 from elasticdl.python.common.args import parse_master_args
 from elasticdl.python.master.master import Master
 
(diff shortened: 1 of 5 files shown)


elasticdl/python/master/main.py
```diff
@@ -6,8 +8,8 @@ def main():
     args = parse_master_args()
     master = Master(args)
     master.prepare()
-    master.run()
+    return master.run()
```
terrytangyuan, line 11: Why not exit directly from `master.run()`? Is this a better practice?

chunyang-wen: I think they are similar. But if we add `sys.exit(master.run())`, we have to make sure that this is the last function call. Using `sys.exit(main())` under the `__name__=="__main__"` works at the command line. Maybe it is more suitable.

brightcoder01: What's the returned value used for?

elasticdl/python/master/master.py
```diff
@@ -146,6 +146,17 @@ def __init__(self, args):
         # Initialize instance manager
         self.instance_manager = self._create_instance_manager(args)
 
+        self._should_stop = False
+        self._exit_code = 0
+
+    def request_stop(self, err_msg=None):
```
brightcoder01, line 152: Which component will call this Api?

elasticdl/python/master/main.py
chunyang-wen: To tell whether this command executes successfully or not.

elasticdl/python/master/master.py
chunyang-wen: Any component. It provides a mechanism to stop master early. In another PR, I will use this API.

elasticdl/python/master/task_dispatcher.py
```diff
@@ -30,6 +31,30 @@ def _info(self):
         )
 
 
+class JobCounters(object):
```
brightcoder01: JobCounter instead of JobCounters?

elasticdl/python/master/task_dispatcher.py
```diff
@@ -30,6 +31,30 @@ def _info(self):
...
+    """Counters for job"""
+
+    def __init__(self, total_records=None, failed_records=None):
```
skydoorkai: Why not use `0` as argument default?

chunyang-wen: Good catch.

elasticdl/python/master/task_dispatcher.py
chunyang-wen: Fixed

skydoorkai (approved):

PR #1481, opened 2019-11-19
add deepFM ci test using new PS

QiJune:
This PR add deepFM ci test using new PS. Currently, we only substitute a training job. After checkpoint related features are done, we will substitute all the jobs.

### DIFF
--- a/elasticdl/docker/Dockerfile.dev
+++ b/elasticdl/docker/Dockerfile.dev
@@ -24,5 +24,10 @@ RUN pip install -r /requirements-dev.txt --extra-index-url=${EXTRA_PYPI_INDEX}
 # rest part of ElasticDL.  The generated data will be in /data.
 COPY elasticdl/python/data/recordio_gen/image_label.py /var/image_label.py
 RUN python /var/image_label.py --dataset mnist --fraction 0.25 \
-	--records_per_shard 4096 /data && \
-    rm -rf /root/.keras/datasets
+	--records_per_shard 4096 /data
+
+# Copy frappe dataset
+COPY elasticdl/python/data/recordio_gen/frappe_recordio_gen.py /var/frappe_recordio_gen.py
+RUN python /var/frappe_recordio_gen.py --data /root/.keras/datasets --output_dir /data/frappe \
+    --fraction 0.25
+RUN rm -rf /root/.keras/datasets
--- a/elasticdl/docker/Dockerfile.dev
+++ b/elasticdl/docker/Dockerfile.dev
@@ -29,5 +29,5 @@ RUN python /var/image_label.py --dataset mnist --fraction 0.25 \
 # Copy frappe dataset
 COPY elasticdl/python/data/recordio_gen/frappe_recordio_gen.py /var/frappe_recordio_gen.py
 RUN python /var/frappe_recordio_gen.py --data /root/.keras/datasets --output_dir /data/frappe \
-    --fraction 0.25
+    --fraction 0.1
 RUN rm -rf /root/.keras/datasets
--- a/elasticdl/python/master/tensorboard_service.py
+++ b/elasticdl/python/master/tensorboard_service.py
@@ -29,6 +29,8 @@ class TensorboardService(object):
             for k, v in dictionary.items():
                 if isinstance(v, np.ndarray) and len(v) == 1:
                     v = v[0]
+                elif isinstance(v, dict) and len(v):
+                    v = list(v.values())[0]
                 tf.summary.scalar(k, v, step=version)
 
     def start(self):
--- a/elasticdl/python/master/tensorboard_service.py
+++ b/elasticdl/python/master/tensorboard_service.py
@@ -29,7 +29,7 @@ class TensorboardService(object):
             for k, v in dictionary.items():
                 if isinstance(v, np.ndarray) and len(v) == 1:
                     v = v[0]
-                elif isinstance(v, dict) and len(v):
+                elif isinstance(v, dict) and v:
                     v = list(v.values())[0]
                 tf.summary.scalar(k, v, step=version)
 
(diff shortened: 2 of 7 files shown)


scripts/validate_job_status_ps.sh
```diff
@@ -0,0 +1,70 @@
+#!/usr/bin/env bash
```
terrytangyuan, line 1: Why not modifying existing script instead of copying + modifying?

QiJune: Umm, we have master+ps+worker case, and master+worker case. Now, the validate job status script is hard-code. Let me see how to make it more general.

elasticdl/python/master/tensorboard_service.py
```diff
@@ -29,6 +29,8 @@ def write_dict_to_summary(self, dictionary, version):
             for k, v in dictionary.items():
                 if isinstance(v, np.ndarray) and len(v) == 1:
                     v = v[0]
+                elif isinstance(v, dict) and len(v):
+                    v = list(v.values())[0]
```
QiJune, line 33: In deepFM test, the dictionary is as following:

```
dict_keys(['logits', 'probs'])
[2019-11-19 04:05:57,638] [INFO] [evaluation_service.py:255:complete_task] {'logits': {'accuracy': <tf.Tensor: id=4011, shape=(), dtype=float32, numpy=0.66985446>}, 'probs': {'auc': <tf.Tensor: id=4044, shape=(), dtype=float32, numpy=0.5908479>}}
```

The value is also a dictionary, we have to handle such case.

scripts/validate_job_status_ps.sh
QiJune: I will fix this issue in next PR, which provides a general validate job status script.

elasticdl/python/master/tensorboard_service.py
chunyang-wen: replace `len(v)` with `v`

elasticdl/python/ps/parameter_server.py
```diff
@@ -82,6 +82,15 @@ def run(self):
                 elif master_pod.status.phase == "Failed":
                     self.logger.info("Master pod is Failed")
                     break
+                elif (
+                    master_pod.status.phase == "Running"
```
chunyang-wen: Status should be an enum variable instead of magic strings.

elasticdl/python/ps/parameter_server.py
```diff
@@ -82,6 +82,15 @@ def run(self):
...
+                    and master_pod.metadata.labels["status"] == "Finished"
+                ):
+                    self.logger.info(
+                        "Task is finished, "
+                        + "master pod is still running tensorboard service"
```
chunyang-wen: remove redundant `+`

scripts/client_test.sh
```diff
@@ -6,21 +6,25 @@ if [[ "$JOB_TYPE" == "train" ]]; then
     elasticdl train \
       --image_base=elasticdl:ci \
       --model_zoo=model_zoo \
-      --model_def=mnist_functional_api.mnist_functional_api.custom_model \
-      --training_data=/data/mnist/train \
-      --validation_data=/data/mnist/test \
-      --num_epochs=2 \
+      --model_def=deepfm_functional_api.deepfm_functional_api.custom_model \
+      --training_data=/data/frappe/train \
```
chunyang-wen, line 10: Any reason to change the dataset for testing?

elasticdl/docker/Dockerfile.dev
```diff
@@ -24,5 +24,10 @@ RUN pip install -r /requirements-dev.txt --extra-index-url=${EXTRA_PYPI_INDEX}
 # rest part of ElasticDL.  The generated data will be in /data.
 COPY elasticdl/python/data/recordio_gen/image_label.py /var/image_label.py
 RUN python /var/image_label.py --dataset mnist --fraction 0.25 \
-	--records_per_shard 4096 /data && \
-    rm -rf /root/.keras/datasets
+	--records_per_shard 4096 /data
+
+# Copy frappe dataset
+COPY elasticdl/python/data/recordio_gen/frappe_recordio_gen.py /var/frappe_recordio_gen.py
+RUN python /var/frappe_recordio_gen.py --data /root/.keras/datasets --output_dir /data/frappe \
```
chunyang-wen, line 31: After building the new docker image, we won't need to download the dataset each time? It seems that a unit test fails due to the error of downloading a dataset from GitHub when I run tests locally.

elasticdl/python/ps/parameter_server.py
QiJune: Done

elasticdl/python/ps/parameter_server.py
QiJune: Done

scripts/client_test.sh
QiJune: We want to test deepFM model with elasticdl embedding layer using new PS in CI. The frappe dataset is used for deepFM model.

elasticdl/docker/Dockerfile.dev
QiJune: Yes, I have cache the dataset in travis CI. But, it has to download the dataset at the first time.

I just follow the former code here. If we remove the directory, the size of this image will be smaller.

And for the travis, at the first time, the dataset will download inside docker. Since we mount the host directory to docker container, the dataset will also be write to host disk. Then, we cache the dataset in host. At the second time, the dataset will be used directly, which speed up the ci.


Local test is the same with travis test.

elasticdl/python/master/tensorboard_service.py
QiJune: Done

skydoorkai (approved):

PR #1486, opened 2019-11-19
Use async GRPC call to parallel data communcation between worker and ps

### LINKED ISSUE #1485: Use async grpc call to parallel worker-ps communication

skydoorkai:
Tested on deepFM sample.
For 2-ps 1-worker, training time for 1 epoch is reduced from 53s to 38s.
Note that the CPU number in `worker_resource_request` should be larger than 1 to get the speedup benefit.

Fix #1485

### DIFF
--- a/elasticdl/python/worker/worker.py
+++ b/elasticdl/python/worker/worker.py
@@ -203,16 +203,16 @@ class Worker(object):
 
     def get_model_from_ps(self, version, method):
         model_version = -1
-        variable_future_and_id = []
+        variable_future_and_id_pairs = []
         req = empty_pb2.Empty()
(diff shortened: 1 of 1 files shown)


elasticdl/python/worker/worker.py
```diff
@@ -237,11 +242,15 @@ def pull_embedding_vector(self, layer_name, embedding_ids):
 
         embeddings = []
         index = []
+        pb_future_and_id = []
```
brightcoder01: pb_future_and_id_array?

skydoorkai: Renamed to `pb_future_and_id_pairs`.

QiJune (approved):
Excellent!

PR #1489, opened 2019-11-20
clean worker reporting gradients to master related codes

QiJune:
Worker training with master is not supported. Since we have a deepFM training ci test with new PS design, we could remove related code safely.


skydoorkai:
Need to change the default value for `num_ps_pods` to 1 from 0, so master-based ps will not be used.

terrytangyuan (approved):
LGTM

Issue #1476, opened 2019-11-18
Custom model changes needed to unify with SQLFlow model zoo


typhoonzero:
### Background

Unify model zoo implementation of SQLFlow and ElasticDL: https://github.com/sql-machine-learning/models/issues/22
WIP PR: https://github.com/sql-machine-learning/models/pull/27

### Custom Model Requirements for Unifying ModelZoo

1. Support `feature_columns` argument when initializing a model.
1. Support a default `eval_metrics_fn`, so that this function is not "required" when writing a custom model definition.
1. Current `loss(output, labels)` function can not be reused in `kerasmodel.compile`, should be compatible with keras loss functions, like:` keras.losses.mean_squared_error(y_true, y_pred)`: https://keras.io/losses/
1. `dataset_fn` is still needed when reading data from MaxCompute: https://github.com/sql-machine-learning/elasticdl/blob/develop/model_zoo/odps_iris_dnn_model/odps_iris_dnn_model.py

terrytangyuan:
1. We already support this. Users just need to add feature columns definitions as part of their model definition. Something like the following would work. See tutorial in https://www.tensorflow.org/tutorials/structured_data/feature_columns for details. 
```
feature_layer = tf.keras.layers.DenseFeatures(feature_columns)
model = tf.keras.Sequential([
  feature_layer,
  layers.Dense(128, activation='relu'),
  layers.Dense(128, activation='relu'),
  layers.Dense(1, activation='sigmoid')
])
```
2. What would be a good default for `eval_metrics_fn`? I think this is highly dependent on the type of model that's defined, e.g. regression vs. classification.
3. I thought it should be compatible already. @LiMinghao1994 @workingloong @brightcoder01 can take a look at this if that's not the case.
4. It's not needed if you are using SQLFLow directly since `dataset_fn` is automatically generated by ElasticDL codegen.

typhoonzero:
> What would be a good default for eval_metrics_fn? I think this is highly dependent on the type of model that's defined, e.g. regression vs. classification.

Well, I'll consider about this. It's true that the `eval_metrics_fn` is needed for every model.

> It's not needed if you are using SQLFLow directly since dataset_fn is automatically generated by ElasticDL codegen.

Is it possible to test a model using MaxCompute as the dataset, so that the `dataset_fn` is automatically generated by ElasticDL, so that if we need to test a model if it's working, we don't need to write a `dataset_fn` again.

[...]

PR #1490, opened 2019-11-20
Allow user to specify tf.keras style loss function

terrytangyuan:
This addresses the third issue on `loss` in https://github.com/sql-machine-learning/elasticdl/issues/1476 to support loss functions from `tf.keras.losses`. Also added a test case.

cc: @typhoonzero 


typhoonzero (approved):
`keras_loss(labels, predictions)` is compatible with keras model.compile now

PR #1501, opened 2019-11-21
Add logic for allreduce failure handling in Worker

terrytangyuan:
Based on design doc proposed in #1420.

### DIFF
--- a/elasticdl/python/worker/worker.py
+++ b/elasticdl/python/worker/worker.py
@@ -5,6 +5,7 @@ import traceback
 import numpy as np
 import tensorflow as tf
 from google.protobuf import empty_pb2
+import random
 
 from elasticdl.proto import elasticdl_pb2, elasticdl_pb2_grpc
 from elasticdl.python.collective_ops.communicator import CollectiveCommunicator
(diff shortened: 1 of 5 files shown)


elasticdl/python/worker/worker.py
```diff
@@ -546,17 +555,62 @@ def forward_process(self, features):
...
+
+    def _get_local_model_params(self):
+        return {}
+
+    def _broadcast_model_params(self):
+        status = self._collective_communicator.barrier()
+        if status == CollectiveCommunicatorStatus.FAILED:
+            logger.warning("Failed to perform barrier operation")
+            return False
+        root_ip = random.choice(
```
skydoorkai: If each worker selects different `root_ip`, `broadcast` will not work.

elasticdl/python/worker/worker.py
```diff
@@ -569,7 +623,18 @@ def _collect_gradients_without_allreduce(self, grads):
     def _run_training_task(self, features, labels):
         loss, grads = self.training_process(features, labels)
         if self._distribution_strategy == DistributionStrategy.ALLREDUCE:
-            return (*self._collect_gradients_with_allreduce(grads), loss)
+            # TODO: Delay certain amount of time before retrying
+            for _ in range(self._max_allreduce_retry_num + 1):
+                accepted, loss = self._collect_gradients_with_allreduce_robust(
```
skydoorkai: Why get `loss` here?

elasticdl/python/worker/worker.py
skydoorkai: Need to choose an old worker?

elasticdl/python/worker/worker.py
terrytangyuan: Thanks. Fixed!

elasticdl/python/worker/worker.py
terrytangyuan: Good catch! This worker should be selected by master. I just moved this from `CollectiveCommunicator` to `Worker._get_broadcast_root_worker_ip()` and added TODO notes.

skydoorkai (approved):

PR #1503, opened 2019-11-22
Save checkpoints for training.

workingloong:
Implement #1466


elasticdl/python/ps/servicer.py
```diff
@@ -216,3 +223,24 @@ def wrap_optimizer_and_set_slot(self):
                 self._optimizer.slot_initial_value,
             )
             self._use_wrap_opt = True
+
+    def _increment_params_version(self):
+        self._parameters.version += 1
+        self._save_params_to_checkpoint_file()
+
+    def _save_params_to_checkpoint_file(self):
```
QiJune: Do we have a method which loads a checkpoint to PS?

elasticdl/python/master/checkpoint_service.py
```diff
@@ -60,16 +70,40 @@ def need_to_checkpoint(self, version):
...
                 while len(self._checkpoint_list) > self._max_versions:
                     file_to_delete = self._checkpoint_list.pop(0).file
                     os.remove(file_to_delete)
+                    # Remove the directory if empty
+                    delete_dir_name = os.path.dirname(file_to_delete)
+                    if not os.listdir(delete_dir_name):
+                        try:
+                            os.rmdir(delete_dir_name)
+                        except Exception:
+                            pass
```
QiJune, line 106: It seems that checkpoint is not a service, there are just some util functions. And these util functions could be used in PS under ParameterServer Strategy, or worker under Allreduce Strategy.

Maybe we could move it to some other place later.

elasticdl/python/ps/servicer.py
```diff
@@ -154,7 +161,7 @@ def push_gradient(self, request, _):
                     self._optimizer.apply_gradients(grad_vars)
                     self._grads_n = 0
                     self._grads_buffer.clear()
-                    self._parameters.version += 1
+                    self._increment_params_version()
```
QiJune: It seems that `_increment_params_version` is a very thin wrapper. And to make the code more readable, we could directly write:

```python
self._parameters.version += 1
if self._checkpoint_service and self._parameters.version % self._checkpoint_service._steps == 0:
    self._save_params_to_checkpoint()
```

elasticdl/python/ps/servicer.py
workingloong: I will implement the method to restore parameters from a checkpoint to PS instances in a future PR. I have created the issue #1506 to implement it.

elasticdl/python/master/checkpoint_service.py
workingloong: Agree with it! I will move it to the `common` directory later and I have created a issue #1506 .

QiJune (approved):
Excellent!

Issue #1508, opened 2019-11-22
Restore model parameters from a checkpoint and assign parameters to a Keras model.


workingloong:
The worker restores model parameters from the lastest checkpoint and converts embedding tables to dense variables. Then those restored variables will be assigned to trained weights of the Keras model.

PR #1511, opened 2019-11-25
Restore model from a checkpoint directory to export tensorflow SavedModel

workingloong:
Implement #1508.


elasticdl/python/common/args.py
```diff
@@ -593,6 +593,7 @@ def parse_ps_args(ps_args=None):
 def parse_worker_args(worker_args=None):
     parser = argparse.ArgumentParser(description="ElasticDL Worker")
     add_common_args_between_master_and_worker(parser)
+    add_train_params(parser)
```
skydoorkai: We can move checkpoint related args into common_args

skydoorkai (commented):
After we clean up the master code for the old ps code path, we can remove checkpoint_service so the code looks more straightforwad.

elasticdl/python/common/model_utils.py
```diff
@@ -181,3 +183,79 @@ def get_non_embedding_trainable_vars(model, embedding_layers):
         if not is_embedding_item:
             non_embedding_trainable_vars.append(var)
     return non_embedding_trainable_vars
+
+
+def restore_model_params_from_checkpoint(
+    checkpoint_dir, shard_index, shard_num
```
QiJune, line 192: Not sure why we need `shard_index` and `shard_num` here. It seems that this function should load all shards to restore a complete model.

workingloong: We need `shard_index` and `shard_num` when we restore parameters for a PS instance. I will implement in the next PR.

QiJune: I see. The checkpoint is used in two cases:

- export to standard Keras model, usually for serving. In this case, we assume the model size is not very large, and just set ps_num to 1.
- load to PS instance, usually for retraining. The PS instances number could also get changed.

skydoorkai (approved):

elasticdl/python/common/model_utils.py
workingloong: Yes!

PR #1516, opened 2019-11-25
Report gradients to local model for allreduce strategy

terrytangyuan:

### DIFF
--- a/elasticdl/python/worker/worker.py
+++ b/elasticdl/python/worker/worker.py
@@ -139,7 +139,7 @@ class Worker(object):
         self._get_model_steps = args.get_model_steps
         if self._get_model_steps > 1:
             self._opt = self._opt_fn()
-            self._non_embed_grads = None
+        self._non_embed_grads = {}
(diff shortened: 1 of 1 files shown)


elasticdl/python/worker/worker.py
```diff
@@ -423,9 +423,18 @@ def report_gradient_to_ps(self, grads):
         # TODO: choose the last response temporarily
         return res.accepted, res.model_version
 
-    # TODO: Reuse common code in report_gradient_to_ps and
-    # report_gradient_to_master
     def report_gradient_locally(self, grads):
+        # TODO: Automatically replace EDL embedding layer
```
brightcoder01: I think for AllReduce strategy, the ElasticDL embedding layer shouldn't exist. Raising exception is fine. The TODO is not necessary.

terrytangyuan: Thanks. Removed.

brightcoder01 (approved):

PR #1523, opened 2019-11-26
Create k8s services for worker-worker communication

terrytangyuan:
This would enable communication among workers for allreduce distribution strategy.

### DIFF
--- a/elasticdl/python/common/k8s_client.py
+++ b/elasticdl/python/common/k8s_client.py
@@ -14,6 +14,8 @@ ELASTICDL_APP_NAME = "elasticdl"
 ELASTICDL_JOB_KEY = "elasticdl-job-name"
 ELASTICDL_REPLICA_TYPE_KEY = "elasticdl-replica-type"
 ELASTICDL_REPLICA_INDEX_KEY = "elasticdl-replica-index"
+_PS_SERVICE_PORT = 2222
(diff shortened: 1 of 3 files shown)


elasticdl/python/common/k8s_client.py
```diff
@@ -175,6 +186,17 @@ def get_ps_service(self, ps_id):
             logger.warning("Exception when reading PS service: %s\n" % e)
             return None
 
+    def get_worker_service(self, worker_id):
+        try:
+            return self.client.read_namespaced_service(
+                # Worker service has the same name as pod name
+                name=self.get_worker_pod_name(worker_id),
```
ywskycn: Although it is the same name, here we can use `get_worker_service_name(worker_id)`, right?

terrytangyuan: Good call. Just updated the PR.

ywskycn (approved):

elasticdl/python/common/k8s_client.py
```diff
@@ -89,23 +91,32 @@ def _watch(self):
             except Exception:
                 traceback.print_exc()
 
+    def _get_service_address(self, service_name, port):
```
brightcoder01: How about static method?

PR #1526, opened 2019-11-27
Add ability for dumping YAML file of the ElasticDL job

ywskycn:
Another option to start the job. Not calling the k8s API, but dumping a YAML file and users can use the YMAL to start the job.


terrytangyuan (approved):

elasticdl/requirements.txt
```diff
@@ -3,4 +3,4 @@ kubernetes
 docker
 pyrecordio>=0.0.6
 odps
-redis-py-cluster
\ No newline at end of file
+redis-py-cluster
```
skydoorkai, line 6: Why redis-py-cluster is still needed?

ywskycn: oh, I think here I just add a new blank line there. `redis-py-cluster` is there before.. Not sure whether we have removed all redis code.

Issue #1527, opened 2019-11-27
Add a sparse embedding layer


mhaoli:
This embedding layer should inherit `tf.keras.layers.Layer`.

PR #1530, opened 2019-11-27
Implement a sparse embedding layer using TF embedding_lookup_sparse

workingloong:
Implement #1527

### DIFF
--- a/elasticdl/python/common/model_handler.py
+++ b/elasticdl/python/common/model_handler.py
@@ -26,7 +26,7 @@ def _get_trained_params_from_checkpoint(checkpoint_dir):
     return trained_params
 
 
-def _convert_embedding_table_to_array(embedding_table, embedding_shape):
+def _convert_embedding_table_to_numpy_array(embedding_table, embedding_shape):
     """Convert an embedding table to a np.ndarray which can be assigned
     to trainable weights in keras embedding layers.
 
@@ -161,7 +161,7 @@ class ParameterServerModelHandler(ModelHandler):
         trained_params = _get_trained_params_from_checkpoint(checkpoint_dir)
         for var in model.trainable_variables:
             if isinstance(trained_params[var.name], EmbeddingTable):
-                embedding_params = _convert_embedding_table_to_array(
+                embedding_params = _convert_embedding_table_to_numpy_array(
                     trained_params[var.name], var.shape
                 )
                 var.assign(embedding_params)
@@ -227,7 +227,9 @@ class ParameterServerModelHandler(ModelHandler):
 
         def _clone_function(layer):
             if type(layer) == Embedding:
-                # ElasticDL embedding only accept a string type initializer
+                logger.info("Replace embedding layer with "
+                            "elasticdl.layers.Embedding")
+                # The combiner is not None only for SparseEmbedding,
                 if layer.combiner is not None:
                     embedding_layer = SparseEmbedding(
                         output_dim=layer.output_dim,
@@ -261,13 +263,15 @@ class ParameterServerModelHandler(ModelHandler):
         """
         for name, value in model.__dict__.items():
             if type(value) in [tf.keras.layers.Embedding, SparseEmbedding]:
-                init = tf.keras.initializers.serialize(
+                logger.info("Replace embedding layer with "
+                            "elasticdl.layers.Embedding")
+                initializer_name = tf.keras.initializers.serialize(
                     value.embeddings_initializer
                 )["class_name"]
                 embedding_layer = Embedding(
                     output_dim=value.output_dim,
                     input_dim=value.input_dim,
-                    embeddings_initializer=init,
+                    embeddings_initializer=initializer_name,
                     mask_zero=value.mask_zero,
                     input_length=value.input_length,
                 )
@@ -283,7 +287,9 @@ class ParameterServerModelHandler(ModelHandler):
         """
         for name, value in model.__dict__.items():
             if type(value) == Embedding:
+                # The combiner is not None only for SparseEmbedding,
                 if value.combiner is not None:
+                    logger.info("Replace elasticdl with SparseEmbedding")
                     embedding_layer = SparseEmbedding(
                         output_dim=value.output_dim,
                         input_dim=value.input_dim,
@@ -293,6 +299,9 @@ class ParameterServerModelHandler(ModelHandler):
                         combiner=value.combiner,
                     )
                 else:
+                    logger.info("Replace elasticdl with ",
+                                "tf.kerasl.layers.Embedding"
+                                )
                     embedding_layer = tf.keras.layers.Embedding(
                         output_dim=value.output_dim,
                         input_dim=value.input_dim,
--- a/elasticdl/python/common/model_handler.py
+++ b/elasticdl/python/common/model_handler.py
@@ -202,16 +202,23 @@ class ParameterServerModelHandler(ModelHandler):
                     layer.embeddings_initializer
                 )["class_name"]
 
-                embedding_layer = Embedding(
-                    output_dim=layer.output_dim,
-                    input_dim=layer.input_dim,
-                    embeddings_initializer=init,
-                    mask_zero=layer.mask_zero,
-                    input_length=layer.input_length,
-                    name=layer.name,
-                )
-                if type(layer) == SparseEmbedding:
-                    embedding_layer.combiner = layer.combiner
+                if type(layer) == tf.keras.layers.Embedding:
+                    embedding_layer = Embedding(
+                        output_dim=layer.output_dim,
+                        input_dim=layer.input_dim,
+                        embeddings_initializer=init,
+                        mask_zero=layer.mask_zero,
+                        input_length=layer.input_length,
+                        name=layer.name,
+                    )
+                else:
+                    embedding_layer = Embedding(
+                        output_dim=layer.output_dim,
+                        input_dim=layer.input_dim,
+                        embeddings_initializer=init,
+                        name=layer.name,
+                        combiner=layer.combiner,
+                    )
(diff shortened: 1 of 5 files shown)


elasticdl/python/keras/layers/sparse_embedding.py
```diff
@@ -0,0 +1,84 @@
+import tensorflow as tf
+from tensorflow.python.keras import backend as K
+from tensorflow.python.keras import initializers
+from tensorflow.python.ops import embedding_ops, math_ops
+
+
+class SparseEmbedding(tf.keras.layers.Layer):
```
QiJune, line 7: It seems that `tf.keras.layers.Embedding` does support SparseTensor as an input. So we provide a custom `SparseEmbedding` layer to users? As a result, users have to import SparseEmbedding to build their model.

elasticdl/python/common/model_handler.py
```diff
@@ -153,8 +161,8 @@ def get_model_to_export(self, model, dataset):
         trained_params = _get_trained_params_from_checkpoint(checkpoint_dir)
         for var in model.trainable_variables:
             if isinstance(trained_params[var.name], EmbeddingTable):
-                embedding_params = _convert_embedding_vector_to_variable(
-                    var.shape, trained_params[var.name]
+                embedding_params = _convert_embedding_table_to_array(
```
brightcoder01: _convert_embedding_table_to_array -> _convert_embedding_table_to_numpy_array

elasticdl/python/keras/layers/sparse_embedding.py
workingloong: No, `tf.keras.layers.Embedding` does not support SparseTensor as an input. I have created an issue on `Tensorflow` about embedding with SparseTensor. But I don't receive any feedback. The `SparseEmbedding` is a Keras-native layer, so the user can use it with SparseTensor in any platform. In another way, we can use `tf.feature_column.embedding_column` for `SparseTensor` with variable-length data which can not use the partitioned embedding table with multiple PS instances in ElasticDL. When the custom `embedding_column` is ready in #1535, we can use it to replace `tf.feature_column.embedding_column`.

elasticdl/python/common/model_handler.py
```diff
@@ -221,20 +220,85 @@ def _clone_function(layer):
...
-        elasticdl.layers.Embedding layer.
+    def _clone_model_with_keras_embedding(model):
+        """Clone a new model and replace the `elasticdl.layers.Embedding`
+        layers with `tf.keras.layers.Embedding` or `SparseEmbedding` layers
+        """
+
+        def _clone_function(layer):
+            if type(layer) == Embedding:
+                # ElasticDL embedding only accept a string type initializer
+                if layer.combiner is not None:
```
brightcoder01, line 242: Add comments here:
Only for the scenario of SparseEmbedding, the combiner is not None

elasticdl/python/common/model_handler.py
```diff
@@ -221,20 +220,85 @@ def _clone_function(layer):
...
+    @staticmethod
+    def _replace_attr_with_edl_embedding(model):
+        """Replace the keras embedding attributes in the model with
+        `elasticdl.layers.Embedding` layers.
         """
         for name, value in model.__dict__.items():
-            if type(value) == src_embedding_class:
-                embedding_layer = dst_embedding_class(
+            if type(value) in [tf.keras.layers.Embedding, SparseEmbedding]:
+                init = tf.keras.initializers.serialize(
```
brightcoder01: init -> initializer_name

elasticdl/python/keras/layers/sparse_embedding.py
```diff
@@ -0,0 +1,84 @@
...
+            trainable=True,
+        )
+        self.built = True
+
+    def call(self, inputs):
+        if self.combiner not in ["sum", "mean", "sqrtn"]:
+            raise ValueError(
+                "combiner must set sum, mean or sqrtn for sparse input"
+            )
+        if not isinstance(inputs, tf.SparseTensor):
```
brightcoder01, line 56: Please add comments here why we do this conversion.

elasticdl/python/common/model_handler.py
```diff
@@ -7,10 +7,12 @@
 from elasticdl.python.common.log_utils import default_logger as logger
 from elasticdl.python.common.save_utils import CheckpointSaver
 from elasticdl.python.elasticdl.layers.embedding import Embedding
+from elasticdl.python.keras.layers import SparseEmbedding
```
chunyang-wen, line 10: The difference between `elasticdl.python.elasticdl.layers` and  `elasticdl.python.keras.layers `?

elasticdl/python/keras/layers/sparse_embedding.py
```diff
@@ -0,0 +1,84 @@
...
+            shape=(self.input_dim, self.output_dim),
+            initializer=self.embeddings_initializer,
+            name="embeddings",
+            trainable=True,
+        )
+        self.built = True
+
+    def call(self, inputs):
+        if self.combiner not in ["sum", "mean", "sqrtn"]:
+            raise ValueError(
```
chunyang-wen: Move this checking code to the constructor in order to fail fast.

elasticdl/python/keras/layers/sparse_embedding.py
```diff
@@ -0,0 +1,84 @@
...
+                "combiner must set sum, mean or sqrtn for sparse input"
+            )
+        if not isinstance(inputs, tf.SparseTensor):
+            idx = tf.where(tf.not_equal(inputs, 0))
+            inputs = tf.SparseTensor(
+                idx, tf.gather_nd(inputs, idx), (-1, self.input_dim)
+            )
+
+        dtype = K.dtype(inputs)
+        if dtype != "int32" and dtype != "int64":
```
chunyang-wen, line 63: There are a couple of magic words here.  "int32" and "int64" are defined somewhere in keras?

elasticdl/python/common/model_handler.py
```diff
@@ -221,20 +220,85 @@ def _clone_function(layer):
...
+                setattr(model, name, embedding_layer)
+        return model
+
+    @staticmethod
+    def _replace_attr_with_keras_embedding(model):
+        """Replace the elasticdl.layers.Embedding attributes in the model
+        with `tf.keras.layers.Embedding` or `SparseEmbedding` layers.
+        """
+        for name, value in model.__dict__.items():
+            if type(value) == Embedding:
```
chunyang-wen, line 308: Please add logs here.

elasticdl/python/common/model_handler.py
workingloong: I create a `elasticdl.python.keras.layers` directory to place the layer to show that the layer is Keras-native and can be used by users to define their models. I think the layers in `difference between elasticdl.python.elasticdl.layers ` will not be exposed.

elasticdl/python/common/model_handler.py
workingloong: `logger.info()` ?

elasticdl/python/keras/layers/sparse_embedding.py
workingloong: Yeah!

elasticdl/python/keras/layers/sparse_embedding.py
workingloong: https://github.com/tensorflow/tensorflow/blob/1cf0898dd4331baf93fe77205550f2c2e6c90ee5/tensorflow/python/keras/layers/embeddings.py#L179-L184

elasticdl/python/keras/layers/sparse_embedding.py
```diff
@@ -0,0 +1,88 @@
...
+            inputs = math_ops.cast(inputs, "int32")
+        out = embedding_ops.safe_embedding_lookup_sparse(
```
brightcoder01, line 65: Nice for safe_embedding_lookup_sparse

elasticdl/python/keras/layers/sparse_embedding.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+class SparseEmbedding(tf.keras.layers.Layer):
+    """
+    Input: indexes for the embedding entries with a shape of
+      (batch_size, input_length). Input can be either dense tensor
+      or SparseTensor.
+    Output:
+      corresponding (combined) embeddings with a shape of
+      (batch_size, input_length, output_dim) if combiner is None
```
skydoorkai: combiner cannot be None for sparse embedding.

workingloong: May we set a default value like "mean" ? The default value of`safe_embedding_lookup_sparse`  is "mean" .

brightcoder01: Yes, the default value is mean

elasticdl/python/tests/model_handler_test.py
```diff
@@ -161,6 +173,23 @@ def _get_dataset():
...
+        model_inst = custom_model_with_sparse_embedding()
+        train_model = self.model_handler.get_model_to_train(model_inst)
+        export_model = self.model_handler.get_model_to_export(
+            train_model, dataset=None
+        )
+        test_data = tf.SparseTensor(
+            indices=[[0, 0]], values=[0], dense_shape=(1, 1)
+        )
+        result = export_model.call(test_data).numpy()
+        self.assertEqual(result[0][0], 3.0)
```
brightcoder01, line 198: Please add comments in the test case why the result value is calculated from the parameter value loaded from checkpoint. And the parameter values in the checkpoint can be mentioned.

brightcoder01 (approved):

PR #1535, opened 2019-11-28
Initial version of elasticdl customized embedding column

brightcoder01:
TensorFlow Native embedding_column contains a embedding variable to store the embedding table. It can't leverage the embedding table partition from ElasticDL parameter server.
We choose to customized an elasticdl embedding column. Its signature is the same with the native embedding column.

### DIFF
--- a/elasticdl/python/elasticdl/feature_column/feature_column.py
+++ b/elasticdl/python/elasticdl/feature_column/feature_column.py
@@ -1,10 +1,10 @@
 import collections
 import math
-import numpy as np
+
 import tensorflow as tf
 from tensorflow.python.feature_column import feature_column as fc_old
 from tensorflow.python.feature_column import feature_column_v2 as fc_lib
-from tensorflow.python.framework import ops, tensor_shape
+from tensorflow.python.framework import tensor_shape
 from tensorflow.python.ops import init_ops
 
 
@@ -101,7 +101,8 @@ class EmbeddingColumn(
         ):
             raise ValueError(
                 "In embedding_column: {}. "
-                "categorical_column must not be of type SequenceCategoricalColumn. "
+                "categorical_column must not be of "
+                "type SequenceCategoricalColumn. "
                 "Suggested fix A: If you wish to use DenseFeatures, use a "
                 "non-sequence categorical_column_with_*. "
                 "Suggested fix B: If you wish to create sequence input, use "
--- a/elasticdl/python/elasticdl/feature_column/feature_column.py
+++ b/elasticdl/python/elasticdl/feature_column/feature_column.py
@@ -1,6 +1,5 @@
 import collections
 import math
-
 import numpy as np
 import tensorflow as tf
 from tensorflow.python.feature_column import feature_column as fc_old
--- a/elasticdl/python/elasticdl/feature_column/feature_column.py
+++ b/elasticdl/python/elasticdl/feature_column/feature_column.py
@@ -13,18 +13,48 @@ def embedding_column(
     dimension,
     combiner="mean",
     initializer=None,
-    ckpt_to_load_from=None,
-    tensor_name_in_ckpt=None,
     max_norm=None,
     trainable=True,
 ):
+    """
+    Create a customized EmbeddingColumn for ElasticDL.
+    The native EmbeddingColumn will create a variable to
+    store the entire embedding table. It can't leverage the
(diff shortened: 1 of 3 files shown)


elasticdl/python/elasticdl/feature_column/feature_column.py
```diff
@@ -0,0 +1,156 @@
...
+def embedding_column(
+    categorical_column,
+    dimension,
+    combiner="mean",
+    initializer=None,
+    ckpt_to_load_from=None,
+    tensor_name_in_ckpt=None,
+    max_norm=None,
+    trainable=True,
+):
```
workingloong, line 18: Add a docstring to explain those args.

elasticdl/python/elasticdl/feature_column/feature_column.py
```diff
@@ -0,0 +1,156 @@
...
+    def parse_example_spec(self):
+        """See `FeatureColumn` base class."""
+        return self.categorical_column.parse_example_spec
+
+    @property
+    def variable_shape(self):
+        """See `DenseColumn` base class."""
+        return tensor_shape.TensorShape([self.dimension])
+
+    def create_state(self, state_manager):
```
workingloong: What's the use of the function?

elasticdl/python/elasticdl/feature_column/feature_column.py
```diff
@@ -0,0 +1,156 @@
...
+                batch_embedding, idx, segment_ids
+            )
+        elif self.combiner == "sqrtn":
+            batch_embedding = tf.sparse.segment_sqrt_n(
+                batch_embedding, idx, segment_ids
+            )
+
+        return batch_embedding
+
+    def lookup_embedding(self, unique_ids):
```
workingloong, line 172: Can we use the `lookup_embedding` in `ElasticDL.layers.Embedding` ?

workingloong:
How can we train the weights in the custom embedding columns?

elasticdl/python/elasticdl/feature_column/feature_column.py
brightcoder01: This function is just like the layer.build call. For native embedding column, it's used to create the embedding variables. But for the customized embedding column, we don't need create this variable. So we set the implementation to be empty.

elasticdl/python/elasticdl/feature_column/feature_column.py
brightcoder01: Yes, we will reuse the same lookup_embedding in ElasticDL.layers.Embedding. The refactor and code reuse will be added in the next PRs.

brightcoder01:
> How can we train the weights in the custom embedding columns?

We will call tape.watch(batch_embedding) in get_dense_tensor

elasticdl/python/elasticdl/feature_column/feature_column.py
brightcoder01: Good idea, will add it.

workingloong:
> > How can we train the weights in the custom embedding columns?
> 
> We will call tape.watch(batch_embedding) in get_dense_tensor

OK. I wonder whether we can abstract a customized embedding_op from `elasticdl.layers.Embedding`. Then, we can implement `Embedding` layer and feature column by calling the embedding_op like Tensorflow.

brightcoder01:
> OK. I wonder whether we can abstract a customized embedding_op from `elasticdl.layers.Embedding`. Then, we can implement `Embedding` layer and feature column by calling the embedding_op like Tensorflow.

Good idea, we can have a try.

elasticdl/python/elasticdl/feature_column/feature_column.py
```diff
@@ -0,0 +1,179 @@
...
+    combiner="mean",
+    initializer=None,
+    max_norm=None,
+    trainable=True,
+):
+    """
+    Create a customized EmbeddingColumn for ElasticDL.
+    The native EmbeddingColumn will create a variable to
+    store the entire embedding table. It can't leverage the
+    benifit from the ElasticDL parameter server to partition
```
skydoorkai: benifit -> benefit

workingloong (approved):

elasticdl/python/elasticdl/feature_column/feature_column.py
brightcoder01: Good catch, updated.

PR #1543, opened 2019-12-02
Dynamically update service addresses after worker/ps relaunch

terrytangyuan:
This solves the following issues:
* Currently the ps pod being relaunched uses the same id, which is problematic.
* The worker and ps addresses are not updated when a pod relaunches. Under allreduce strategy, we need to know the updated list of service addresses.

### DIFF
--- a/elasticdl/python/master/k8s_instance_manager.py
+++ b/elasticdl/python/master/k8s_instance_manager.py
@@ -49,8 +49,8 @@ class InstanceManager(object):
         self._image_pull_policy = image_pull_policy
         self._envs = envs
         self._task_d = task_d
-        self._next_worker_id = itertools.count(num_workers).__next__
-        self._next_ps_id = itertools.count(num_ps).__next__
+        self._next_worker_id = itertools.count().__next__
+        self._next_ps_id = itertools.count().__next__
 
         # Protects followed variables, which are accessed from event_cb.
         self._lock = threading.Lock()
@@ -151,8 +151,8 @@ class InstanceManager(object):
             self._start_worker(self._next_worker_id())
 
     def start_all_ps(self):
-        for i in range(self._num_ps):
-            self._start_ps(i)
+        for _ in range(self._num_ps):
+            self._start_ps(self._next_worker_id())
 
     def _remove_worker(self, worker_id):
         logger.info("Removing worker: %d", worker_id)
--- a/elasticdl/python/master/k8s_instance_manager.py
+++ b/elasticdl/python/master/k8s_instance_manager.py
@@ -150,9 +150,9 @@ class InstanceManager(object):
         for _ in range(self._num_workers):
             self._start_worker(self._next_worker_id())
 
-    def start_all_ps(self):
+    def start_parameter_servers(self):
         for _ in range(self._num_ps):
-            self._start_ps(self._next_worker_id())
+            self._start_ps(self._next_ps_id())
 
     def _remove_worker(self, worker_id):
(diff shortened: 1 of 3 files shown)


elasticdl/python/master/k8s_instance_manager.py
```diff
@@ -71,10 +74,14 @@ def __init__(
         self._relaunch_deleted_live_ps = True
 
         self._k8s_client = k8s.Client(event_callback=self._event_cb, **kwargs)
-        self._ps_addrs = self._get_ps_addrs()
+        self._ps_addrs = self._get_addrs(
```
brightcoder01, line 77: Why not store _ps_addrs as a string array instead of a string concatenated with separator? Or restore as a dictionary of id -> addr_string.

elasticdl/python/master/k8s_instance_manager.py
```diff
@@ -232,15 +240,29 @@ def _event_cb(self, event):
...
+                new_worker_id,
+                self._worker_addrs,
+                addr_get_fn=self._k8s_client.get_worker_service_address,
+            )
+        elif relaunch_ps and ps_id:
             logger.info("Relaunching ps.")
-            self._start_ps(ps_id)
+            new_ps_id = self._next_ps_id()
+            self._start_ps(new_ps_id)
+            self._update_addr(
```
brightcoder01, line 260: _update_addr is a static method. We can InstanceManager._update_addr instead of self._update_addr

brightcoder01:
> * Currently the ps pod being relaunched uses the same id, which is problematic.
Can you share more information about what problem it will bring to use the same id?

elasticdl/python/master/k8s_instance_manager.py
terrytangyuan: It's only used inside this class now and is marked as internal so I'd like to keep it as `self._update_addr`.

elasticdl/python/master/k8s_instance_manager.py
terrytangyuan: Because we need to pass this as a string to `worker.main`.

terrytangyuan:
> > * Currently the ps pod being relaunched uses the same id, which is problematic.
> >   Can you share more information about what problem it will bring to use the same id?

@brightcoder01 Because the pod may not have been deleted or successfully deleted yet. It's easier to differentiate the existing pods and the newly created pods if we use different IDs.

brightcoder01:
> > > * Currently the ps pod being relaunched uses the same id, which is problematic.
> > >   Can you share more information about what problem it will bring to use the same id?
> 
> @brightcoder01 Because the pod may not have been deleted or successfully deleted yet. It's easier to differentiate the existing pods and the newly created pods if we use different IDs.

I'm not sure whether the fault tolerance of PS will rely on the unchanged PS id. @skydoorkai could you please take a look at it?

skydoorkai:
> > > > * Currently the ps pod being relaunched uses the same id, which is problematic.
> > > >   Can you share more information about what problem it will bring to use the same id?
> > 
> > 
> > @brightcoder01 Because the pod may not have been deleted or successfully deleted yet. It's easier to differentiate the existing pods and the newly created pods if we use different IDs.
> 
> I'm not sure whether the fault tolerance of PS will rely on the unchanged PS id. @skydoorkai could you please take a look at it?

As long as we can use the same service domain name to access the newly launched PS, it is fine with PS fault tolerance.

brightcoder01 (approved):

PR #1544, opened 2019-12-02
Improve code style in TaskDataService

terrytangyuan:

### DIFF
--- a/elasticdl/python/worker/task_data_service.py
+++ b/elasticdl/python/worker/task_data_service.py
@@ -96,9 +96,6 @@ class TaskDataService(object):
                     self._reported_record_count -= task.end - task.start
                     self._pending_tasks.popleft()
                     self._do_report_task(task, err_msg)
-                    # Since a single batch comes from multiple tasks,
-                    # we just report the number of failing records together
-                    # with the first task
                     self._failed_record_count = 0
                 if self._pending_tasks:
                     self._current_task = self._pending_tasks[0]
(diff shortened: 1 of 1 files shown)


elasticdl/python/worker/task_data_service.py
```diff
@@ -94,8 +96,8 @@ def report_record_done(self, count, err_msg=""):
                     self._reported_record_count -= task.end - task.start
                     self._pending_tasks.popleft()
                     self._do_report_task(task, err_msg)
-                    # Becuase a single batch comes from multiple tasks,
-                    # We just report the number of failing records together
+                    # Since a single batch comes from multiple tasks,
```
brightcoder01: IMO, in current implementation, a single batch only comes from a single task. We have a parameter: num_minibatches_per_task.

elasticdl/python/worker/task_data_service.py
```diff
@@ -83,9 +85,9 @@ def report_record_done(self, count, err_msg=""):
             if err_msg:
                 self._log_fail_records(task, err_msg)
 
-            # Keep poping: batch_size may be larger than
-            # task.end - task.start, so we keep poping task until
-            # reported count is less than current data size
+            # Keep popping tasks until the reported record count is less
```
brightcoder01, line 88: I think you want to mean "Keep popping tasks until the reported record count exceeds the size of the current data"?

terrytangyuan: No that's not what I meant since the code is `self._reported_record_count >= (
                    self._pending_tasks[0].end - self._pending_tasks[0].start
                )`

elasticdl/python/worker/task_data_service.py
terrytangyuan: Agreed. Just removed this comment. The original purpose of this PR was only for improving code style and fixing English grammar in code comments.

brightcoder01 (approved):

PR #1545, opened 2019-12-03
expose data partition args in env

QiJune:


skydoorkai:
This does not work, as our partition looks like "dt=20191126". Then envs should be set as `PARTITION=dt=20191126`. This will create invalid env as current code will get confused with two `=`.

PR #1551, opened 2019-12-04
Add logic to broadcast model params to all workers 

terrytangyuan:


elasticdl/python/worker/worker.py
```diff
@@ -547,31 +548,48 @@ def forward_process(self, features):
...
+            self._get_local_model_params()
+            if is_broadcast_root_worker
+            else None
         )
-        if status == CollectiveCommunicatorStatus.FAILED:
-            logger.warning("Failed to broadcast model parameters")
-            return False
-        self._update_local_model_params(model_params)
+
+        if not is_broadcast_root_worker and model_params is not None:
```
skydoorkai, line 590: If the current worker is_broadcast_root_worker,  it also needs to broadcast its model_params to other workers.

terrytangyuan: Good catch! This should be fixed now.

skydoorkai (approved):

PR #1559, opened 2019-12-08
Implement the training process using ElasticDL EmbeddingColumn.

brightcoder01:
- Implement the training process of the model with ElasticDL embedding column: Calculate and report the gradients of the embedding values of ElasticDL embedding column.

- Extract a common component `EmbeddingDelegate` to interact with the external EmbeddingStorage such as Parameter Server. It's shared by ElasticDL embedding layer and embedding column.

- Currently we don't update the EmbeddingLayer with the EmbeddingDelegate for the memory leak investigation.

### DIFF
--- a/elasticdl/python/worker/worker.py
+++ b/elasticdl/python/worker/worker.py
@@ -353,7 +353,7 @@ class Worker(object):
                 embedding_info = embedding_infos.add()
                 embedding_info.name = column.name
                 embedding_info.dim = column.dimension
-                # TODO(brightcoder): The initializer in embedding column is
+                # TODO(brightcoder01): The initializer in embedding column is
                 # a variable initializer function. For embedding layer, it's a
                 # tf.keras.initializers. Keep aligned between these two.
                 embedding_info.initializer = "uniform"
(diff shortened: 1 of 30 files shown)


elasticdl/python/worker/worker.py
```diff
@@ -335,6 +376,17 @@ def report_variable(self):
         for ps_id in range(len(self._ps_stubs)):
             self.report_variable_to_ps(ps_id)
 
+    def collect_edl_embedding_name_values(self):
```
workingloong: Add a docstring to explain the output.

elasticdl/python/worker/worker.py
```diff
@@ -335,6 +376,17 @@ def report_variable(self):
...
+        embedding_name_values = []
+        for layer in self._embedding_layers:
+            embedding_name_values.append((layer.name, layer.embedding_and_ids))
+        for column in self._embedding_columns:
+            embedding_name_values.append(
+                (column.name, column.embedding_and_ids)
+            )
```
workingloong, line 417: May `layer.name` be the same as `column.name` ?

workingloong:
Add a unit test for ElasticDL embedding_column.

elasticdl/python/worker/worker.py
brightcoder01: Good suggestion, added.

elasticdl/python/worker/worker.py
brightcoder01: The name format for column.name is {column_field}_embedding_elasticdl, the layer name is user defined. I think the probability is low for name conflict.

brightcoder01:
> Add a unit test for ElasticDL embedding_column.

Good suggestion, add more test cases in feature_column_test.py to cover the gradients calculation.

elasticdl/python/worker/worker.py
workingloong: Whether do we need to check there are name conflicts? If conflicts exist, we can raise an error or warning.

brightcoder01: Good suggestion, added the check logic

workingloong (approved):

PR #1563, opened 2019-12-13
Relaunch failed ps/worker pod due to OOM from eager mode memory leak

### LINKED ISSUE #1562: Master needs to relaunch a failed ps/worker pod for OOM workaround
TensorFlow 2.0 has memory leak issues in the eager mode. A ps/worker pod may fail due to OOM after running for a while.

To work around it, the master can delete the failed pod so it can be relaunched.

We can remove this workaround after all memory leak issues are resolved.

skydoorkai:
This is a workaround for the memory leak so the training can continue after OOM.
We can remove it after all memory leak issues are resolved.

Fix #1562

### DIFF
--- a/elasticdl/python/master/k8s_instance_manager.py
+++ b/elasticdl/python/master/k8s_instance_manager.py
@@ -73,6 +73,8 @@ class InstanceManager(object):
         self._ps_pod_name_to_id = {}
         self._relaunch_deleted_live_ps = True
 
+        self._failed_pods = []
+
         self._k8s_client = k8s.Client(event_callback=self._event_cb, **kwargs)
         self._ps_addrs = self._get_addrs(
             self._num_ps, self._k8s_client.get_ps_service_address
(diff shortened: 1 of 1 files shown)


elasticdl/python/master/k8s_instance_manager.py
```diff
@@ -218,10 +220,18 @@ def _event_cb(self, event):
...
                 worker_id = self._worker_pod_name_to_id.get(pod_name)
                 self._worker_pods_phase[worker_id] = (pod_name, phase)
-                if evt_type == "DELETED":
+                # Workaround for memory leak issues in tf eager mode.
+                # A pod may fail due to OOM from tf eager mode memory leak.
+                failed_pod = False
+                if evt_type == "MODIFIED" and phase == "Failed":
+                    self._failed_pods.append(pod_name)
+                    failed_pod = True
+                if evt_type == "DELETED" or failed_pod:
```
brightcoder01, line 234: Need we add a counter for each failed pod? In order to avoid the infinite recover for other causes.

skydoorkai: When a pod fails, it is moved into `self._failed_pods.append` and a new pod will be launched to substitute it with a different pod name (with the same service). Each pod will only fail once.

Also, there is no way to know the cause of failure from pod status.

terrytangyuan (approved):

PR #1564, opened 2019-12-15
Add dnn model with feature columns using the census data in the model zoo.

brightcoder01:
- Add this model and recordio_gen to verify the training process using ElasticDL embedding column. The verification result is passed and we can get the expected test accuracy value.
- Write this sample model with using Sequential/Functional/Subclass way. In the next step, we will summarize the general pattern/standard of the model definition.

### DIFF
--- a/elasticdl/python/data/recordio_gen/census_recordio_gen.py
+++ b/elasticdl/python/data/recordio_gen/census_recordio_gen.py
@@ -9,9 +9,13 @@ import recordio
 import tensorflow as tf
 from sklearn.model_selection import train_test_split
 
-URL = (
+TRAIN_DATA_URL = (
     "https://archive.ics.uci.edu/ml/machine-learning-databases/adult/"
     "adult.data"
+)
+TEST_DATA_URL = (
+    "https://archive.ics.uci.edu/ml/machine-learning-databases/adult/"
+    "adult.test"
 )
 
 __COLUMN_NAMES = [
@@ -63,6 +67,9 @@ def convert_series_to_tf_feature(data_series, columns, dtype_series):
     Return:
         A dict of feature name -> tf.train.Feature
     """
+    if data_series.hasnans:
+        return
+
     features = {}
     for numeric_feature_key in NUMERIC_FEATURE_KEYS:
         feature = tf.train.Feature(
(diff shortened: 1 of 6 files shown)


skydoorkai:
Any unit test to put into `python/tests/example_test.py`?

brightcoder01:
> python

Good suggestion.
The downloading of the source data from https://archive.ics.uci.edu/ml/machine-learning-databases/adult/adult.data is not stable. If we add the census_recordio_gen.py into the Dockerfile.dev, the image build may hang for some time when the network is unstable. So I choose not to put the verification using census dataset into the unit test.
I'll add a test case using the small heart dataset instead to cover the ElasticDL embedding column.

skydoorkai:
> > python
> 
> Good suggestion.
> The downloading of the source data from https://archive.ics.uci.edu/ml/machine-learning-databases/adult/adult.data is not stable. If we add the census_recordio_gen.py into the Dockerfile.dev, the image build may hang for some time when the network is unstable. So I choose not to put the verification using census dataset into the unit test.
> I'll add a test case using the small heart dataset instead to cover the ElasticDL embedding column.

We can generate a small dataset with random data in  `create_recordio_file` (test_utils.py), similar to other tests.

model_zoo/census_dnn_model/census_feature_columns.py
```diff
@@ -0,0 +1,56 @@
+import tensorflow as tf
+
+from elasticdl.python.elasticdl.feature_column import feature_column
```
workingloong, line 3: The feature_column.py only contains `embedding_column`. Maybe it better to rename feature_column.py to embedding_column.py?

elasticdl/python/data/recordio_gen/census_recordio_gen.py
```diff
@@ -0,0 +1,186 @@
...
+
+    feature = tf.train.Feature(
+        int64_list=tf.train.Int64List(value=[data_series[LABEL_KEY]])
+    )
+    features[LABEL_KEY] = feature
+
+    return features
+
+
+def convert_to_recordio_files(data_frame, dir_name, records_per_shard):
```
terrytangyuan, line 94: Maybe reuse `data.odps_recordio_conversion_utils.write_recordio_shards_from_iterator()`?

skydoorkai (approved):

Issue #1567, opened 2019-12-16
Memory leak using tf.py_function in customized embedding layers with eager execution


workingloong:
https://github.com/sql-machine-learning/elasticdl/blob/020c928ead6e6e24d33833f25d5579adc5e05687/elasticdl/python/elasticdl/layers/embedding.py#L176-L178

The memory leak bug about `tf.py_function`  has been submitted to TensorFlow [issue 35010](https://github.com/tensorflow/tensorflow/issues/35010)

PR #1571, opened 2019-12-16
Don't use tf.py_function to wrap lookup_embedding with eager execution

workingloong:
Avoid #1567.

### DIFF
--- a/elasticdl/python/elasticdl/layers/embedding.py
+++ b/elasticdl/python/elasticdl/layers/embedding.py
@@ -1,5 +1,6 @@
 import numpy as np
 import tensorflow as tf
+from tensorflow.python.framework import ops
 from tensorflow.python.keras.utils import tf_utils
 
 from elasticdl.python.elasticdl.embedding_delegate import EmbeddingAndIds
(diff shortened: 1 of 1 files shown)


elasticdl/python/elasticdl/layers/embedding.py
```diff
@@ -170,12 +176,23 @@ def call(self, input):
         ids = tf.convert_to_tensor(input, name="embedding_ids")
         flat_ids = tf.reshape(ids, [-1])
         unique_ids, idx = tf.unique(flat_ids)
+
+        # There is a memory leak using `tf.py_function` with eager execution.
+        # So, we directly call `lookup_embedding` when inputs are EagerTensor.
```
skydoorkai: There is a memory leak when using `tf.py_function` with eager execution.
So, we only use `tf.py_function` in graph mode.

The comment below can be deleted.

elasticdl/python/elasticdl/layers/embedding.py
```diff
@@ -138,6 +139,11 @@ def _check_id_valid(self, ids):
                 "than input_dim. id = %d is not in [0, %d)"
                 % (first_may_exceed_id, self.input_dim)
             )
+        if any(ids < 0):
+            raise ValueError(
+                "The embedding id cannot less than zero "
```
skydoorkai: The embedding id cannot be less than zero
"Found invalid id %d " % ...

skydoorkai (approved):

PR #1574, opened 2019-12-16
Only relaunch failed pod with reason as OOMKilled

skydoorkai:


elasticdl/python/master/k8s_instance_manager.py
```diff
@@ -225,9 +225,19 @@ def _event_cb(self, event):
...
-            if evt_type == "MODIFIED" and phase == "Failed":
+            if (
+                evt_type == "MODIFIED"
+                and phase == "Failed"
+                and evt_obj.status.container_statuses
+                and evt_obj.status.container_statuses[0].state.terminated
+                and evt_obj.status.container_statuses[
+                    0
+                ].state.terminated.reason
+                == "OOMKilled"
```
terrytangyuan, line 236: Why only this reason?

skydoorkai: If other reasons, there must be bugs in the code/model, and no need to relaunch.

terrytangyuan (approved):

PR #1576, opened 2019-12-18
Use warning when id < 0

workingloong:
The id < 0 may be used as a default value, so we should raise a warning not an error.

### DIFF
--- a/elasticdl/python/elasticdl/layers/embedding.py
+++ b/elasticdl/python/elasticdl/layers/embedding.py
@@ -3,6 +3,7 @@ import tensorflow as tf
 from tensorflow.python.framework import ops
 from tensorflow.python.keras.utils import tf_utils
 
+from elasticdl.python.common.log_utils import default_logger as logger
 from elasticdl.python.elasticdl.embedding_delegate import EmbeddingAndIds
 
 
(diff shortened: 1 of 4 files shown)


elasticdl/python/elasticdl/layers/embedding.py
```diff
@@ -140,7 +141,7 @@ def _check_id_valid(self, ids):
                 % (first_may_exceed_id, self.input_dim)
             )
         if any(ids < 0):
-            raise ValueError(
+            logger.warning(
```
chunyang-wen: It's better that we can make it configurable.

workingloong: But, how and where to config it? `tf.nn.safe_embedding_lookup_sparse` auto prunes invalid IDs ( < 0) and `tf.nn.embedding_lookup_sparse` will raise an error with  invalid IDs ( < 0).

elasticdl/python/elasticdl/layers/embedding.py
```diff
@@ -140,7 +141,7 @@ def _check_id_valid(self, ids):
...
                 "The embedding id cannot be less than zero. "
```
terrytangyuan: Here it says cannot be zero. Need to modify the message to tell users what could go wrong if id < 0

workingloong:
I will refactor the embedding implementation in embedding_delegate.py and prune the invalid IDs (<0) like `tf.nn.safe_embedding_lookup_sparse`. So I close the PR.

PR #1581, opened 2019-12-20
Support learning rate scheduler

skydoorkai:
The user can optionally provide a learning rate scheduler method, which takes a model version as its input,  and returns a corresponding learning rate value.

### DIFF
--- a/elasticdl/python/common/lr_scheduler.py
+++ b/elasticdl/python/common/lr_scheduler.py
@@ -1,0 +1,47 @@
+import threading
+
+
+class LearningRateScheduler:
+    def __init__(self, learning_rate=0.001, model_version=0):
+        """Constructs a `LearningRateScheduler` instance.
+        Args:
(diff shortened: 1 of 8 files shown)


elasticdl/python/common/lr_scheduler.py
```diff
@@ -0,0 +1,47 @@
+import threading
+
+
+class LearningRateScheduler:
+    def __init__(self, learning_rate=0.001, model_version=0):
```
workingloong, line 5: Do we need to support scheduling the learning rate by epoch?

QiJune (approved):
Excellent!

elasticdl/python/common/lr_scheduler.py
skydoorkai: PS does not know how many batches in an epoch. The user knows it and can set the scheduler accordingly.

PR #1584, opened 2019-12-20
implement a parallel transform for data preprocessing

QiJune:
Using py_function in `tf.data.Dataset` could not be parallelized due to "GIL" of Python Interpreter. For more details, please refer to this [issue](https://github.com/tensorflow/tensorflow/issues/15694#issuecomment-401857127) and [question](https://stackoverflow.com/questions/48484689/parallelism-isnt-reducing-the-time-in-dataset-map).

The implementation of ParallelTransform takes the [DataLoader](https://github.com/pytorch/pytorch/blob/master/torch/utils/data/dataloader.py) of PyTorch as a reference. It could parallelize the data transform process with low latency.

### DIFF
--- a/elasticdl/python/data/parallel_transform.py
+++ b/elasticdl/python/data/parallel_transform.py
@@ -26,6 +26,30 @@ def worker_loop(records, index_q, result_q, transform_fn):
 
 
 class ParallelTransform(object):
+    """
+    The ParallelTransform applies transform_fn to records with a number
+    of workers with low latency.
+
+    Worker process side:
+    1. get index from index queue
+    2. transform_fn(records[index])
+    3. put result to result queue
+
+    Main process side:
+    1. put index to index queue of workers in round-robin way
+    2. get transformed data from result queue
+
+    We want to achieve two things:
+    1. Data samples in input records are transformed with low latency.
+       We transform the data samples one by one. Once we call `next_data`
+       method to get a transformed result, the next index is put to a
+       index queue and triggers a worker process.
+
+    2. The output order is strictly the same with the input order.
+       We use a `cache_dict` to adjust the order, since the result order
+       getting from result queue is sequential sometimes.
+    """
+
     def __init__(self, records, num_workers, transform_fn):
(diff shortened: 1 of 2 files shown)


elasticdl/python/data/parallel_transform.py
```diff
@@ -0,0 +1,81 @@
+import itertools
+from multiprocessing import Process, Queue
+
+
+def worker_loop(records, index_q, result_q, transform_fn):
```
workingloong, line 5: Add a docstring to explain those arguments.

QiJune: Done

elasticdl/python/data/parallel_transform.py
```diff
@@ -0,0 +1,99 @@
...
+    while True:
+        index = index_q.get()
+        if index is None:
+            break
+        record = records[index]
+        res = transform_fn(record)
+        result_q.put((index, res))
+
+
+class ParallelTransform(object):
```
brightcoder01, line 28: Can we add an overall description for the ParellelTransform?

QiJune: Done

elasticdl/python/data/parallel_transform.py
```diff
@@ -0,0 +1,123 @@
...
+       We transform the data samples one by one. Once we call `next_data`
+       method to get a transformed result, the next index is put to a
+       index queue and triggers a worker process.
+
+    2. The output order is strictly the same with the input order.
+       We use a `cache_dict` to adjust the order, since the result order
+       getting from result queue is sequential sometimes.
+    """
+
+    def __init__(self, records, num_workers, transform_fn):
```
skydoorkai: `num_workers` is a little confusing in ElasticDL.
`num_parallel_processes` is better.

QiJune: Done

skydoorkai (approved):

PR #1588, opened 2019-12-25
Sum up the values of the duplicated indices in the grads of the Index…

brightcoder01:
For the dense embedding, the gradients from tape is IndexedSlices. If the input_idx of embedding is large, the gradients payload is (input_idx.size, embedding_dim), it will be large with many duplicated indices.
At the stage of sending gradients from worker to parameter server, the scatter and serialization of this payload will be time consuming. We choose to sum up the value of duplicated indices at the worker side. 
From the time profiling (1 Worker with 4 CPU, 1 PS), the time cost for collecting embedding variable and building PushGradientRequest reduce from 0.33 seconds to 0.08 seconds.

### DIFF
--- a/elasticdl/python/common/tensor_utils.py
+++ b/elasticdl/python/common/tensor_utils.py
@@ -6,3 +6,21 @@ def merge_indexed_slices(*args):
         tf.concat([i.values for i in args], axis=0),
         tf.concat([i.indices for i in args], axis=0),
     )
+
+
+def deduplicate_indexed_slices(values, indices):
+    """Sums `values` associated with any non-unique `indices`.
+    Args:
+        values: A `Tensor` with rank >= 1.
+        indices: A one-dimensional integer `Tensor`, indexing into the first
+        dimension of `values` (as in an IndexedSlices object).
+    Returns:
+        A tuple of (`summed_values`, `unique_indices`) where `unique_indices`
+        is a de-duplicated version of `indices` and `summed_values` contains
+        the sum of `values` slices associated with each unique index.
+    """
+    unique_indices, new_index_positions = tf.unique(indices)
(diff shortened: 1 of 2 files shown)


skydoorkai (approved):

elasticdl/python/common/tensor_utils.py
```diff
@@ -9,18 +9,21 @@ def merge_indexed_slices(*args):
 
 
 def deduplicate_indexed_slices(values, indices):
-    """Sums `values` associated with any non-unique `indices`.
+    """
+    Sum up the values associated with duplicated indices in the IndexedSlices.
```
skydoorkai: No `IndexedSlices` in this method.
Can rephrase as:
Sums up values associated with the same indices and returns unique indices with corresponding summed values.

elasticdl/python/common/tensor_utils.py
```diff
@@ -9,18 +9,21 @@ def merge_indexed_slices(*args):
...
     Args:
-        values: A `Tensor` with rank >= 1.
-        indices: A one-dimensional integer `Tensor`, indexing into the first
-        dimension of `values` (as in an IndexedSlices object).
+        values: A Tensor with rank >= 1. Particularly IndexedSlices.values.
```
skydoorkai: Also removes IndexedSlices here.

skydoorkai (approved):

elasticdl/python/common/tensor_utils.py
brightcoder01: Good suggestion, updated.

elasticdl/python/common/tensor_utils.py
brightcoder01: Good suggestion, updated.

PR #1596, opened 2019-12-27
Replace embedding layers whose memory is bigger than 2MB

workingloong:
It is inefficient to train the small embedding layers using ElasticDL embedding layers. So we only replace the big embedding layers.

### DIFF
--- a/elasticdl/python/common/model_handler.py
+++ b/elasticdl/python/common/model_handler.py
@@ -50,7 +50,8 @@ def _need_partition_embedding(layer):
     bigger than 2MB.
     """
     EMBEDDING_SIZE_THRESHOLD_FOR_PARTITION = 2 * 1024 * 1024  # 2MB
-    weights_memory = layer.input_dim * layer.output_dim * 4
+    FLOAT32_BYTES = 4
+    weights_memory = layer.input_dim * layer.output_dim * FLOAT32_BYTES
     return weights_memory > EMBEDDING_SIZE_THRESHOLD_FOR_PARTITION
 
 
--- a/elasticdl/python/common/model_handler.py
+++ b/elasticdl/python/common/model_handler.py
@@ -45,8 +45,8 @@ def _convert_embedding_table_to_numpy_array(embedding_table, embedding_shape):
 
 
 def _is_big_embedding(layer):
-    """Determine whether the embedding is small by the condition that
-    the memory of the layer.train_weights is less than 2MB.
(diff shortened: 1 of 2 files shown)


elasticdl/python/common/model_handler.py
```diff
@@ -44,6 +44,15 @@ def _convert_embedding_table_to_numpy_array(embedding_table, embedding_shape):
     return embedding_weights
 
 
+def _is_big_embedding(layer):
```
brightcoder01: Please make this API compatible with feature column

elasticdl/python/common/model_handler.py
```diff
@@ -44,6 +44,15 @@ def _convert_embedding_table_to_numpy_array(embedding_table, embedding_shape):
...
+    """Determine whether the embedding is big by the condition that
+    the memory of the layer.train_weights is bigger than 2MB.
+    """
+    max_memory = 2 * 1024 * 1024  # 2MB
```
brightcoder01: Name constant as all cap.
EMBEDDING_SIZE_THRESHOLD_FOR_PARTITION

elasticdl/python/common/model_handler.py
brightcoder01: _is_big_embedding -> _need_partition?

workingloong: Good name.

workingloong: I will implement the replacement for `embedding_column` in the next PR.

elasticdl/python/common/model_handler.py
```diff
@@ -44,6 +44,16 @@ def _convert_embedding_table_to_numpy_array(embedding_table, embedding_shape):
     return embedding_weights
 
 
+def _need_partition_embedding(layer):
+    """The embedding layer will be partitioned on multiple
+    PS instances if the memory of the layer.train_weights is
+    bigger than 2MB.
+    """
+    EMBEDDING_SIZE_THRESHOLD_FOR_PARTITION = 2 * 1024 * 1024  # 2MB
+    weights_memory = layer.input_dim * layer.output_dim * 8
```
QiJune: If the weight type is float32, it should be `layer.input_dim * layer.output_dim * 4`, since one float32 occupy 4 bytes.

workingloong: Yes, the dtype of embedding weights is float32.

brightcoder01 (approved):

Issue #1592, opened 2019-12-27
add timeout mechanism to master


QiJune:
The master will watch the status of each worker. If a worker reports nothing exceeding a certain time, the master will kill the worker pod and relaunch a new one.

PR #1598, opened 2019-12-27
clean event callback code

QiJune:
This PR cleans event callback code and exposes two interfaces,  `relaunch_worker` and `relaunch_ps`. We could reuse these two interfaces for the further [timeout feature](https://github.com/sql-machine-learning/elasticdl/issues/1592) development.

### DIFF
--- a/elasticdl/python/master/k8s_instance_manager.py
+++ b/elasticdl/python/master/k8s_instance_manager.py
@@ -172,22 +172,22 @@ class InstanceManager(object):
     def stop_relaunch_and_remove_workers(self):
         with self._lock:
             self._relaunch_deleted_live_worker = False
-            for _, (worker_id, _) in self._worker_pods_phase.items():
+            for (worker_id, _) in self._worker_pods_phase.values():
                 self._k8s_client.delete_worker(worker_id)
 
     def stop_relaunch_and_remove_all_ps(self):
         with self._lock:
             self._relaunch_deleted_live_ps = False
-            for _, (ps_id, _) in self._ps_pods_phase.items():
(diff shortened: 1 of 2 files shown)


elasticdl/python/master/k8s_instance_manager.py
```diff
@@ -236,49 +262,14 @@ def _event_cb(self, event):
...
-        elif relaunch_ps:
-            logger.info("Relaunching ps.")
-            # Note: the ID and service address for relaunched parameter
-            # server are intentionally left unchanged to support fault
-            # tolerance.
-            self._start_ps(ps_id)
+                pods_phase[pod_name][1] = phase
+                if relaunch_flag:
+                    remove_fn(pods_phase[pod_name][0])
+                    relaunch_fn(pod_name)
```
QiJune: If the pod status is `OOMKilled`, we should kill it first, and then relaunch a new one.

skydoorkai: If we kill OOMKilled pods, the logs are lost. We have two choices: keep failed pods so users can check logs, or delete them and the logs are lost.

QiJune: I find that we could use `timeout` mechanism to summarize the following circumstances:

- OOM
- killed by preemption
- hang

The master will record the time when a worker get a task. Then the task will be put to doing list. The master will check doing list, if the time exceeds a value, which means the worker may fail due to some reasones, the task in doing list will be remoed to todo list.

QiJune:
Close this PR, please refer to #1600

PR #1599, opened 2019-12-27
Only use a lock for updating parameters with embedding.

workingloong:
The `apply_gradient` in TensorFlow optimizer is thread-safe, so we don't need to use a lock.

### DIFF
--- a/elasticdl/python/ps/optimizer_wrapper.py
+++ b/elasticdl/python/ps/optimizer_wrapper.py
@@ -177,8 +177,7 @@ class OptimizerWrapper(object):
                 grads_and_vars_new.append((grad, var))
         self._opt.apply_gradients(grads_and_vars_new)
         self._update_embedding_param()
-        with self._update_gradient_lock:
-            self._delete_variables()
+        self._delete_variables()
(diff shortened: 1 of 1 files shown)


skydoorkai (approved):

elasticdl/python/ps/optimizer_wrapper.py
```diff
@@ -155,21 +156,28 @@ def apply_gradients(self, grads_and_vars):
...
-            self._opt.apply_gradients(grads_and_vars_new)
-            self._update_embedding_param()
-            self._delete_variables()
+        if self._has_embedding:
+            with self._update_gradient_lock:
+                self._update_parameter_by_gradients(grads_and_vars)
+        else:
+            self._update_parameter_by_gradients(grads_and_vars)
+
+    def _update_parameter_by_gradients(self, grads_and_vars):
```
brightcoder01: _update_parameter_by_gradients -> _update_parameters_by_gradients

workingloong: Done

PR #1600, opened 2019-12-30
Implement task timeout mechanisim

QiJune:
Fix #1529 

I find that we could use a timeout mechanism to summarize the following circumstances which workers meet:

- OOM
- killed by preemption
- hang

The master will record the time when a worker gets a task. Then the task will be put to the doing list. The master will check the doing list, if the time exceeds a value, which means the worker may fail due to some reasons, the task in doing list will be removed to the todo list.


skydoorkai (commented):
> Fix #1529
> 
> I find that we could use a timeout mechanism to summarize the following circumstances which workers meet:
> 
> * OOM
> * killed by preemption
> * hang
> 
> The master will record the time when a worker gets a task. Then the task will be put to the doing list. The master will check the doing list, if the time exceeds a value, which means the worker may fail due to some reasons, the task in doing list will be removed to the todo list.

Since we can detect if worker pods are deleted or OOMed, it is still worth to detect and relaunch worker pods immediately when these events happen. 
Using timeout will delay the relaunch until timeout happens.

Also, we should set the timeout value according to real task processing times, not set it by default.
Changing model/task size/pod resources should change the value.

QiJune:
Close this PR temporarily

PR #1602, opened 2019-12-31
Read records in the task by small chunks from odps

### LINKED ISSUE #1601: The memory grows fast when the num_minibatches_per_task in args is big using odps table.
A task may contain a larger number of records. If we directly download all records in the task from ODPS, huge memory is needed which may cause a memory leak or OOM. So, we need to split the records in the task to small chunks and download the small chunk to yield each time.

workingloong:


elasticdl/python/data/odps_io.py
```diff
@@ -266,22 +269,41 @@ def _estimate_cache_batch_count(self, columns, table_size, batch_size):
...
+        """Estimate the count of records with 20MB to download
+        """
+        record_size = self._estimate_record_size(columns)
+        max_count = int(MAX_DOWNLOAD_SIZE / record_size)
+        return max_count
+
+    def _estimate_record_size(self, columns):
+        """Estimate the average size of each record in the ODPS table.
+        """
+        sample_count = 10
```
terrytangyuan, line 297: What’s this for?

workingloong: I have submitted another PR #1603 and the PR will be closed.

workingloong:
I have submitted #1603  to resolve the same problem, so i closed it.

PR #1604, opened 2019-12-31
variables from different ps have its own model_version

skydoorkai:
1. variables from different PS have their own model_version in workers.

2. add `sync_version_tolerance` arg so that sync-SGD can accept gradients with different versions from PS (default is 0 to only accept gradients with the same version as PS).

3.  `pull_variable` grpc does not need to return variable values if PS model version is the same as the requested worker.

### DIFF
--- a/elasticdl/python/worker/worker.py
+++ b/elasticdl/python/worker/worker.py
@@ -99,6 +99,8 @@ class Worker(object):
                 ]
                 self._var_to_ps = {}
                 self._ps_num = len(self._ps_stubs)
+        else:
+            self._ps_num = 0
         self._distribution_strategy = args.distribution_strategy
         if (
             self._distribution_strategy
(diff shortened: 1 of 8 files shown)


elasticdl/python/worker/worker.py
```diff
@@ -308,27 +311,32 @@ def get_model(self):
...
                 if not res.model_init_status:
                     # TODO: support PS fault-tolerance
                     raise RuntimeError(
                         "PS pod %d cannot be initialized" % ps_id
                     )
 
-            for tensor_pb in res.model.param:
-                tensor = Tensor.from_tensor_pb(tensor_pb)
-                self._non_embed_vars[tensor.name].assign(tensor.to_ndarray())
+            if res.model.version > self._model_versions_from_ps[ps_id]:
```
QiJune: If `res.model_version <= self._model_versions_from_ps[ps_id]`, the parameter server will not set mdel.param. So this field is empty. The for loop will be skipped. It seems that the logic here in worker is duplicate with PS.

skydoorkai: You are right. No need to check version here.

QiJune (approved):
Excellent!

PR #1611, opened 2020-01-03
Reuse the variables created for embedding vectors and gradients in OptimizerWrapper.

workingloong:
It is inefficient and slow to create variables for each mini-batch gradients. We can only create variables for the first mini-batch and reuse those variables by assigning new values in PS. So, we can speed up the `apply_gradient` in the OptimizerWrapper.

### DIFF
--- a/elasticdl/python/ps/optimizer_wrapper.py
+++ b/elasticdl/python/ps/optimizer_wrapper.py
@@ -72,10 +72,8 @@ class OptimizerWrapper(object):
     ):
         """
         Note:
-            We need to support Redis and ElasticDL parameter server at the
-            same time. If `lookup_embedding_func`/`update_embedding_func`
+            If `lookup_embedding_func`/`update_embedding_func`
             is not None, use parameter server to lookup/update embedding.
-            Otherwise use Redis.
 
         Arguments:
             opt: A TensorFlow optimizer instance.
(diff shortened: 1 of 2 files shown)


elasticdl/python/ps/optimizer_wrapper.py
```diff
@@ -307,12 +324,6 @@ def _delete_variables(self):
                         break
                     else:
                         opt_weight_iter += 1
-        for key in list(self._tls._slot_variables.keys()):
-            del self._tls._slot_variables[key]
-
-        # Delete variables in embed_variables.
-        for key in list(self._tls._embed_variables.keys()):
-            del self._tls._embed_variables[key]
```
skydoorkai, line 315: Rename method name, and update comments.

By the way, also delete those comments related to Redis in this file.

elasticdl/python/ps/optimizer_wrapper.py
```diff
@@ -241,40 +246,52 @@ def _create_slot_variable(self, layer_name, slot_name, initial_value):
...
+            slot_var.assign(initial_value)
         return slot_var
 
     # This is a function modified from TensorFlow optimizers.
     # https://github.com/tensorflow/tensorflow/blob/
     # 69b1feac62276edcc509ac88af229c6236e645fe/tensorflow/python
     # /keras/optimizer_v2/optimizer_v2.py#L567
-    def _create_slot_variable_in_optimizer(
-        self, embed_var, slot_name, shape, initial_value
+    def _insert_slot_variable_to_optimizer(
```
terrytangyuan: insert -> update?

skydoorkai (approved):

PR #1612, opened 2020-01-03
fix a worker relaunch bug when worker is relaunched after all training tasks are done

skydoorkai:
Need to handle the case that when a worker pod is relaunched, there are no more tasks or only WAIT or SAVE_MODEL tasks.

### DIFF
--- a/elasticdl/python/worker/task_data_service.py
+++ b/elasticdl/python/worker/task_data_service.py
@@ -1,4 +1,5 @@
 import threading
+import time
 from collections import deque
 
 import tensorflow as tf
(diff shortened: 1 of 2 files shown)


elasticdl/python/worker/task_data_service.py
```diff
@@ -153,11 +154,22 @@ def get_dataset(self):
...
             # records so this should not be time consuming.
             if self._warm_up_task is None and not self._has_warmed_up:
-                task = self._worker.get_task()
-                self._warm_up_task = task
-                for _ in self.data_reader.read_records(task):
-                    break
-                self._has_warmed_up = True
+                while True:
+                    task = self._worker.get_task()
+                    if task.type != elasticdl_pb2.WAIT:
```
chunyang-wen, line 159: Will the program hang here forever?

skydoorkai: No, this is the same logic as before (without relaunch).
(1) the master will stop sending `WAIT` task after doing tasks are done. 
(2) In the unlikely event that the master fails, `get_task` RPC call will fail the worker pod.

terrytangyuan: Perhaps it would be good to add some comments on this logic?

terrytangyuan (approved):

PR #1613, opened 2020-01-03
add cpp build system with bazel

QiJune:


elasticdl/cc/common/tensor.h
```diff
@@ -0,0 +1,72 @@
+#pragma once
```
skydoorkai: Should we use `gragma once` or include guards? Seems that the only difference is that if the header file is copied around different locations by the compiler, `gragma once` won't work.

QiJune: I have no preference for using `#pragma once` or include guards. If google style suggests include guards, I will follow it.

elasticdl/cc/common/type.cc
```diff
@@ -0,0 +1 @@
+#include "elasticdl/cc/common/type.h"
```
skydoorkai: Why need `type.cc` which contains only one line inlude ?

QiJune: Fix it.

QiJune:
Close this PR. We plan to make C++ very thin, and Makefile is enough.

PR #1624, opened 2020-01-09
Add broadcast op and use gossip + pytorch implementation of FTLib

terrytangyuan:
* Broadcast op is supported in FTLib now so we switch to use that.
* Gossip consensus implementation is preferred and removes the requirement on shared storage.
* PyTorch implementation of communication library is preferred since it's more mature than the dummy NCCL implementation. We switch to use this to speed up the development of the POC.


skydoorkai:
When ftlib is installed, is pytorch automatically installed?

terrytangyuan:
@skydoorkai Right now yes but it will probably be changed to optional later, depending on whether the user wants to use pytorch backend or not.

skydoorkai (approved):

PR #1625, opened 2020-01-10
add golang support in Dockerfile.dev

skydoorkai:
1. Go and related tools

2. protoc-gen-go to support GRPC in Go.

### DIFF
--- a/elasticdl/docker/scripts/install-go.bash
+++ b/elasticdl/docker/scripts/install-go.bash
@@ -1,0 +1,15 @@
+#!/bin/bash
+
+set -e
+
+curl --silent https://dl.google.com/go/go1.13.4.linux-amd64.tar.gz | tar -C /usr/local -xzf -
+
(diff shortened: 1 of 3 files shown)


elasticdl/docker/scripts/install-go.bash
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+set -e
+
+curl --silent https://dl.google.com/go/go1.13.4.linux-amd64.tar.gz | tar -C /usr/local -xzf -
+
+go get github.com/golang/protobuf/protoc-gen-go
```
terrytangyuan, line 7: Are all these required? Maybe we can add some of them later on when they become necessary

skydoorkai: I copied it from sqlflow.
golint, goyacc etc are used for pre-commit check of the go files.
gotest, goveralls are needed for testing.

terrytangyuan (approved):

PR #1636, opened 2020-01-14
Create golang PS server

skydoorkai:


elasticdl/pkg/server/psserver.go
```diff
@@ -0,0 +1,86 @@
+package main
```
QiJune: package main --> package ps?

And what about 
elasticdl/pkg/server/psserver.go -->  elasticdl/pkg/ps/server.go

elasticdl/pkg/server/psserver_test.go
```diff
@@ -0,0 +1,88 @@
+package main
```
QiJune: The same.

elasticdl/pkg/server/psserver.go
skydoorkai: Renamed the directory and filename.

For the package name, it must be `main`, as this is an executable go file with `func main`.

QiJune (approved):
Excellent!

PR #1637, opened 2020-01-15
Add SGD Kernel Go Package

QiJune:
The SGD kernel is implemented in C++ and exposed to Go via CGo.

I make a wrapper `kernel.go`, so other functions in Go could directly call SGD function in Go.


skydoorkai:
In Dockerfile, we should copy `elasticdl/pkg` to somewhere under`GOPATH`, then `go install`

elasticdl/proto/elasticdl.proto
```diff
@@ -1,146 +1,162 @@
 syntax = "proto3";
 
 import "google/protobuf/empty.proto";
-import "elasticdl/proto/tensor_dtype.proto";
 
 package master;
 option go_package = "elasticdl/pkg/proto";
 
+enum TensorDtype {
```
skydoorkai: Probably use a separate PR to combine `tensor_dtype.proto` and `elasticdl.proto` together.

QiJune: Please refer to #1638

scripts/codestyle/cpplint_precommit.hook
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
```
wangkuiyi, line 1: Other than crafting our own scripts, could we use https://github.com/bmorcos/pre-commit-hooks-cpp or other pre-made C++ hooks for pre-commit?

elasticdl/pkg/kernel/capi/kernel_api.cc
```diff
@@ -0,0 +1,15 @@
+extern "C" {
```
wangkuiyi: In my experience, `extern "C"` is often used in header files other source files. For example

```c++
#ifdef __cplusplus  
extern "C" { 
#endif 
    /* Declarations of this file */
#ifdef __cplusplus 
} 
#endif 
```

Is this some style requirement to use it in C++ source files?

elasticdl/pkg/kernel/capi/kernel_api.h
```diff
@@ -0,0 +1,5 @@
+#ifndef ELASTICDL_KERNEL_API_H_
+#define ELASTICDL_KERNEL_API_H_
+
+void SGD(float* grad, float* param, double lr, long long size);
+#endif
```
wangkuiyi, line 13: The Google C++ style guide has some requirements on guard https://google.github.io/styleguide/cppguide.html#The__define_Guard

elasticdl/pkg/kernel/capi/kernel_api.cc
QiJune: Yes, you are right. In we want to use a C header file in a C++ source file, we have to add `extern "C"` to the C header file. There are two cases:

- If the C header file is written by myself, I'd better add macros to C header filer.
- If the C header file is from other places, for example, a Linux system C header file, I have to add `extern "C"` in C++ source file.

I will fix it.

elasticdl/pkg/kernel/capi/kernel_api.h
QiJune: Done

scripts/codestyle/cpplint_precommit.hook
QiJune: The precommit hook files are just modified from the paddle repo.

I tried this one https://github.com/bmorcos/pre-commit-hooks-cpp, but got the following error:

```
[INFO] Installing environment for https://github.com/bmorcos/pre-commit-hooks-cpp.
[INFO] Once installed this environment will be reused.
[INFO] This may take a few minutes...
An unexpected error has occurred: CalledProcessError: Command: ('/Users/qijun/.cache/pre-commit/repo4qxgesqb/py_env-python3.7/bin/python3.7', '/Users/qijun/.cache/pre-commit/repo4qxgesqb/py_env-python3.7/bin/pip', 'install', '.')
Return code: 1
Expected return code: 0
Output: (none)
Errors:
    ERROR: Directory '.' is not installable. Neither 'setup.py' nor 'pyproject.toml' found.

Check the log at /Users/qijun/.cache/pre-commit/pre-commit.log
```

We could use our own scripts currently. After debugging this, we could transfer to pre-made C++ hooks.

elasticdl/docker/Dockerfile
```diff
@@ -6,7 +6,7 @@ FROM ${BASE_IMAGE}
 COPY elasticdl/docker/bashrc /etc/bash.bashrc
 RUN chmod a+rwx /etc/bash.bashrc
 
-RUN apt-get update && apt-get install -y unzip curl git software-properties-common
+RUN apt-get update && apt-get install -y unzip curl git software-properties-common libeigen3-dev
```
skydoorkai, line 9: For the last line of this file, it will fail as Go is not installed.
We can change it to 
`RUN make -f elasticdl/Makefile python_pb`
for now, and figure out how to install the PS Go version later.

QiJune: Done

skydoorkai (approved):

PR #1646, opened 2020-01-17
add tensor type in Go

QiJune:


elasticdl/pkg/common/tensor.go
```diff
@@ -0,0 +1,54 @@
...
+// TODO(qijun): handle different tensor dtype
+type Tensor struct {
+	Name    string
+	Value   []float32
+	Dim     []int64
+	Indices []int64
+}
+
+// GetTensorSize returns the size of a tensor
+func GetTensorSize(t *Tensor) int64 {
```
skydoorkai: We can use method of struct instead:
```
func (t *Tensor)Size() int64 {
   ...
}
```
Then for a `tensor`, call `tensor.Size()` to get its size.

elasticdl/pkg/common/tensor.go
```diff
@@ -0,0 +1,54 @@
...
+	var size int64 = 1
+	for _, d := range t.Dim {
+		size *= d
+	}
+	return size
+}
+
+// DeserializeTensorPB transforms pb to tensor
+func DeserializeTensorPB(pb *proto.Tensor, t *Tensor) {
```
skydoorkai: What is the usage of this func? Should `Tensor` be pre-defined? Or we can define a func with a input as `*proto.Tensor` and returns a `*Tensor`?

elasticdl/pkg/common/tensor.go
```diff
@@ -0,0 +1,54 @@
...
+	copy(t.Dim, pb.Dim)
+	t.Indices = make([]int64, len(pb.Indices))
+	copy(t.Indices, pb.Indices)
+	t.Value = make([]float32, len(pb.Content)/4)
+	br := bytes.NewReader(pb.GetContent())
+	binary.Read(br, binary.LittleEndian, &t.Value)
+}
+
+// SerializeTensor transforms tensor to pb
+func SerializeTensor(t *Tensor, pb *proto.Tensor) {
```
skydoorkai: Same for this method.

QiJune: Done

elasticdl/pkg/common/tensor.go
QiJune: Modify this function, to get a `*proto.Tensor`, and return a `*Tensor`

elasticdl/pkg/common/tensor.go
QiJune: Rename It to `GetDimProduct`, so it could be used for any slice.

elasticdl/pkg/common/tensor.go
```diff
@@ -0,0 +1,64 @@
...
+// TODO(qijun): handle different tensor dtype
+type Tensor struct {
+	Name    string
+	Value   []float32
+	Dim     []int64
+	Indices []int64
+}
+
+// GetDimProduct returns the product of a slice of dim
+func GetDimProduct(dim []int64) int64 {
```
skydoorkai: Will `GetDimProduct` be used outside this file? If not, change it to an unexported func. If yes, keep as it is.

QiJune: I am not very sure. 
We could set it to an unexported func first. If it is used in another place, we change it.

skydoorkai (approved):

PR #1663, opened 2020-01-19
add Parameter struct in Go

### LINKED ISSUE #1657: Add Parameter struct in Go

QiJune:
fix #1657

### DIFF
--- a/elasticdl/pkg/ps/parameter.go
+++ b/elasticdl/pkg/ps/parameter.go
@@ -1,6 +1,6 @@
 package ps
 
-import "elasticdl.org/elasticdl/common/tensor"
+import "elasticdl.org/elasticdl/pkg/common"
 
 // Parameter contains non-embedding param and embedding param
 type Parameter struct {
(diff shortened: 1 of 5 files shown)


elasticdl/pkg/ps/parameter.go
```diff
@@ -0,0 +1,16 @@
...
+
+import "elasticdl.org/elasticdl/pkg/common"
+
+// Parameter contains non-embedding param and embedding param
+type Parameter struct {
+	NonEmbeddingParam map[string]common.Tensor
+}
+
+// GetNonEmbeddingParam returns non-embedding tensor pointer
+func GetNonEmbeddingParam(p *Parameter, name string) *common.Tensor {
```
skydoorkai: Use struct method:
```
func (p *Parameter) NonEmbeddingParam(name string) *common.Tensor {
```

QiJune: Done

skydoorkai (approved):

PR #1673, opened 2020-01-20
add Adam kernel (#1660)

lwpyr:


elasticdl/pkg/kernel/kernel_test.go
```diff
@@ -25,3 +26,58 @@ func TestSGD(t *testing.T) {
...
+	expectedM := make([]float32, size)
+	expectedV := make([]float32, size)
+
+	for i := 0; i < size; i++ {
+		expectedM[i] = float32(beta1)*m[i] + (1-float32(beta1))*grad[i]
+		expectedV[i] = float32(beta2)*v[i] + (1-float32(beta2))*grad[i]*grad[i]
+	}
+
+	for i := 0; i < size; i++ {
+		expectedParam[i] = param[i] - float32(lr)*expectedM[i]/(1-float32(math.Pow(beta1, float64(step))))/(float32(math.Sqrt(float64(expectedV[i]/(1-float32(math.Pow(beta2, float64(step))))))+epsilon))
```
QiJune, line 70: This Line is too long.  Try `gofmt -w kernel_test.go`?

elasticdl/pkg/kernel/kernel_test.go
```diff
@@ -25,3 +26,58 @@ func TestSGD(t *testing.T) {
...
+	}
+
+	Adam(grad, param, m, v, lr, int64(size), step, beta1, beta2, epsilon)
+
+	diffParam := 0
+	diffM := 0
+	diffV := 0
+
+	for i := 0; i < size; i++ {
+		if math.Abs(float64(expectedM[i]-m[i])) > 1e-6 {
```
QiJune: It seems that we will compare two float array often, could we add a common util `Cmp` function to do this?

Here is a reference  https://dev.to/juliaferraioli/testing-in-go-testing-floating-point-numbers-4i0a

elasticdl/pkg/kernel/capi/kernel_api.cc
```diff
@@ -11,3 +12,33 @@ void SGD(float* grad, float* param, double lr, long long size) {
 
   ep -= lr * eg;
 }
+
+void Adam(float* grad,
```
skydoorkai, line 16: According to [tensorflow doc ](https://www.tensorflow.org/api_docs/python/tf/keras/optimizers/Adam,)there is another bool argument `amsgrad`.

lwpyr: AMSGRAD is a modification of Adam, which needs to maintain the running means of m and v,  maybe it is better to write another kernel.

skydoorkai: When amsgrad == true, one more value is required, `vhat` for the maximum of all 2nd moment running avg so far. We can put it in Adam as other logics are similar.

elasticdl/pkg/kernel/kernel_test.go
lwpyr: I used the pre-commit run, do I still need to manually run this command?

elasticdl/pkg/kernel/kernel_test.go
lwpyr: fixed

elasticdl/pkg/kernel/kernel_test.go
QiJune: I am not sure why `pre-commit` is not worked as expected. I will debug on this.
You could run `gofmt` manually in this PR.

skydoorkai: Go does not limit line length. It is up to the user to make it reasonable.

elasticdl/pkg/kernel/capi/kernel_api.cc
```diff
@@ -11,3 +12,42 @@ void SGD(float* grad, float* param, double lr, long long size) {
...
+  ev = beta2 * ev + (1.0 - beta2) * eg.square();
+
+  if (max_square != NULL) {
+    Eigen::Map<Eigen::Array<float, 1, Eigen::Dynamic>> ems{
+        max_square, static_cast<Eigen::Index>(size)};
+    ems = ems.cwiseMax(ev);
+    ep -= lr * em / (1.0 - pow(beta1, step)) /
+          ((ems / (1.0 - pow(beta2, step))).sqrt() + epsilon);
+  } else {
+    ep -= lr * em / (1.0 - pow(beta1, step)) /
```
skydoorkai, line 50: We may optimize the `pow` computations fo `beta1` and `beta2` by send the `pow` result as a pre-computed argument. This can be done in another PR. We can check-in this PR now.

skydoorkai (approved):

PR #1684, opened 2020-01-29
Restructure the reasoning of Data Analysis

wangkuiyi:


docs/designs/data_transform.md
```diff
@@ -1,4 +1,84 @@
-# Data Tranform Design Doc
+# Data Analysis and Tranformation
+
+## The Problem
+
+We want to add data analysis, in addition to data transformation, into the end-to-end machine learning pipeline generated by SQLFlow.
```
brightcoder01, line 5: In this design doc, we also want to add some built-in functions for transformation. (`standardize`, `normalize`, `log` and more on demand in the future)
For example, `standardize` is mapped to a python function.

docs/designs/data_transform.md
```diff
@@ -1,4 +1,84 @@
...
+
+Please be aware that if the user wants normalization, she must write `normalize(diameter)` instead of `diameter`.  The following example uses the raw value of `diameter` without normalization.
+
+```sql
+SELECT * FROM new_plates TO PREDICT price USING a_model COLUMN diameter
+```
+
+## The Challenge
+
+From the above examples, we see that a challenge is that the TO TRAIN clause must be able to save the input field names and the data analysis results and transformation steps together with the model, so to make sure that the predictions using the same data transformation as the training.
```
brightcoder01, line 87: Besides batch prediction job using `TO PREDICT` expression, we may also consider the scenario of model serving.

docs/designs/data_transform.md
```diff
@@ -1,4 +1,84 @@
...
+
+In the terminology of TFX Transform, the word *analysis* refers to data statistics that derives the above parameters, given which, the *transforming* refers to the conversion of raw data into model inputs.  This design document is about making SQLFlow support data analysis, in addition to transformation.
+
+
+## The Syntax Extension
+
+We can ask users to write SQL statements for data analysis.  For example, we might want to assume that the following code snippet trains a model to predict the plate price from its size.
+
+```sql
+min, max = SELECT min(age), max(age) FROM plates;
```
brightcoder01: We can also write SQL statement to do data transformation. How about adding the following case for comparison?
```SQL
SELECT 1.00 * (t1.age - t2.age_min) / t2.age_range
FROM plates t1
JOIN
(
    SELECT
        MIN(age) AS age_min,
        MAX(age) - MIN(age) AS age_range
    FROM plates
) t2
ON 1 = 1
```
The reason why we don't choose SQL for transformation: This transformation described by SQL can't be used directly in model serving.

wangkuiyi: Does this SQL statement really work? If so, your mastery of SQL the language is fantastic!

docs/designs/data_transform.md
wangkuiyi: I am afraid that it is too early to tell such details and distracts readers from the most important points.

docs/designs/data_transform.md
wangkuiyi: I am afraid that it is too much detailed to mentioned multiple ways of serving. We can have another document on prediction services. At the right moment, we can assume that we have some prediction services that can load models trained with SQLFlow.

docs/designs/data_transform.md
```diff
@@ -1,4 +1,84 @@
...
+To use the trained model to make predictions, we hope the users don't have to rewrite the COLUMN clause.   For example, the following statement assumes that the table `new_plates` has a field `size`. The converted Python code should normalize `size` before sending it to the input of `a_mode`.
+
+```sql
+SELECT * FROM new_plates TO PREDICT price USING a_model;
+```
+
+If the user doesn't want normalization, but the raw value, as the input, she could write the following.
+
+```sql
+SELECT * FROM new_plate TO PREDICT price USING a_model COLUMN size
```
brightcoder01, line 64: I'm afraid that the Prediction job will fail if use can write COLUMN in the `TO PREDICT` expression.
Let's take iris model for example:
If we train the model using
```SQL
SELECT * FROM iris.train
TO TRAIN DNNClassifer
WITH hidden_units = [10, 10], n_classes = 3, EPOCHS = 10
COLUMN sepal_length, sepal_width, petal_length, petal_width
LABEL class
INTO sqlflow_models.my_dnn_model;
```
And make prediction using:
```
SELECT * FROM iris.test
TO PREDICT iris.predict.class
USING sqlflow.my_dnn_model
COLUMN sepal_length, sepal_width, petal_length
```
The prediction job will fail. The reason is the expected input of the trained model contains four elements. But the actual input contains three.

wangkuiyi: And this failure is expected in our design. SQLFlow should prompt a message like "The inputs to the model during prediction are not compatible with the model specification.".  So, I think it is not a problem to allow a COLUMN clause in TO PREDICT.

docs/designs/data_transform.md
brightcoder01: Yes, it works, I have verified this expression in MySQL.

brightcoder01 (approved):

PR #1687, opened 2020-01-31
add DeserializeModelPB

QiJune:


elasticdl/pkg/ps/parameter_test.go
```diff
@@ -27,3 +28,39 @@ func TestParameterInit(t *testing.T) {
...
+	p, err := DeserializeModelPB(&modelPB)
+
+	assert.Nil(t, err)
+	assert.Contains(t, p.EmbeddingParam, "e1")
+
+	e1 := p.GetEmbeddingParam("e1")
+	assert.Equal(t, int64(2), e1.Dim)
+	assert.Equal(t, 3, len(e1.EmbeddingVector))
+
+	ev1 := e1.GetEmbeddingVector(1)
```
terrytangyuan, line 58: Test these in a loop and reduce the amount of duplicated code?

QiJune: A loop will not reduce the lines of test code here.
It's just a unit test, I think some duplicated code is acceptable.

lwpyr (approved):

PR #1692, opened 2020-02-03
add PushModel/PushEmbeddingInfo rpc in go

### LINKED ISSUE #1653: Implement PushEmbeddingInfo grpc service in Go

### LINKED ISSUE #1656: Implement PushModel grpc service in Go

QiJune:
Fix #1653 
Fix #1656

### DIFF
--- a/elasticdl/pkg/main/main.go
+++ b/elasticdl/pkg/main/main.go
@@ -32,7 +32,7 @@ func main() {
 	flag.Parse()
 	address := fmt.Sprintf("%s:%d", os.Getenv("MY_POD_IP"), *port)
 	serverDone := make(chan bool)
-	ps.CreateServer(address, int32(*psID), "SGD", 0.1, serverDone)
+	ps.CreateServer(address, *psID, "SGD", 0.1, serverDone)
 	log.Println("PS service started at ", address)
 	for {
 		select {
--- a/elasticdl/pkg/main/main.go
+++ b/elasticdl/pkg/main/main.go
@@ -32,7 +32,7 @@ func main() {
 	flag.Parse()
 	address := fmt.Sprintf("%s:%d", os.Getenv("MY_POD_IP"), *port)
 	serverDone := make(chan bool)
-	ps.CreateServer(address, serverDone)
+	ps.CreateServer(address, int32(*psID), "SGD", 0.1, serverDone)
 	log.Println("PS service started at ", address)
 	for {
 		select {
--- a/elasticdl/pkg/ps/server.go
+++ b/elasticdl/pkg/ps/server.go
@@ -15,20 +15,16 @@ type Server struct {
 	pb.PserverServer
 	Param *Parameter
 	Opt   Optimizer
-	PSID  int32
-	lock  *sync.Mutex
+	ID    int
+	lock  sync.Mutex
 }
 
 // NewServer creates a Server instance
-func NewServer(PSID int32, opt string, lr float32) *Server {
-	var ps Server
-	ps.Param = NewParameter()
-	if opt == "SGD" {
-		ps.Opt = NewSGDOptimizer(lr)
-	}
-	ps.PSID = PSID
-	ps.lock = &sync.Mutex{}
-	return &ps
+func NewServer(ID int, opt string, lr float32) *Server {
+	return &Server{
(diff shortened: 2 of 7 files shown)


elasticdl/pkg/main/main.go
```diff
@@ -32,7 +32,7 @@ func main() {
 	flag.Parse()
 	address := fmt.Sprintf("%s:%d", os.Getenv("MY_POD_IP"), *port)
 	serverDone := make(chan bool)
-	ps.CreateServer(address, serverDone)
+	ps.CreateServer(address, int32(*psID), "SGD", 0.1, serverDone)
```
wangkuiyi: Why do we constrain the type of ID to be int32? Can we use int instead to save the type conversion?

elasticdl/pkg/ps/server.go
```diff
@@ -7,47 +7,82 @@ import (
...
 	pb.PserverServer
+	Param *Parameter
+	Opt   Optimizer
+	PSID  int32
+	lock  *sync.Mutex
 }
 
-func (s *psServer) PullVariable(ctx context.Context, in *pb.PullVariableRequest) (*pb.PullVariableResponse, error) {
+// NewServer creates a Server instance
+func NewServer(PSID int32, opt string, lr float32) *Server {
```
wangkuiyi: PSID => ID

elasticdl/pkg/ps/server.go
```diff
@@ -7,47 +7,82 @@ import (
...
+	var ps Server
+	ps.Param = NewParameter()
+	if opt == "SGD" {
+		ps.Opt = NewSGDOptimizer(lr)
+	}
+	ps.PSID = PSID
+	ps.lock = &sync.Mutex{}
+	return &ps
```
wangkuiyi: You cannot return the address of a variable ps created in the stack.  Once the function call completes and the stack frame pops, the address becomes invalid.

The correct way is 

```go
return &Server {
    Param : NewParameter(),
    Opt: NewOptimizer(opt),
    ID: ID
}
```

It doesn't make sense to write

```go
ps.lock = & sync.Mutex()
```

because `lock ` should be defined in type Server as a variable but not an address:

```go
type Server {
    lock sync.Mutex
}
```

wangkuiyi (changes requested):

elasticdl/pkg/main/main.go
QiJune: Yes, int is fine. I will fix it.

elasticdl/pkg/ps/server.go
QiJune: 1. Go support escape analysis. Please refer to https://stackoverflow.com/questions/38234487/returning-a-pointer-on-stack . And there is also a more detailed article https://www.ardanlabs.com/blog/2017/05/language-mechanics-on-escape-analysis.html.

When return a pointer inside a function, it escapes from the stack to heap automatically. So it's fine to define a server variable inside a function, and return its address.

2. Yes, I think `sync.Mutex` is more clear, what about following code:

```go
type Server struct {
	pb.PserverServer
	Param *Parameter
	Opt   Optimizer
	ID  int
	lock  sync.Mutex
}

func NewOptimizer(opt string, lr float32) Optimizer {
    if opt == "SGD" {
        return NewSGDOptimizer(lr)
    }
    return nil
}

// NewServer creates a Server instance
func NewServer(ID int, opt string, lr float32) *Server {
	return &Server{
        Param : NewParameter(),
        Opt: NewOptimizer(opt, lr),
        ID: ID}
}
```

elasticdl/pkg/ps/server.go
QiJune: Done

mhaoli:
LGTM!

wangkuiyi (approved):

elasticdl/pkg/ps/server.go
wangkuiyi: Parameter => Parameters, Model ?

QiJune: Model is more approriate. 
I am writing the concepts design doc. And it seems that the code change will be a little large. I will refactor all related codes after the doc is done.

Issue #1690, opened 2020-02-03
Support tf.keras.callbacks in ElasticDL


workingloong:
A custom callback is a powerful tool to customize the behavior of a Keras model during training, evaluation, or inference, including reading/changing the Keras model. 
For example:
1. We can use early stopping in training to prevent overfitting.
2. We can post-process the trained model.

The callback methods in `tf.keras.callbacks.Callback` is listed in https://www.tensorflow.org/guide/keras/custom_callback#an_overview_of_callback_methods
We can support the callback methods one by one when the method is required.

Define callbacks in model_def
```python
import tensorflow as tf

class MyCustomCallback(tf.keras.callbacks.Callback):
    def on_train_begin(self, logs=None):
        print("Training: begins at {}".format(datetime.datetime.now().time()))
    
    def on_train_end(self, logs=None):
        print("Training: begins at {}".format(datetime.datetime.now().time()))

def callbacks():
    return [MyCustomCallbak()]
```
For the method `on_train_end`, the master will exeute it after all tasks have been completed.

workingloong:
The issue has been discussed in #1106

Issue #1106, opened 2019-09-03
Support custom callbacks


terrytangyuan:
Currently ElasticDL supports custom processing logic for prediction outputs via `PredictionOutputsProcessor` provided in model definition file. However, we may want to make this more flexible so that users can implement custom callbacks during different phases of ElasticDL job. For example, `PredictionOutputsProcessor` can be seen as a callback that is executed after prediction outputs are made. Another use case is to run some custom logic before training starts https://github.com/wangkuiyi/elasticdl/issues/1039.

Some references:
* [Keras callbacks](https://www.tensorflow.org/beta/guide/keras/custom_callback)
* [SessionRunHook](https://www.tensorflow.org/api_docs/python/tf/train/SessionRunHook) that can be used in [EstimatorSpec](https://www.tensorflow.org/api_docs/python/tf/estimator/EstimatorSpec)

chunyang-wen:
Should we create a callback class for ElasticDL? In ALPS, we use `TrainerHook` which is similar to `tf.keras.CallBack`.  In ElasticDL's worker, we have:
+ `training_process`
+ `evaluation_process`
+ `prediction_process`

And there may be other places we should add support for callback such as:
+ `on_train_begin`
+ `on_train_end`
+ `on_batch_begin`
+ `on_batch_end`

The places where users can insert their logics can vary as our user number increases and we support more use cases.

So Let's start early.

terrytangyuan:
Yes, I think this interface should be similar to `tf.keras.CallBack` so users can implement it in model definition files in a familiar keras style. 
Some use cases we observed that we want to support are (we can add more later on):
* Callbacks used in testing that checks retry times and model parameter updates: https://github.com/sql-machine-learning/elasticdl/pull/1229
* Callback similar to `PredictionOutputsProcessor` that is executed after prediction outputs are made. 
* Callback to write additional summaries to TensorBoard after we completed each evaluation job, e.g. currently we are writing TensorBoard summaries in `EvaluationService.complete_task()`.
* Callback to modulate learning rate as we are currently doing in `MasterServicer._modulate_lr_if_needed`.
* Callback for logic to be executed before training starts such as starting embedding service.
* Callback to perform early stopping when certain conditions are met.

cc: @chunyang-wen

terrytangyuan:
[...]

PR #1693, opened 2020-02-03
Add tf.keras.callbacks definition in model_def

workingloong:
Implement #1690

### DIFF
--- a/elasticdl/python/master/master.py
+++ b/elasticdl/python/master/master.py
@@ -23,6 +23,7 @@ from elasticdl.python.common.model_utils import (
     get_module_file_path,
     load_model_from_module,
     load_module,
+    load_callbacks_from_module,
 )
 from elasticdl.python.data.reader.data_reader_factory import create_data_reader
 from elasticdl.python.master.evaluation_service import EvaluationService
@@ -95,6 +96,9 @@ class Master(object):
         ).__dict__
         self.model_inst = load_model_from_module(
             args.model_def, self.model_module, args.model_params
+        )
+        self.callbacks = load_callbacks_from_module(
+            args.callbacks, self.model_module
         )
         model_handler = ModelHandler.get_model_handler(
             args.distribution_strategy, checkpoint_dir=args.checkpoint_dir
@@ -184,6 +188,8 @@ class Master(object):
         try:
             while True:
                 if self.task_d.finished():
+                    for callback in self.callbacks:
+                        callback.on_train_end()
                     if self.instance_manager:
                         self.instance_manager.update_status(
                             InstanceManagerStatus.FINISHED
@@ -229,6 +235,7 @@ class Master(object):
                     )
                     break
         self.logger.info("Master stopped")
+
 
     @staticmethod
     def _get_job_type(args):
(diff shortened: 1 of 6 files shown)


elasticdl/python/master/master.py
```diff
@@ -184,6 +190,8 @@ def run(self):
         try:
             while True:
                 if self.task_d.finished():
+                    for callback in self.callbacks:
+                        callback.on_train_end()
```
brightcoder01: From this code, this callback is executed on the master pod.
How about callback to upload the exported model to remote storage, do master or worker execute it?
We may consider that the model is exported by the worker.

workingloong: "How about callback to upload the exported model to remote storage, do master or worker execute it?"
The master will tar the model and upload.

Now, the model is exported from checkpoints by the worker.  If we want to execute callbacks with fault-tolerant,  the master should create a task for each callback on the epoch end and train/test end.

brightcoder01: I assume you mean that the model is exported by the worker, and then the master will compress and upload the model to the remote storage.
I'm wondering how master accesses the exported model from worker? Maybe shared storage in glusterfs?

How about letting one single pod do the model export and upload work? We can compare these two options.

workingloong: I have removed the code about callback execution on the master and only remain the callback definition. The callback execution needs further consideration for fault-tolerant.

terrytangyuan (commented):
It's not clear how the tf.keras.callbacks will affect different phases of ElasticDL though. We might want to design more carefully. See https://github.com/sql-machine-learning/elasticdl/issues/1106 for related discussions.

terrytangyuan:
Related issue: #1106

workingloong:
> It's not clear how the tf.keras.callbacks will affect different phases of ElasticDL though. We might want to design more carefully. See #1106 for related discussions.

There are many methods in tf.keras.callbacks.  https://www.tensorflow.org/guide/keras/custom_callback#an_overview_of_callback_methods
```python
on_(train|test|predict)_begin(self, logs=None)
on_(train|test|predict)_end(self, logs=None)
on_(train|test|predict)_batch_begin(self, batch, logs=None)
on_(train|test|predict)_batch_end(self, batch, logs=None)

on_epoch_begin(self, epoch, logs=None)
on_epoch_end(self, epoch, logs=None)
```
It may be difficult to design all the methods executed in ElasticDL. The methods for different phase is independent, so we can split those methods and design them for each phase. Firstly, we can design the methods for training phase and design the execution in ElasticDL.

elasticdl/python/master/master.py
workingloong: > I assume you mean that the model is exported by the worker, and then the master will compress and upload the model to the remote storage.
> I'm wondering how master accesses the exported model from worker? Maybe shared storage in glusterfs?
> 
> How about letting one single pod do the model export and upload work? We can compare these two options.

The model is exported to glusterfs by the worker and the `save_model_path` is sent to the worker by the master. So the master can access the exported model by `save_model_path`. Before, we assign a saved_model task to a worker because we think the memory of a worker is enough to load a model.  I feel like it is complicated to assign a single pod to execute the method of the callback and some callback must be called by the master. Then we can discuss the problem about calling the methods of callback in  #1706.

workingloong:
I close the PR  and will submit again after the design to support callbacks is merged. We can discuss the design on #1706.

PR #1694, opened 2020-02-03
add PullVariable/PullEmbeddingVector rpc in go

### LINKED ISSUE #1654: Implement PullEmbeddingVector grpc service in Go

### LINKED ISSUE #1655: Implement PullVariable grpc service in Go

QiJune:
Fix #1654 
Fix #1655

### DIFF
--- a/elasticdl/pkg/common/embedding_table.go
+++ b/elasticdl/pkg/common/embedding_table.go
@@ -26,8 +26,7 @@ func (table *EmbeddingTable) GetEmbeddingVector(index int64) *Tensor {
 		return value
 	}
 	// TODO(qijun) only support zero initializer now
-	newVector := NewVector(table.Dim)
-	newVector.Name = table.Name
+	newVector := NewVector(table.Dim, table.Name)
 	table.EmbeddingVector[index] = newVector
 	return newVector
 }
@@ -35,8 +34,7 @@ func (table *EmbeddingTable) GetEmbeddingVector(index int64) *Tensor {
 // GetEmbeddingVectors returns embedding vectors giving an array of indices
 func (table *EmbeddingTable) GetEmbeddingVectors(indices []int64) *Tensor {
 	d := []int64{int64(len(indices)), table.Dim}
-	t := NewTensor(d)
-	t.Name = table.Name
+	t := NewTensor(d, table.Name)
 	t.Indices = indices
 	for i, index := range indices {
 		start := int64(i) * table.Dim
@@ -51,7 +49,7 @@ func (table *EmbeddingTable) SetEmbeddingVectors(indices []int64, value []float3
 		return fmt.Errorf("Embedding vectors dim not match")
 	}
 	for i, index := range indices {
-		table.EmbeddingVector[index] = NewVector(table.Dim)
+		table.EmbeddingVector[index] = NewVector(table.Dim, table.Name)
 		start := int64(i) * table.Dim
 		copy(table.EmbeddingVector[index].Value, value[start:start+table.Dim])
 	}
--- a/elasticdl/pkg/common/embedding_table.go
+++ b/elasticdl/pkg/common/embedding_table.go
@@ -27,6 +27,7 @@ func (table *EmbeddingTable) GetEmbeddingVector(index int64) *Tensor {
 	}
 	// TODO(qijun) only support zero initializer now
 	newVector := NewVector(table.Dim)
+	newVector.Name = table.Name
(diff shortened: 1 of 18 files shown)


elasticdl/pkg/common/embedding_table.go
```diff
@@ -27,6 +27,7 @@ func (table *EmbeddingTable) GetEmbeddingVector(index int64) *Tensor {
 	}
 	// TODO(qijun) only support zero initializer now
 	newVector := NewVector(table.Dim)
+	newVector.Name = table.Name
```
wangkuiyi: How about `newVector := NewVecotr(table.Dim, table.Name)`?

elasticdl/pkg/ps/server.go
```diff
@@ -2,52 +2,110 @@ package ps
...
+	"sync"
 )
 
-type psServer struct {
+// Server defines servicer of ps
+type Server struct {
 	pb.PserverServer
+	Param *Parameter
+	Opt   Optimizer
+	PSID  int32
```
wangkuiyi: PSID => ID, because the fullname is `ps.Server.ID`

elasticdl/pkg/ps/server.go
```diff
@@ -2,52 +2,110 @@ package ps
...
+	lock  *sync.Mutex
 }
 
-func (s *psServer) PullVariable(ctx context.Context, in *pb.PullVariableRequest) (*pb.PullVariableResponse, error) {
-	// TODO: implement the service.
-	return &pb.PullVariableResponse{}, nil
+// NewServer creates a Server instance
+func NewServer(PSID int32, opt string, lr float32) *Server {
```
wangkuiyi: Need a comment about ID (PSID) -- does it have to be a zero-based successive integer number, or any number, e.g., an MD5 hash value? If it is the latter case, users should not have to specify it in the NewServer function.

elasticdl/pkg/ps/server_test.go
```diff
@@ -22,61 +27,171 @@ func createClient() (pb.PserverClient, context.Context, *grpc.ClientConn, contex
...
-		t.Errorf("Failed to pull variable")
+		t.Errorf("Failed to push model")
 	}
+
+	assert.True(t, s.Param.InitStatus)
+	assert.Len(t, s.Param.NonEmbeddingParam, 2)
+	assert.Contains(t, s.Param.NonEmbeddingParam, "t1")
+	assert.Contains(t, s.Param.NonEmbeddingParam, "t2")
+	assert.True(t, common.CompareFloatArray(a, s.Param.GetNonEmbeddingParam("t1").Value, 0.001))
+	assert.True(t, common.CompareFloatArray(b, s.Param.GetNonEmbeddingParam("t2").Value, 0.001))
```
mhaoli, line 63: Do we need to check parameter version here?

elasticdl/pkg/ps/server.go
```diff
@@ -2,52 +2,110 @@ package ps
...
+	if !s.Param.InitStatus {
+		res.ModelInitStatus = false
+		return &res, nil
+	}
+	res.ModelInitStatus = true
+	res.Model = &pb.Model{}
+	res.Model.Version = s.Param.Version
+	if s.Param.Version > in.CurrentModelVersion {
+		for _, v := range s.Param.NonEmbeddingParam {
+			fmt.Println(v.Value)
```
mhaoli: Remove fmt.Println here?

QiJune: Done

elasticdl/pkg/ps/server.go
QiJune: Done, it's the former case,  add comments.

elasticdl/pkg/ps/server.go
QiJune: Done

elasticdl/pkg/common/embedding_table.go
QiJune: Done

elasticdl/pkg/ps/server_test.go
QiJune: The model version in PS is directly set to the version in request. 

The key point of PushModel rpc is actually `InitStatus`. Several workers may push model to one PS, and once the model from one worker is  accepted, the others will be skipped.

We  have check the `InitStatus` already in this unit test.

mhaoli (approved):

PR #1695, opened 2020-02-03
get optimizer type and arguments in master for go ps

### LINKED ISSUE #1651: Master extracts optimizer type and related hyper-parameters (learning rate etc) from python code and sends them to Go PS as arguments

skydoorkai:
fix #1651

### DIFF
--- a/elasticdl/python/common/model_utils.py
+++ b/elasticdl/python/common/model_utils.py
@@ -187,6 +187,7 @@ def get_optimizer_info(optimizer):
     """
     Get optimizer type and its argument values
     """
+    # TODO: add more supported optimizers
     OPT_ARGUMENTS = {
         "SGD": ["learning_rate", "momentum", "nesterov"],
         "Adam": ["learning_rate", "beta_1", "beta_2", "epsilon", "amsgrad"],
--- a/elasticdl/python/common/model_utils.py
+++ b/elasticdl/python/common/model_utils.py
@@ -1,6 +1,8 @@
 import importlib.util
 import os
 
+import tensorflow as tf
+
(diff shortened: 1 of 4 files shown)


elasticdl/python/common/model_utils.py
```diff
@@ -179,3 +181,30 @@ def get_non_embedding_trainable_vars(model, embedding_layers):
...
             non_embedding_trainable_vars.append(var)
     return non_embedding_trainable_vars
+
+
+def get_optimizer_info(optimizer):
+    """
+    Get optimizer type and its argument values
+    """
+    # TODO: add more supported optimizers
+    OPT_ARGUMENTS = {
```
terrytangyuan, line 191: This is very hard to maintain though. Also if TF introduces new optimizers, we have to add them here. Backwards compatibility is also hard to guarantee.

skydoorkai: This is just an interface for TF. We will maintain this interface with the tf version we supported in Dockerfile. 
All optimizers are standard computations. For example, TF SGD optimizer contains 2 optimizers,  standard SGD and Nesterov variation, depending on `nesterov` argument. 
TF Adam optimizer also contains 2 optimizers, standard Adam and its `amsgrad` variation.

In Golang PS, we will implement these 4  optimizers, and select one depending on the input arguments. In this way, our Golang PS can support other ML frameworks such as PyTorch, as long as the optimizer type and arguments are given.

For backwards compatibility, for the interface, we only support the TF version used in Dockerfile. For optimizer computation, these are standard algorithms used by all ML frameworks.

QiJune (approved):
Excellent

Issue #1664, opened 2020-01-19
Design SQLFlow syntax extension for data transform.


brightcoder01:
The root of the discussion series is #1670 

The following transform functions are common used. We can support these in the first stage.

| Name             |    Transformation                                     | Statitical Parameter | Input Type | Output Type |
|:----------------:|:-----------------------------------------------------:|:--------------------:|:-------------------:|:---------------------:|
| NORMALIZE(x)     | Scale the inputs to the range [0, 1]. `out = x - x_min / (x_max - x_min)` |     x_min, x_max     | number | float64 |
| STANDARDIZE(x)   | Scale the inputs to z-score subtracts out the mean and divides by standard deviation. `out = x - x_mean / x_stddev` |    x_mean, x_stddev  | number | float64 |
| BUCKETIZE(x, num_buckets, boundaries) | Transform the numeric features into categorical ids using a set of thresholds. |   boundaries  | Number | int64 |
| HASH_BUCKET(x, hash_bucket_size) | Map the inputs into a finite number of buckets by hashing. `out_id = Hash(input_feature) % bucket_size` |   hash_bucket_size   | string, int32, int64 | int64 |
| VOCABULARIZE(x)  | Map the inputs to integer ids by looking up the vocabulary | vocabulary_list | string, int32, int64 | int64 |
| EMBEDDING(x, dimension) | Map the inputs to embedding vectors             |          N/A         | int32, int64 | float32 |
| CROSS(x1, x2, ..., xn, hash_bucket_size)  | Hash(cartesian product of features) % hash_bucket_size |          N/A         | string, number   | int64 |
| CONCAT(x1, x2, ..., xn)   | Concatenate multiple tensors representing categorical ids into one tensor. | N/A | int32, int64 | int64 |

There are three options for the style of the generated transform code:
1. Feature Column API. Integrate it with model definition using tf.keras.layers.DenseFeatures;
2. Customized Keras Layer provided from ElasticDL. The functionality should cover all the common used feature engineering operations above;
3. [Keras Preprocess Layer](https://github.com/tensorflow/community/pull/188). This will be ready in TF2.2;

brightcoder01:
[...]

Issue #1667, opened 2020-01-19
Generate data analysis SQL using code_gen from data transform expression in SQLFlow.


brightcoder01:
The following transform functions contains analysis. The analysis work should be done at first to make the transform logic concrete.

Name | Feature Column Template | Analyzer
-- | -- | --
STANDARDIZE(x) | numeric_column({var_name}, normalizer_fn=lambda x : x - {mean} / {std}) | MEAN, STDDEV
NORMALIZE(x) | numeric_column({var_name}, normalizer_fn=lambda x : x - {min} / {max} - {min}) | MAX, MIN
BUCKETIZE(x, bucket_num=y) | bucketized_column({var_name}, boundaries={percentiles}) | PERCENTILE
APPLY_VOCAB(x) | categorical_column_with_vocabulary({var_name}, vocabulary_list={vocabulary_list}) | DISTINCT
HASH(x, hash_bucket_size) | categorical_column_with_hash_bucket | COUNT(DISTINCT)

The SQLFlow syntax for data transform and analysis is discussed in #1664

brightcoder01:
Let's take the following SQLFlow statement for example:
```SQL
SELECT *
FROM census_income
TO TRAIN DNNClassifier
WITH model.hidden_units = [10, 20]
COLUMN (
    NUMERIC(NORMALIZE(capital_gain)), 
    NUMERIC(STANDARDIZE(age)), 
    EMBEDDING(BUCKETIZE(hours_per_week, bucket_num=5), dim=32),
    EMBEDDING(APPLY_VOCAB(occupation), dim=16),
    EMBEDDING(HASH(workclass), dim=8)
LABEL label
```

The generated analysis SQL in MaxCompute is:

*Calculate the min&max of `capital_gain`, mean&stddev of `age`, the recommended hash_bucket_size for `workclass`*
```SQL
SELECT 
    MIN(capital_gain) AS _capital_gain_min_, 
    MAX(capital_gain) AS _capital_gain_max_, 
    AVG(age) AS _age_mean_, 
    STDDEV(age) AS _age_stddev_,
    (COUNT(DISTINCT(workclass)) * 3) AS _workclass_hash_bucket_num_
FROM census_income;
```

*Calculate the bucket boundary of `hours_per_week`*
```SQL
SELECT 
    percentile(hours_per_week, 0.2) AS _hours_per_week_bkt_boundry_1_,
[...]

PR #1697, opened 2020-02-04
Add the DNN Model & WideAndDeep Model using census dataset for SQLFlow syntax discussion.

brightcoder01:
Relate #1664 

The dataset is the [census income dataset](https://archive.ics.uci.edu/ml/datasets/census+income).


model_zoo/census_model_sqlflow/dnn/census_dnn.sql
```diff
@@ -0,0 +1,19 @@
+SELECT *
+FROM census_income
+TO TRAIN DNNClassifier
+WITH model.hidden_units = [10, 20]
+COLUMN (
+    age, 
+    capital_gain, 
+    capital_loss, 
+    hours_per_week, 
+    EMBEDDING(HASH(workclass, 64), 16),
```
wangkuiyi: I think hash is part of the embedding operation. How about we define only one data transform function `embedding`:

```sql
COLUMN
    embedding(workclass, 64, 16),
...
```

model_zoo/census_model_sqlflow/wide_and_deep/census_wide_and_deep.sql
```diff
@@ -0,0 +1,12 @@
+SELECT *
+FROM census_income
+TO TRAIN WideAndDeepClassifier
+COLUMN (
+    SET GROUP(APPLY_VOCAB(workclass), BUCKETIZE(capital_gain, bucket_num=5), BUCKETIZE(capital_loss, bucket_num=5), BUCKTIZE(hours_per_week, bucket_num=6)) AS group_1,
```
wangkuiyi: SET means value assignment here -- it is a very different philosophy from the functional programming style we have been used in the design of SQLFlow syntax. Can we don't use SET here?

@typhoonzero @shendiaomo Do you have suggestions here?

model_zoo/census_model_sqlflow/dnn/census_dnn.sql
brightcoder01: Sometimes user will call `lookup vocabulary` to convert workclass from string to id, and then convert the id to an embedding vector.
Maybe we need add a flag to differentiate `hash` or `lookup vocabulary` for the id conversion.

typhoonzero: We actually have https://www.tensorflow.org/api_docs/python/tf/feature_column/categorical_column_with_hash_bucket as a feature column. The problem is how to calculate the bucket size (64 here) using some SQL statement and pass it here.

model_zoo/census_model_sqlflow/wide_and_deep/census_wide_and_deep.sql
typhoonzero: `SET` can be used in three ways:

1. user-defined variables
2. setting variables in procedures
3. setting system variables like charset.

see: https://dev.mysql.com/doc/refman/5.7/en/set-variable.html, https://dev.mysql.com/doc/refman/5.7/en/user-variables.html

In this case, I assume that you want to use `SET` to set a user-defined variable, and the variable is a list of feature columns. For a user-defined variable, it can be set in one SQL statement and referred to in another SQL statement. Then it should look like: 

```sql
SET group_1 = GROUP(APPLY_VOCAB(workclass), ...)
SELECT * .... COLUMN EMBEDDING(group1) ...
```

Then the statement `SET group_1 = GROUP(APPLY_VOCAB(workclass), ...)` will be passed to the third-party parser and it will report parsing error. And, implement the whole `SET` grammar parsing inside `COLUMN` does not make sense either.

I recommend the grammar should look like:

```sql
COLUMN
    EMBEDDING(
        CONCAT(
            CATEGORY_ID_HASH(col1, 64),
            CATEGORY_ID_VOCAB(col2),
            col3,
        )
    ) FOR linear_feature_columns
COLUMN .... FOR dnn_feature_columns
```

typhoonzero: NOTE, the feature column `CATEGORY_ID_VOCAB` can be omitted since the vocabulary of the column can be done by feature derivation.

model_zoo/census_model_sqlflow/dnn/census_dnn.sql
brightcoder01: We can use the following SQL statement to calculate the hash_bucket_size for `workclass`. `3` is  a configurable factor to decrease the conflict of the hash function.

```SQL
SELECT 
    (COUNT(DISTINCT(workclass)) * 3) AS _workclass_hash_bucket_num_
FROM census_income;
```

If a user writes `HASH` function, the `hash_bucket_size` is optional. If it's missing, SQLFlow will run the analysis SQL above to calculate the `hash_bucket_size` value.

For more details for analysis SQL generation, we can see the discussion issue #1667

model_zoo/census_model_sqlflow/wide_and_deep/census_wide_and_deep.sql
brightcoder01: > NOTE, the feature column `CATEGORY_ID_VOCAB` can be omitted since the vocabulary of the column can be done by feature derivation.

I curious about how we calculate the vocabulary list inside feature derivation?

I believe we need do a statistic work using SQL to calculate the vocabulary just like [this](https://github.com/sql-machine-learning/elasticdl/issues/1667#issuecomment-580044742)

brightcoder01: > ```sql
> COLUMN
>     EMBEDDING(
>         CONCAT(
>             CATEGORY_ID_HASH(col1, 64),
>             CATEGORY_ID_VOCAB(col2),
>             col3,
>         )
>     ) FOR linear_feature_columns
> ```

I assume that the suggestion is to write two COLUMN clauses - one for linear/wide, another for deep, we can see there are many transform expression in common between these two.
I'm afraid the following long clauses will be written twice. It will be user unfriendly if there are hundreds of columns which is a common scenario for search&recommendation:

```
GROUP(APPLY_VOCAB(workclass), BUCKETIZE(capital_gain, bucket_num=5), BUCKETIZE(capital_loss, bucket_num=5), BUCKTIZE(hours_per_week, bucket_num=6)),
GROUP(HASH(education), HASH(occupation), APPLY_VOCAB(martial_status), APPLY_VOCAB(relationship)),
GROUP(BUCKETIZE(age, bucket_num=5), HASH(native_country), APPLY_VOCAB(race), APPLY_VOCAB(sex)),
```

Using `SET` or other equivalent expression is to make the common transform clause which `group1`, `group2`, `group3` standards for, can be reused by linear/wide and deep.

workingloong: Maybe we use a nested COLUMN syntax to implement complicated transformation like
```python
SELECT *
FROM census_income
TO TRAIN WideAndDeepClassifier
COLUMNS (
    [EMBEDDING(group1, 1), EMBEDDING(group2, 1)] AS wide_embeddings
    [EMBEDDING(group1, 8), EMBEDDING(group2, 8), EMBEDDING(group3, 8)] AS deep_embeddings
    FROM(
        CONCAT(APPLY_VOCAB(workclass), BUCKETIZE(capital_gain, bucket_num=5), BUCKETIZE(capital_loss, bucket_num=5), BUCKTIZE(hours_per_week, bucket_num=6)) AS group_1,
        CONCAT(HASH(education), HASH(occupation), APPLY_VOCAB(martial_status), APPLY_VOCAB(relationship)) AS group_2,
        CONCAT(BUCKETIZE(age, bucket_num=5), HASH(native_country), APPLY_VOCAB(race), APPLY_VOCAB(sex)) AS group_3,
    )
)
LABEL label
```
Then, we can parse the COLUMN expression in order.

workingloong: We also can place embedding in the model definition and users don't need to write Embedding.
```sql
SELECT *
FROM census_income
TO TRAIN WideAndDeepClassifier
COLUMNS (
    [group1, group2] AS wide
    [group1, group2, group3] AS deep
    FROM(
        CONCAT(APPLY_VOCAB(workclass), BUCKETIZE(capital_gain, bucket_num=5), BUCKETIZE(capital_loss, bucket_num=5), BUCKTIZE(hours_per_week, bucket_num=6)) AS group_1,
        CONCAT(HASH(education), HASH(occupation), APPLY_VOCAB(martial_status), APPLY_VOCAB(relationship)) AS group_2,
        CONCAT(BUCKETIZE(age, bucket_num=5), HASH(native_country), APPLY_VOCAB(race), APPLY_VOCAB(sex)) AS group_3,
    )
)
LABEL label
```
The model definition is 
https://github.com/sql-machine-learning/elasticdl/blob/151184859989452e7105de100c5478e1770031c4/model_zoo/census_wide_deep_model/wide_deep_functional_api.py#L147

model_zoo/census_model_sqlflow/dnn/census_feature_column.py
```diff
@@ -0,0 +1,54 @@
...
+
+def get_feature_columns():
+    feature_columns = []
+
+    for numeric_feature_key in NUMERIC_FEATURE_KEYS:
+        numeric_feature = tf.feature_column.numeric_column(numeric_feature_key)
+        feature_columns.append(numeric_feature)
+
+    for categorical_feature_key in CATEGORICAL_FEATURE_KEYS:
+        embedding_feature = tf.feature_column.embedding_column(
```
terrytangyuan, line 30: Will all of these be automatically generated by the SQL statement?

model_zoo/census_model_sqlflow/dnn/census_dnn.sql
wangkuiyi: @brightcoder01 Let us use `embedding(workclass, 64, 16)` instead of `embedding(hash(workclass, 64), 16)`. If `embedding` creates a `map[int]Vector`, it can call `hash` to convert `workclass` into the integer ID.

model_zoo/census_model_sqlflow/dnn/census_feature_column.py
wangkuiyi: I have the same question. @brightcoder01 Why do you add some Python source files in this PR? How would these Python files come?

brightcoder01: Yes, the python file is the corresponding generated code from SQL statement.

terrytangyuan: Got it. Sounds good to me then.

model_zoo/census_model_sqlflow/dnn/census_dnn.sql
brightcoder01: > @brightcoder01 Let us use `embedding(workclass, 64, 16)` instead of `embedding(hash(workclass, 64), 16)`. If `embedding` creates a `map[int]Vector`, it can call `hash` to convert `workclass` into the integer ID.

Good idea. The SQLFlow expression describes users' intention and don't need expose the details about how to convert string to id. We will use embedding(workclass, dimension=16) directly.

model_zoo/census_model_sqlflow/wide_and_deep/census_wide_and_deep.sql
brightcoder01: After discussion with @typhoonzero @workingloong @shendiaomo , We will rename `GROUP` to `CONCAT`.

One key point is how to reuse the common transform logic in `COLUMN` clause. For example, the following clause will be shared by both wide embedding(dimension=1) and deep embedding(dimension=8). We want to write these clauses just once instead of twice.
```SQL
CONCAT(APPLY_VOCAB(workclass), BUCKETIZE(capital_gain, bucket_num=5), BUCKETIZE(capital_loss, bucket_num=5), BUCKTIZE(hours_per_week, bucket_num=6)),
CONCAT(HASH(education), HASH(occupation), APPLY_VOCAB(martial_status), APPLY_VOCAB(relationship)),
```
We have three options of syntax design for this SQL statement.

```SQL
SELECT *
FROM census_income
TO TRAIN WideAndDeepClassifier
COLUMN (
    SET CONCAT(APPLY_VOCAB(workclass), BUCKETIZE(capital_gain, bucket_num=5), BUCKETIZE(capital_loss, bucket_num=5), BUCKTIZE(hours_per_week, bucket_num=6)) AS group_1,
    SET CONCAT(HASH(education), HASH(occupation), APPLY_VOCAB(martial_status), APPLY_VOCAB(relationship)) AS group_2,
    SET CONCAT(BUCKETIZE(age, bucket_num=5), HASH(native_country), APPLY_VOCAB(race), APPLY_VOCAB(sex)) AS group_3,

    (EMBEDDING(group1, 1), EMBEDDING(group2, 1)) AS wide_embeddings
    (EMBEDDING(group1, 8), EMBEDDING(group2, 8), EMBEDDING(group3, 8)) AS deep_embeddings
)
LABEL label
```

```SQL
SELECT *
FROM census_income
TO TRAIN WideAndDeepClassifier
COLUMN (
    (EMBEDDING(group1, 1), EMBEDDING(group2, 1)) AS wide_embeddings
    (EMBEDDING(group1, 8), EMBEDDING(group2, 8), EMBEDDING(group3, 8)) AS deep_embeddings
    WITH (
        CONCAT(APPLY_VOCAB(workclass), BUCKETIZE(capital_gain, bucket_num=5), BUCKETIZE(capital_loss, bucket_num=5), BUCKTIZE(hours_per_week, bucket_num=6)) AS group_1,
        CONCAT(HASH(education), HASH(occupation), APPLY_VOCAB(martial_status), APPLY_VOCAB(relationship)) AS group_2,
        CONCAT(BUCKETIZE(age, bucket_num=5), HASH(native_country), APPLY_VOCAB(race), APPLY_VOCAB(sex)) AS group_3,
    )
)
LABEL label
```

```SQL
SELECT *
FROM census_income
TO TRAIN WideAndDeepClassifier
COLUMN
    EMBEDDING(CONCAT(APPLY_VOCAB(workclass), BUCKETIZE(capital_gain, bucket_num=5), BUCKETIZE(capital_loss, bucket_num=5), BUCKTIZE(hours_per_week, bucket_num=6)) AS group_1, 8),
    EMBEDDING(CONCAT(HASH(education), HASH(occupation), APPLY_VOCAB(martial_status), APPLY_VOCAB(relationship)) AS group_2, 8),
    EMBEDDING(CONCAT(BUCKETIZE(age, bucket_num=5), HASH(native_country), APPLY_VOCAB(race), APPLY_VOCAB(sex)) AS group_3, 8)
    FOR deep_embeddings
COLUMN
    EMBEDDING(group1, 1),
    EMBEDDING(group2, 1)
    FOR wide_embeddings
LABEL label
```

terrytangyuan: I'd vote for the second option.

brightcoder01: After more discussion with @wangkuiyi , the third option is more inline with the SQL style. We will choose this design.

```SQL
SELECT *
FROM census_income
TO TRAIN WideAndDeepClassifier
COLUMN
    EMBEDDING(CONCAT(APPLY_VOCAB(workclass), BUCKETIZE(capital_gain, bucket_num=5), BUCKETIZE(capital_loss, bucket_num=5), BUCKTIZE(hours_per_week, bucket_num=6)) AS group_1, 8),
    EMBEDDING(CONCAT(HASH(education), HASH(occupation), APPLY_VOCAB(martial_status), APPLY_VOCAB(relationship)) AS group_2, 8),
    EMBEDDING(CONCAT(BUCKETIZE(age, bucket_num=5), HASH(native_country), APPLY_VOCAB(race), APPLY_VOCAB(sex)) AS group_3, 8)
    FOR deep_embeddings
COLUMN
    EMBEDDING(group1, 1),
    EMBEDDING(group2, 1)
    FOR wide_embeddings
LABEL label
```

terrytangyuan (approved):

PR #1700, opened 2020-02-04
rewrite data structure 

lwpyr:
* support generic tensor data type


elasticdl/pkg/common/tensor.go
```diff
@@ -1,84 +1,305 @@
...
+
+// ZeroInit return a closure
+func ZeroInit(length int, dtype DataType) InitialzeFunc {
+	return func(ptr unsafe.Pointer) {
+		C.memset(ptr, 0, C.ulong(length*int(dtype.Size)))
+	}
+}
+
+// NewEmptyVector create an zero-initialized vector
+func NewEmptyVector(length int, dtype DataType, initializer InitialzeFunc) *Vector {
```
wangkuiyi, line 81: Can we use Go's memory allocation instead of C's malloc to create vectors here?

QiJune:
First, this PR is too large, we could seperate it into several small PRs.

Second, we should clarify following three concepts:

- `Buffer`
- `Tensor`
- `Vector`

`Buffer` is a struct contains a pointer to some contiguous memory, its size, and its type.

```c
struct Buffer {
    void* data,
    size_t len,
    size_t dtype
};
```

`Tensor` is a `Buffer` with some extra information, such as Name, Dim and Indices.

`Vector` is actually a 1-dim Tensor.

We could design and implement the concept `Buffer` first.

PR #1706, opened 2020-02-05
Add design to support custom callback using Keras API

workingloong:
Related to #1106

### DIFF
--- a/docs/designs/keras_callback.md
+++ b/docs/designs/keras_callback.md
@@ -1,14 +1,13 @@
-# Design to Support Keras Callback
+# Design to Support Custom Callback Function Using Keras
 
 This document describes the design for supporting callback to customize the behavior of model during training, evaluation and inference in ElasticDL.
 
 ## Motivation
 
 In deep learning, we generally need to customize the behavior of model during training, evaluation or
-inference, including reading/changing the model. According to the call frequency, we may perform the behavior per batch, per epoch or at the start and end of job. `tf.keras.callbacks.Callback` is an abstract base class and has methods to perform the behavior at different call frequency, such as `on_bath_end`, `on_epoch_end` and so on. So, we adopt the interfaces of `tf.keras.callbacks.Callback` for users to customize the behavior of model in ElasticDL. 
+inference, including reading/changing the model. We may perform the behavior per batch, per epoch or at the start and end of job. `tf.keras.callbacks.Callback` is an abstract base class and has methods to perform the behavior at different call frequency, such as `on_bath_end`, `on_epoch_end` and so on. So, we adopt the interfaces of `tf.keras.callbacks.Callback` to customize the behavior of model in ElasticDL. 
 
-Now we have implemented some modules which are similar to callback in ElasticDL such as `LearingRateScheduler` and `PredictionOutputsProcessor`. And user should write definitions
-in the model definition file for each module like:
+Now we have implemented some modules similar to callback in ElasticDL, such as `LearingRateScheduler` and `PredictionOutputsProcessor`. And users should write definitions in the model definition file for each module like:
 ```python
 def custom_model():
     ...
@@ -26,56 +25,40 @@ def learning_rate_scheduler(model_version):
 class PredictionOutputsProcessor(BasePredictionOutputsProcessor):
     ...
 ```
-There will be multiple definition APIs for users to define different behaviors of model and the interfaces of APIs may be different. It is more convenient for users to define those behaviors using `tf.keras.callbacks.Callback`. 
+There will be multiple definition APIs for users to define different behaviors of model. The interfaces of APIs may be different. It is more convenient for users to define those behaviors using `tf.keras.callbacks.Callback`. 
 
 Some use cases we observed that we want to support are:
 * Case 1: Callback similar to `PredictionOutputsProcessor` that is executed after prediction outputs are made.
 * Case 2: Callback to modulate learning rate as we are currently doing in `LearningRateScheduler`.
-* Case 3: Callback to checkpoint model weights during training.
-* Case 4: Callback to write additional summaries to TensorBoard after each evaluation job completes.
-* Case 5: Callback to perform early stopping when certain conditions are met. For example, the metrics are met after an evaluation job.
-* Case 6: Callback to upload model to remote storage after completing training.
+* Case 3: Callback to write additional summaries to TensorBoard after each evaluation job completes.
+* Case 4: Callback to perform early stopping when certain conditions are met. For example, the metrics are met after an evaluation job.
+* Case 5: Callback to export model using SavedModel after the training is completed.
+* Case 6: Callback to upload model to remote storage after the training is completed.
 
-
-To implement callbacks, we need to solve the following problems:
-1. How to define callbacks for the model in ElasticDL.
-2. How to call the methods of callbacks in ElasticDL.
+The following, we will design how to define and implement callbacks to support those cases.
 
 ## Define Callbacks for the Model in ElasticDL
 
 In the model definition file, users can add `callbacks` API to define callbacks like:
 ```python
+From elasticdl.python.callbacks import (
+    PredictionOutputsProcessor,
+    LearningRateScheduler
+)
+
 def custom_model():
     ...
 
 def callbacks():
-    prediction_output_processor = PredictionOutputsProcessor()
-    learning_rate_scheduler = LearningRateScheduler()
+    prediction_output_processor = PredictionOutputsProcessor(process_fn)
+    learning_rate_scheduler = LearningRateScheduler(schedule_fn)
     return [prediction_output_processor, learning_rate_scheduler]
-
-
-class PredictionOutputsProcessor(tf.keras.callbacks.Callback):
-    def on_predict_batch_end(self, batch, logs=None):
-        predictions = logs["predictions"]
-        process(predictions)
-
-
-class LearningRateScheduler(tf.keras.callbacks.Callback):
-    def __init__(self, schedule):
-        super(LearningRateScheduler, self).__init__()
-        self.schedule = schedule
-
-    def on_train_batch_begin(self, model_version, logs=None):
-        if not hasattr(self.model.optimizer, 'lr'):
-            raise ValueError('Optimizer must have a "lr" attribute.')
-        lr = self.schedule(model_version)
-        K.set_value(self.model.optimizer.lr, lr)
 ```
 
 ## Initialize Callbacks and Set Callback Attributes in ElasticDL
 
 ### Use a Container for Callbacks Defined in the Model
-We may define several callbacks for the model in a job. Tensorflow creates a container `CallbackList` to wrap the callbacks to conveniently call the methods in callbacks. We can call the methods in callbacks by calling the methods in `CallbackList`. For example:
+We may define several callbacks for the model in a job. Tensorflow creates a container `CallbackList` to wrap the callbacks to conveniently call the methods in callbacks. For example:
 ```python
 class CallbackList():
     def on_batch_end(self, batch, logs=None):
@@ -95,7 +78,7 @@ callback_list.on_batch_end(batch)
 
 ### Set Default Attributes for Callbacks.
 There are `set_model` and `set_params` in `tf.keras.callbacks.Callback`. `set_model` can set a model object to the attribute `model` of callback and `set_params` can set a dict object to the attribute `params` of callback.
-We also can use those methods of `CallbackList` to set model and parameters for the callbacks in `CallbackList`. 
+We also can use those methods of `CallbackList` to set `model` and `params` for the callbacks in `CallbackList`. 
 ```python
 model = custom_model()
 callbacks = callbacks()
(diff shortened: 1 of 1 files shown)


docs/designs/keras_callback.md
```diff
@@ -0,0 +1,196 @@
+# Design to Support Keras Callback
+
+This document describes the design for supporting callback to customize the behavior of model in ElasticDL.
```
skydoorkai: Not customize the model, but the model training process.

workingloong: Not only the training process but also the prediction and evaluation process.
"Customize the behavior during the model iteration" ?

docs/designs/keras_callback.md
```diff
@@ -0,0 +1,196 @@
...
+So, we can also use CallbackList to wrap the callbacks in ElasticDL.
+```python
+from tensorflow.python.keras.callbacks import CallbackList
+callbacks = callbacks()
+callback_list = CallbackList(callbacks)
+callback_list.on_batch_end(batch)
+```
+
+
+### Call the Methods of Callback using ParameterServerStrategy
```
skydoorkai: We may separate the callbacks in workers and in PS.  Add something like `on_update_start`, `on_update_end` for PS.
This will make it more clear.

docs/designs/keras_callback.md
```diff
@@ -0,0 +1,196 @@
...
+
+Some use cases we observed that we want to support are:
+* Callbacks used in testing that checks retry times and model parameter updates,
+* Callback similar to `PredictionOutputsProcessor` that is executed after prediction outputs are made.
+* Callback to write additional summaries to TensorBoard after each evaluation job completes.
+* Callback to modulate learning rate as we are currently doing in `LearningRateScheduler`.
+* Callback for logic to be executed before training starts such as starting embedding service.
+* Callback to perform early stopping when certain conditions are met.
+* Callback to upload model to remote storage after completing training.
+
```
terrytangyuan, line 37: How about the current checkpointing strategy? Can they be implemented as callback as well?

docs/designs/keras_callback.md
```diff
@@ -0,0 +1,196 @@
...
+The worker and parameter server will both call the `on_train_batch_begin|end` and per batch. So, the 2 methods will be called twice and can be called by either worker or parameter server. For example, the `on_train_batch_begin` of `LearningRateScheduler` can only be called by parameter server to adjust the learning rate for each batch. There are 2 proposals to solve the problem.
+
+1. Only the parameter server calls `on_train_batch_begin|end` and the worker doesn't. Because, we 
+have not observed a use case for callback to execute per batch by the worker.
+
+2. ElasticDL sets an argmuent `instance_role` in `params` of each callback and users get the `instance_role` in `on_train_batch_begin|end`. It may need users to understand why to get the `instance_role`.
+```python
+class LearningRateScheduler(tf.keras.callbacks.Callback):
+    def on_train_batch_begin(self, batch, logs=None):
+        instance_role = self.params.get("instance_role", None)
```
terrytangyuan: Need to define `self.params` and outline what `__init__()` looks like

docs/designs/keras_callback.md
```diff
@@ -0,0 +1,196 @@
...
+        if instance_role != "ParameterServer":
```
terrytangyuan: Also need to define the possible instance roles in constant class

docs/designs/keras_callback.md
```diff
@@ -0,0 +1,196 @@
...
+    finally:
+        self.callback_list.on_(train|test|predict)_end(self, logs=None)
+        self._stop()
+    return self._exit_code
+```
+
+
+### Call the Methods of Callback using AllReduceStrategy
+
+The design to call the methods using AllReduceStrategy will be considered later.
```
terrytangyuan: This case would be even simpler - there are only master and workers.

docs/designs/keras_callback.md
workingloong: `on_update_start` and `on_update_end` are not in `tf.keras.callbacks.Callback`. But we can add those methods specified for PS.

docs/designs/keras_callback.md
workingloong: Yes, the checkpointing strategy can also be implemented as a callback which can call a method on the PS every updating gradients of a batch. Maybe we need `on_update_batch_end` to implement checkpointing strategy.

docs/designs/keras_callback.md
workingloong: Yeah, the worker can call all methods in `tf.keras.callbacks.Callback` using AllReduceStrategy.

docs/designs/keras_callback.md
workingloong: Maybe we can add a method named `on_update_batch_begin` specified for PS.

docs/designs/keras_callback.md
```diff
@@ -0,0 +1,208 @@
+# Design to Support Keras Callback
+
+This document describes the design for supporting callback to customize the behavior of model during training, evaluation and inference in ElasticDL.
+
+## Motivation
+
+In deep learning, we generally need to customize the behavior of model during training, evaluation or
```
brightcoder01, line 7: What's the meaning of `changing the model`?

docs/designs/keras_callback.md
```diff
@@ -0,0 +1,208 @@
...
+inference, including reading/changing the model. According to the call frequency, we may perform the behavior per batch, per epoch or at the start and end of job. `tf.keras.callbacks.Callback` is an abstract base class and has methods to perform the behavior at different call frequency, such as `on_bath_end`, `on_epoch_end` and so on. So, we adopt the interfaces of `tf.keras.callbacks.Callback` for users to customize the behavior of model in ElasticDL.
```
brightcoder01: IMO, we don't use `tf.keras.callbacks.Callback` directly?
Because we add more methods `on_update_batch_end` which are not from `tf.keras.callbacks.Callback`.

docs/designs/keras_callback.md
```diff
@@ -0,0 +1,208 @@
...
+from tensorflow.python.keras.callbacks import CallbackList
+callbacks = callbacks()
+callback_list = CallbackList(callbacks)
+callback_list.on_batch_end(batch)
+```
+
+
+### Call the Methods of Callback using ParameterServerStrategy
+
+Using ParameterServerStrategy in ElasticDL, we will launch instances as master, worker and parameter server.
```
brightcoder01: The Callback interface contains multiple methods.
How do we decide which methods execute on master or worker? From the method name?

docs/designs/keras_callback.md
```diff
@@ -0,0 +1,208 @@
...
+        self.callbacks_list.on_predict_batch_end(
+            batch=batch_index, logs=prediction_logs
+        )
+```
+
+### Call Methods of Callback Per Updating Batch Gradients on PS
+
+"Case 2" and "Case 3" needs to customize behaviors After updating model weights using batch gradients on the PS. So, we add `on_update_batch_end` method, though it is not in `tf.keras.callbacks.Callback`.
+```python
+def on_update_batch_end(self, version, logs=None):
```
brightcoder01: Are these two methods `on_update_batch_start|end` exposed to users?

docs/designs/keras_callback.md
workingloong: For example, users may check the model weights and change it. But, we don't have a use case of changing the model now.

docs/designs/keras_callback.md
```diff
@@ -0,0 +1,208 @@
...
+    early_stop_callback = EarlyStopCallback()
+    learning_rate_scheduler = LearningRateScheduler()
+    return [early_stop_callback, learning_rate_scheduler]
+
+
+class EarlyStopCallback(tf.keras.callbacks.Callback):
+    def on_batch_end(self, batch, logs=None):
+        ...
+
+class LearningRateScheduler(tf.keras.callbacks.Callback):
```
brightcoder01: Can we write a complete sample for the class `LearningRateScheduler` here? I'm curious about how we get the `model_version` from the parameters in the method `on_batch_start(self, batch, logs=None)`
```
def learning_rate_scheduler(model_version):
    if model_version < 5000:
        return 0.0003
    elif model_version < 12000:
        return 0.0002
    else:
        return 0.0001
```

docs/designs/keras_callback.md
workingloong: We can add a method in the subclass of `tf.keras.callbacks.Callback` like:
```python
class TestCallback(tf.keras.callbacks.Callback):
    def on_update_batch_end(self, batch, logs=None):
         ...
```
Although `on_update_batch_end` is not from `tf.keras.callbacks.Callback`, but we can also call the method independently on the PS not using `CallbacksList`.
However, due to we are developing a high-performance PS using Golang, I wonder whether or not we need to use `tf.keras.callbacks.Callback` to define the method for PS. There is a question of how PS using Golang calls the method of callbacks defined using python.

docs/designs/keras_callback.md
workingloong: Not from the method name but call frequency. The worker call methods per batch and the master call methods per epoch and job.

docs/designs/keras_callback.md
workingloong:
```
import tensorflow as tf
from tensorflow.python.keras import backend as K

def schedule(model_version):
    if model_version < 5000:
        return 0.0003
    elif model_version < 12000:
        return 0.0002
    else:
        return 0.0001

class LearningRateScheduler(tf.keras.callbacks.Callback):
    def __init__(self, schedule, verbose=0):
        super(LearningRateScheduler, self).__init__()
        self.schedule = schedule

    def on_train_batch_begin(self, model_version, logs=None):
        if not hasattr(self.model.optimizer, 'lr'):
            raise ValueError('Optimizer must have a "lr" attribute.')
        lr = self.schedule(model_version)
        K.set_value(self.model.optimizer.lr, lr)
```

I will add the code snippet into the doc.

docs/designs/keras_callback.md
workingloong: We can expose the two methods to users. But I tend to we provide some premade callbacks class using the two methods like [Keras premade callback classes](https://www.tensorflow.org/api_docs/python/tf/keras/callbacks#classes).

docs/designs/keras_callback.md
```diff
@@ -0,0 +1,208 @@
...
+```
+There will be multiple definition APIs for users to define different behaviors of model and the interfaces of APIs may be different. It is more convenient for users to define those behaviors using `tf.keras.callbacks.Callback`. 
+
+Some use cases we observed that we want to support are:
+* Case 1: Callback similar to `PredictionOutputsProcessor` that is executed after prediction outputs are made.
+* Case 2: Callback to modulate learning rate as we are currently doing in `LearningRateScheduler`.
+* Case 3: Callback to checkpoint model weights during training.
+* Case 4: Callback to write additional summaries to TensorBoard after each evaluation job completes.
+* Case 5: Callback to perform early stopping when certain conditions are met. For example, the metrics are met after an evaluation job.
+* Case 6: Callback to upload model to remote storage after completing training.
```
brightcoder01: For the model uploading work, we need the model export path in glusterfs and the target path in oss. How can we get these parameters from the method `on_train_begin(self, logs=None)`?

docs/designs/keras_callback.md
brightcoder01: Can you please share a scenario of `check the model weights and change it` with the native keras?

workingloong: Emmm, i don't have a scenario. But i think changing the learning rate of optimizer belongs to changing the model.

docs/designs/keras_callback.md
workingloong: There are `model` and `params` attributes in `tf.keras.callbacks.Callback`. We can default set some arguments by [set_params](https://www.tensorflow.org/api_docs/python/tf/keras/callbacks/Callback#set_params) in `params` and user get it by `self.params`.

workingloong: For example:
```python
class UploadModelCallback(tf.keras.callbacks.Callback):
    def __init__(self, remote_url):
        self.remote_url = remote_url
        super(UploadModelCallback, self).__init__()

    def on_train_end(self, logs=None):
        saved_model_path = self.params["saved_model_path"]
        if(
            not os.path.exists(saved_model_path)
            or not os.listdir(saved_model_path)
        ):
            return
        upload(save_model_path, remote_url)

def callbacks():
    upload_model_callback = UploadModelCallback("gluaterfs//xxx")
```
After the training is completed we can call `on_train_end` method like
```python
upload_model_callback.on_train_end()
```

docs/designs/keras_callback.md
workingloong: Due to we are developing PS using Golang, the current checkpointing strategy can not be implemented as callback using ParameterServerStrategy.

docs/designs/keras_callback.md
workingloong: I have refreshed the design of `LearningRateScheduler`, you review it int the latest commit.

docs/designs/keras_callback.md
workingloong: In the new design, the PS does not call any method of callbacks. So, we will not add `on_update_batch_end`.

docs/designs/keras_callback.md
```diff
@@ -0,0 +1,299 @@
...
+        task_data_service: TaskDataService to process data according the task
+        dataset_fn: function to process dataset
+        model_handler: to transform the trained model with ElasticDL embedding layer to Keras native model.
+    """
+    def __init__(self, task_data_service, dataset_fn, model_handler):
+        self._model_handler = model_handler
+        self._task_data_service
+        self._dataset_fn = dataset_fn
+
+    def on_train_end(self, logs=None):
```
skydoorkai, line 137: So these are ElasticDL codes, not user-defined callback codes?

Probably use different styles for ElasticDL codes and user-defined callback codes to make it more clear.

workingloong: Yes, `SavedModelExporter` is ElasticDL code. And we will provide some pre-made user-defined callback where users only need to define the logic like `LearningRateScheduler`. But the methods of all callbacks are the same, such as `on_train_end` and so on. What's the mean of "different styles"?

skydoorkai: "different styles" means markdown formats.

docs/designs/keras_callback.md
```diff
@@ -0,0 +1,301 @@
+# Design to Support Custom Callback Using Keras API
+
+This document describes the design for supporting callback to customize the behavior of model during training, evaluation and inference in ElasticDL.
+
+## Motivation
+
+In deep learning, we generally need to customize the behavior of model during training, evaluation or
+inference, including reading/changing the model. We may perform the behavior per batch, per epoch or at the start and end of job. `tf.keras.callbacks.Callback` is an abstract base class and has methods to perform the behavior at different call frequency, such as `on_bath_end`, `on_epoch_end` and so on. So, we adopt the interfaces of `tf.keras.callbacks.Callback` to customize the behavior of model in ElasticDL. 
+
+Now we have implemented some modules similar to callback in ElasticDL, such as `LearingRateScheduler` and `PredictionOutputsProcessor`. And users should write definitions in the model definition file for each module like:
```
brightcoder01: LearingRateScheduler -> LearningRateScheduler

docs/designs/keras_callback.md
```diff
@@ -0,0 +1,301 @@
...
+    elif model_version < 12000:
+        return 0.0002
+    else:
+        return 0.0001
+
+# Process the prediction outputs for each batch
+class PredictionOutputsProcessor(BasePredictionOutputsProcessor):
+    ...
+```
+There will be multiple definition APIs for users to define different behaviors of model. The interfaces of APIs may be different. It is more convenient for users to define those behaviors using `tf.keras.callbacks.Callback`.
```
brightcoder01: multiple interface definition

brightcoder01: multiple definition APIs => different interface definitions

workingloong: Yeah !

brightcoder01: The interfaces of APIs may be different => The interfaces may be different.

API is `Application programming interface`

docs/designs/keras_callback.md
```diff
@@ -0,0 +1,301 @@
...
+
+Some use cases we observed that we want to support are:
+* Case 1: Callback similar to `PredictionOutputsProcessor` that is executed after prediction outputs are made.
+* Case 2: Callback to modulate learning rate as we are currently doing in `LearningRateScheduler`.
+* Case 3: Callback to write additional summaries to TensorBoard after each evaluation job completes.
+* Case 4: Callback to perform early stopping when certain conditions are met. For example, the metrics are met after an evaluation job.
+* Case 5: Callback to export model using SavedModel after the training is completed.
+* Case 6: Callback to upload model to remote storage after the training is completed.
+
+The following, we will design how to define and implement callbacks to support those cases.
```
brightcoder01: The following, we will design how to define and implement callbacks to support those cases.
=>
Next we will design how to define and implement callbacks to support those cases above.

docs/designs/keras_callback.md
```diff
@@ -0,0 +1,301 @@
...
+        """
+        metrics = logs.get("metrics", None)
+        if metrics is None:
+            return
+        self.metrics_list.append(metrics)
+        self.model.stop = stop_train(self.metrics_list)
+```
+The same as `SummaryWriter`, the master call `on_test_end` of `EarlyStopper` After an evaluation job is completed.
+
+* Case 6: Callback to upload model to remote storage after the training is completed.
```
brightcoder01: Master or worker will execute this callback?

docs/designs/keras_callback.md
```diff
@@ -0,0 +1,301 @@
...
+            logs: dict. Has keys batch and size representing the current batch number and the size of the batch.
+        """
+        if not hasattr(self.model.optimizer, 'lr'):
+            raise ValueError('Optimizer must have a "lr" attribute.')
+        lr = self.schedule_fn(batch)
+        K.set_value(self.model.optimizer.lr, lr)
+```
+Using ParameterServerStrategy, the worker calculates the batch gradients and set gradients to PS. PS updates weights using optimizer after receiving gradients. Although the worker can call `on_train_batch_begin` in `LearningRateScheduler` to adjust the learning rate of optimizer in its model, we should send the learning rate with gradients to PS by GRPC and PS updates weights using the received learning rate.
+
+* Case 3: Callback to write additional summaries to TensorBoard after each evaluation job completes.
```
brightcoder01: Master or worker will execute this callback?

docs/designs/keras_callback.md
```diff
@@ -0,0 +1,301 @@
...
+The master determine whether or not the evaluation job is completed by `EvaluationService.complete_task()`. So, we can call `on_test_end` After `EvaluationService.complete_task()` returns evaluation metrics.
+```python
+if evaluation_task_completed:
+    eval_metrics = self._evaluation_service.complete_task()
+    if eval_metrics is not None:
+        logs = {"metrics": eval_metrics}
+        self._callbacks_list.on_test_end(logs)
+```
+
+* Case 4: Callback to perform early stopping when the metrics are met after an evaluation job.
```
brightcoder01: Master or worker will execute this callback?

docs/designs/keras_callback.md
```diff
@@ -0,0 +1,301 @@
...
+Then, we can call `model` and `params` in the methods like:
+```python
+class CustomCallback(tf.keras.callbacks.Callback):
+    def on_train_batch_begin(self, batch, logs=None):
+        lr = self.model.optimizer.learning_rate,
+        saved_model_path = self.params.get("save_model_path")
+```
+
+
+## Implement Callbacks to Support the Cases in the Motivation
```
brightcoder01, line 117: For each method in the callback interface, please add the description about which role (master/worker/ps) will execute the method.

docs/designs/keras_callback.md
```diff
@@ -0,0 +1,301 @@
...
+    def on_test_end(self, logs=None):
+        """Call on the test job end
+        Args:
+            logs: dict. Has key metrics representing the evaluation metrics on the validation data.
+        """
+        metrics = logs.get("metrics", None)
+        if metrics is None:
+            return
+        self.metrics_list.append(metrics)
+        self.model.stop = stop_train(self.metrics_list)
```
brightcoder01: Does `model` contain a property named `stop`?

workingloong: The default properties of `tf.keras.Model` does not contain `stop`.
`tf.keras.callbacks.EarlyStopping` will set the  property`stop_training` 
https://github.com/tensorflow/tensorflow/blob/e5bf8de410005de06a7ff5393fafdf832ef1d4ad/tensorflow/python/keras/callbacks.py#L1268-L1270

The `stop_training` property of `model` is initialized before model iteration. https://github.com/tensorflow/tensorflow/blob/e5bf8de410005de06a7ff5393fafdf832ef1d4ad/tensorflow/python/keras/engine/training_arrays.py#L260

We also can set the property in ElasticDL and check whether or not to stop.

brightcoder01 (approved):

PR #1707, opened 2020-02-05
add vector definition

lwpyr:


elasticdl/pkg/common/util.go
```diff
@@ -1,23 +1,38 @@
...
+	valb := reflect.ValueOf(b)
+	for i := 0; i < vala.Len(); i++ {
+		if !CompareFloat(vala.Index(i).Float(), valb.Index(i).Float(), tolerance) {
+			return false
+		}
+	}
+	return true
+}
+
+// CompareIntArray compares two float32/64 array
```
QiJune: CompareIntArray compares two int array

elasticdl/pkg/common/vector.go
```diff
@@ -0,0 +1,235 @@
...
+	"reflect"
+	"unsafe"
+)
+
+// DataType definition
+type DataType struct {
+	Type reflect.Type
+	Size int
+	Flag Flag
+}
```
QiJune: What's the purpose to have a field `Type reflect.Type` here,  is an int flag is enough to represent DataType?

Or we could provide a function like `sizeof` and `typeof` to get the size in bytes and type of an int flag.

elasticdl/pkg/common/vector.go
```diff
@@ -0,0 +1,235 @@
...
+	TypeToDataType[Float32Dtype.Type] = Float32Dtype
+	TypeToDataType[Float64Dtype.Type] = Float64Dtype
+}
+
+// Vector definition
+type Vector struct {
+	Data   []byte
+	Length int
+	Dtype  DataType
+}
```
QiJune, line 16: Here, I think `Buffer` is more accurate than `Vector`.
We have a `Tensor` concept already, and `Vector` is usually taken as a 1-dim `Tensor`. 
`Buffer` is usually be regarded a struct contains a pointer to some contiguous memory, its size, and its type.

elasticdl/pkg/common/vector.go
```diff
@@ -0,0 +1,235 @@
...
+func CopyInit(slice interface{}) InitialzeFunc {
+	return func(b []byte) error {
+		buffer := bytes.NewBuffer(b)
+		buffer.Reset()
+		return binary.Write(buffer, binary.LittleEndian, slice)
+	}
+}
+
+// NewEmptyVector create an zero-initialized vector
+func NewEmptyVector(length int, dtype DataType) *Vector {
```
QiJune: Empty is not needed, I think `NewBuffer` is clear enough.

elasticdl/pkg/common/vector.go
```diff
@@ -0,0 +1,235 @@
...
+	var vec = Vector{
+		Data:   make([]byte, bytelen, bytelen),
+		Dtype:  dtype,
+		Length: length,
+	}
+	return &vec
+}
+
+// NewInitializerVector create an initialized vector
+func NewInitializerVector(length int, dtype DataType, initializer InitialzeFunc) *Vector {
```
QiJune: NewInitializerVector is also a little confusing. Pass a callback initializer function here is also introduce complexity. Maybe we could follow numpy style:

```
NewZeroBuffer --> np.zeros
NewConstantBuffer --> np.full
```

elasticdl/pkg/common/vector.go
```diff
@@ -0,0 +1,235 @@
...
+		Data: uintptr(unsafe.Pointer(&v.Data[0])),
+		Cap:  int(v.Length),
+		Len:  int(v.Length),
+	}
+	val := reflect.NewAt(v.Dtype.Type, unsafe.Pointer(&sliceHeader)).Elem()
+	return val.Interface()
+}
+
+// MakeSlice gives a slice copy of the Vector data
+func (v *Vector) MakeSlice() interface{} {
```
QiJune, line 163: Maybe we could merge `InplaceSlice` and `MakeSlice` into one function `ToSlice`

```go
func (b *Buffer) ToSlice(isCopy bool) interface {} {
   ......
}
```

elasticdl/pkg/common/vector.go
```diff
@@ -0,0 +1,235 @@
...
+	var vec = Vector{
+		Data:   data,
+		Dtype:  dtype,
+		Length: length,
+	}
+	return &vec
+}
+
+// NewVectorFrom create a vector using a slice to initialize
+func NewVectorFrom(slice interface{}) *Vector {
```
QiJune: NewVectorFrom --> FromSlice.

We could also choose copy from slice or just take a view from the slice.

elasticdl/pkg/common/util.go
lwpyr: fixed

elasticdl/pkg/common/vector.go
lwpyr: it's becasue "InplaceSlice" and "MakeSlice" need this "reflect.Type" field

elasticdl/pkg/common/vector.go
lwpyr: Because This data structure is planned to be used in EmbeddinTable field `EmbeddingVector map[int64]*Vector`

elasticdl/pkg/common/vector.go
lwpyr: For embeddingTable, when getting a vector which index is not stored in the table, the table will return a emptyvector

elasticdl/pkg/common/vector.go
lwpyr: the initializer is for future concern, this could support custom init function like Gauss or something else.
In the embedding table definition, there is a field named initializer

elasticdl/pkg/common/vector.go
lwpyr: yeah, it's added. inplace or copy

elasticdl/pkg/common/vector.go
lwpyr: when writing codes, Maybe `vec.ToSlice(true)` is not so obvious which of `inplace` or `copy` is true?

elasticdl/pkg/common/embedding_table.go
```diff
@@ -7,51 +7,50 @@ type EmbeddingTable struct {
...
-	table.EmbeddingVector = make(map[int64]*Tensor)
-	return &table
+func NewEmbeddingTable(name string, dim int64, initializer string, dtype DataType) *EmbeddingTable {
+	var e EmbeddingTable
+	e.Name = name
+	e.Dim = dim
+	e.Initializer = initializer
+	e.EmbeddingVector = make(map[int64]*Vector)
+	e.Dtype = dtype
+	return &e
```
wangkuiyi, line 23:
```go
return &EmbeddingTable {
   Name: name,
   Dim: dim,
   Initialzier: initializer, 
...
}
```

elasticdl/pkg/common/tensor.go
```diff
@@ -4,97 +4,170 @@ import (
...
 	Indices []int64
 }
 
-// NewTensor create a new n-dim tensor
-func NewTensor(dim []int64) *Tensor {
-	var t Tensor
-	t.Value = make([]float32, GetDimProduct(dim))
-	t.Dim = dim
+// ProductInt64 get the number of the elements of a tensor of this dim
+func ProductInt64(dim []int64) int64 {
```
wangkuiyi: ProductInt64 => DimSize?

elasticdl/pkg/common/embedding_table.go
lwpyr: fixed

elasticdl/pkg/common/tensor.go
lwpyr: fixed

elasticdl/pkg/common/vector.go
lwpyr: fixed

PR #1708, opened 2020-02-05
support tf.IndexedSlices gradient for dense param

QiJune:
Current optimizer of Go PS could not handle native keras.embedding layer. It will generate a `tf.IndexedSlice` gradient, but the parameter is in NonEmbeddingParam.

This PR fix it.

### DIFF
--- a/elasticdl/docker/Dockerfile.dev
+++ b/elasticdl/docker/Dockerfile.dev
@@ -17,6 +17,26 @@ ARG GO_MIRROR_URL=https://dl.google.com/go
 RUN pip install -r /requirements.txt --extra-index-url=${EXTRA_PYPI_INDEX}
 RUN pip install -r /requirements-dev.txt --extra-index-url=${EXTRA_PYPI_INDEX}
 
+# Copy the data generation package to /var and run them from there.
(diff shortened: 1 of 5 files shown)


elasticdl/docker/Dockerfile.dev
```diff
@@ -37,23 +57,3 @@ RUN cd /tmp/elasticdl && make -f elasticdl/Makefile && \
     export GO111MODULE=on && \
     go mod init elasticdl.org/elasticdl && \
     go install ./... && cd / && rm -rf /tmp/elasticdl
-
-# Copy the data generation package to /var and run them from there.
```
QiJune, line 41: In my local test environment, I find that I have to build the elasticdl:dev if I modify Go code.
Just move commands not related with Go up, since it will be cached. This will reduce rebuild time.

skydoorkai: Better to put this change is a separate PR.
Looks fine to me otherwise.

skydoorkai (approved):

QiJune:
The unit test is under developing. Currently, please do not merge this PR.

PR #1715, opened 2020-02-06
add core concepts of parameter server part in ElasticDL

QiJune:
This PR aims to clarify the core concepts of parameter server part in ElasticDL. The doc will be updated continually. Any comments are welcomed.

### DIFF
--- a/docs/designs/concepts.md
+++ b/docs/designs/concepts.md
@@ -4,17 +4,15 @@ This document describes core concepts used in ElasticDL, and discusses their rep
 
 ## Core Concepts
 
-Fundamentally, machine learning is using algorithms to extract information from raw data and represent it in some type of model.
-
-There are two kinds of parameters in a model, one is dense parameter, the other is sparse parameter. The sparse parameter is used by the embedding layer, so it is also called embedding table. The embedding table maps discrete ID *i* to embedding vector *vᵢ*.
+There are usually two kinds of parameters in a deep learning model, one is dense parameter, the other is sparse parameter. The sparse parameter is used by the embedding layer, which is also called embedding table. The embedding table maps discrete ID *i* to embedding vector *vᵢ*.
 
 Both dense parameter and embedding table includes a tensor data field, and some other auxiliary fields, like name. Embedding table has an extra auxiliary field called indices.
 
-Tensor is used to represented a n-dim data. We support several data types, such as int8/int32/float32/float64.
+Tensor is used to represented a n-dim data. We support several data types, such as int8/int32/int64/float32/float64.
 
-Besides, we have two kinds of gradients, dense gradient and embedding table gradient. The components of dense gradient is the same with dense parameter, and the components of embedding table gradient is also the same with embedding table.
+Besides, we have two kinds of gradients, dense gradient and embedding table gradient. The data structure of dense gradient is the same with dense parameter, and so is the embedding table gradient.
 
-In ElasticDL, only a large embedding layer is represented by an embedding table. For small embedding layer, a dense parameter is used, but its gradient is sparse. It has indices, which is actually an embedding table gradient.
+In ElasticDL, only the parameter of large embedding layer is represented by an embedding table. For small embedding layer, a dense parameter is used, and its gradient also has an indices field. In this case, the data structure of gradient is not dense param, but an embedding table.
 
 Let's make a short summary, following is all the core concepts used in ElasticDL:
 
@@ -25,7 +23,7 @@ Let's make a short summary, following is all the core concepts used in ElasticDL
 
 ## In-Memory Representation
 
-EmbeddingTable is initialized lazily in PS. Worker sends item ids to request embedding vectors from PS. If the ID is not found in PS, the corresponding embedding vector will be initialized and send back to the worker. For better query and increasing embedding vectors, we use a map data structure to represent EmbeddingTable.
+EmbeddingTable is initialized lazily in PS. Worker sends item ids to request embedding vectors from PS. If the ID is not found in PS, the corresponding embedding vector will be initialized and sent back to the worker. In order to query and insert embedding vectors efficiently, we use a map data structure to represent EmbeddingTable.
 
 ```go
 type TensorDtype = int64
@@ -62,9 +60,11 @@ type DenseParam struct {
 type Emdedding = Buffer // alias for better understanding
 
 type EmbeddingTable struct {
-    Name       string
-    Embeddings map[int64]*Emdedding
-    Dtype      TensorDtype
+    Name        string
+    Dim         int64
+    Dtype       TensorDtype
+    Initializer string
+    Embeddings  map[int64]*Emdedding
 }
 
 type Model struct {
@@ -113,10 +113,9 @@ message EmbeddingTable {
 }
 
 message Model {
-  bool initialized = 1;
-  int32 version = 2;
-  repeated DenseParam dense_params = 3;
-  repeated EmbeddingTable embedding_tables = 4;
+  int32 version = 1; // model updated times
+  repeated DenseParam dense_params = 2;
+  repeated EmbeddingTable embedding_tables = 3;
 }
 ```
 
--- a/docs/designs/concepts.md
+++ b/docs/designs/concepts.md
@@ -25,6 +25,54 @@ Let's make a short summary, following is all the core concepts used in ElasticDL
 
 ## In-Memory Representation
 
+```go
+type TensorDtype = int64
+
+const (
+    Invalid TensorDtype = iota
+    Int8 
+    Int16
+    Int32
+    Int64
+    Float16
+    Float32
+    Float64
+    Bool
+)
+
+type Buffer struct {
+    Data   []byte
+    Length int64
+    Dtype  TensorDtype
+}
+
+type Tensor struct {
+    Content *Buffer
+    Dims    []int64
+    Indices []int64
+}
+
+type DenseParam struct {
+    Name   string
+    Tensor Tensor
+}
+
+type Emdedding = Buffer // alias for better understanding
+
+type EmbeddingTable struct {
+    Name       string
+    Embeddings map[string]*Emdedding
+    Dtype      TensorDtype
+}
+
+type Model struct {
+    Initialized     bool
+    Version         int32
+    DenseParams     map[string]*DenseParam
+    EmbeddingTables map[string]*EmbeddingTable
+}
+```
+
 
 ## Message Representation
 
--- a/docs/designs/concepts.md
+++ b/docs/designs/concepts.md
@@ -6,13 +6,15 @@ This document describes core concepts used in ElasticDL, and discusses their rep
 
 Fundamentally, machine learning is using algorithms to extract information from raw data and represent it in some type of model.
 
-There two kinds of parameters in a model, one is dense parameter, the other is sparse parameter. The sparse parameter is used by the embedding layer, so it is also called embedding table.
+There are two kinds of parameters in a model, one is dense parameter, the other is sparse parameter. The sparse parameter is used by the embedding layer, so it is also called embedding table. The embedding table maps discrete ID *i* to embedding vector *vᵢ*.
 
-Both dense parameter and embedding table includes a tensor data field, and some other auxiliary fields. The auxiliary fields indluce a name. Embedding table has a extra auxiliary field, indices field.
+Both dense parameter and embedding table includes a tensor data field, and some other auxiliary fields, like name. Embedding table has an extra auxiliary field called indices.
 
-Tensor is used to represented a n-dim data. There are many data types, such as int8/float32.
+Tensor is used to represented a n-dim data. We support several data types, such as int8/int32/float32/float64.
 
-Each kinds of parameter has its own kind of gradient. Dense parameter has dense parameter gradient, embedding table has embedding table gradient.
+Besides, we have two kinds of gradients, dense gradient and embedding table gradient. The components of dense gradient is the same with dense parameter, and the components of embedding table gradient is also the same with embedding table.
+
+In ElasticDL, only a large embedding layer is represented by an embedding table. For small embedding layer, a dense parameter is used, but its gradient is sparse. It has indices, which is actually an embedding table gradient.
 
 Let's make a short summary, following is all the core concepts used in ElasticDL:
 
@@ -20,10 +22,10 @@ Let's make a short summary, following is all the core concepts used in ElasticDL
 - DenseParam
 - EmbeddingTable
 - Tensor
-- DenseParamGradient
-- EmbeddingTableGradient
 
 ## In-Memory Representation
+
+EmbeddingTable is initialized lazily in PS. Worker sends item ids to request embedding vectors from PS. If the ID is not found in PS, the corresponding embedding vector will be initialized and send back to the worker. For better query and increasing embedding vectors, we use a map data structure to represent EmbeddingTable.
 
 ```go
 type TensorDtype = int64
@@ -117,9 +119,6 @@ message Model {
   repeated EmbeddingTable embedding_tables = 4;
 }
 ```
-
-Please note that the components of dense parameter gradient is the same with dense parameter, and the components of embedding table gradient is also the same with embedding table.
-
 
 ## RPC Service
 
--- a/docs/designs/concepts.md
+++ b/docs/designs/concepts.md
@@ -15,13 +15,6 @@ A dense parameter is a dense tensor with a name.  An embedding table is a map fr
 
 where the curly braces denote zero or more.
 
-It is notable that in practice, we use the rule
-
-- embedding table = tensor + index + name
-- index = map[ID]address
-
-because we append all embedding vectors into the tensor and use index, which is a map from ID to the starting address of the embedding vector in the tensor.
-
 To update the model, workers compute and report gradients.  Accordingly, we have two kinds of gradients:
 
 1. dense gradient, and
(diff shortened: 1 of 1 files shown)


docs/designs/concepts.md
```diff
@@ -0,0 +1,116 @@
+# Concepts Design Doc
+
+This document describes core concepts used in ElasticDL, and discusses their representations.
+
+## Core Concepts
+
+Fundamentally, machine learning is using algorithms to extract information from raw data and represent it in some type of model.
+
+There two kinds of parameters in a model, one is dense parameter, the other is sparse parameter. The sparse parameter is used by the embedding layer, so it is also called embedding table.
```
skydoorkai: There two kinds  -> There are two kinds ...

docs/designs/concepts.md
```diff
@@ -0,0 +1,116 @@
...
+
+Both dense parameter and embedding table includes a tensor data field, and some other auxiliary fields. The auxiliary fields indluce a name. Embedding table has a extra auxiliary field, indices field.
```
skydoorkai: indluce -> include
a extra -> an extra

docs/designs/concepts.md
```diff
@@ -0,0 +1,116 @@
...
+
+message Model {
+  bool initialized = 1;
+  int32 version = 2;
+  repeated DenseParam dense_params = 3;
+  repeated EmbeddingTable embedding_tables = 4;
+}
+```
+
+Please note that the components of dense parameter gradient is the same with dense parameter, and the components of embedding table gradient is also the same with embedding table.
```
skydoorkai: In ElasticDL, only a large embedding layer is represented by an embedding table.  For small embedding layer, a dense parameter is used, but its gradient will have indices.

docs/designs/concepts.md
```diff
@@ -0,0 +1,116 @@
...
+
+message EmbeddingTable {
+  string name = 1;
+  Tensor tensor = 2;
+  repeated int64 indices = 3;
+}
+
+message Model {
+  bool initialized = 1;
+  int32 version = 2;
```
QiJune: Once step updating increase the version one time. The training of a Bert-large model need 1M steps udpating. So int32 is enough here.

docs/designs/concepts.md
```diff
@@ -0,0 +1,165 @@
...
+type DenseParam struct {
+    Name   string
+    Tensor Tensor
+}
+
+type Emdedding = Buffer // alias for better understanding
+
+type EmbeddingTable struct {
+    Name       string
+    Embeddings map[string]*Emdedding
```
QiJune: The key of Embeddings is int64. The embeddings maps from an item id to an embedding vector.

docs/designs/concepts.md
QiJune: Done

docs/designs/concepts.md
QiJune: Done

docs/designs/concepts.md
QiJune: Done

docs/designs/concepts.md
```diff
@@ -0,0 +1,164 @@
...
+
+Fundamentally, machine learning is using algorithms to extract information from raw data and represent it in some type of model.
+
+There are two kinds of parameters in a model, one is dense parameter, the other is sparse parameter. The sparse parameter is used by the embedding layer, so it is also called embedding table. The embedding table maps discrete ID *i* to embedding vector *vᵢ*.
+
+Both dense parameter and embedding table includes a tensor data field, and some other auxiliary fields, like name. Embedding table has an extra auxiliary field called indices.
+
+Tensor is used to represented a n-dim data. We support several data types, such as int8/int32/float32/float64.
+
+Besides, we have two kinds of gradients, dense gradient and embedding table gradient. The components of dense gradient is the same with dense parameter, and the components of embedding table gradient is also the same with embedding table.
```
brightcoder01: The components of dense gradient -> What's the meaning of component?

docs/designs/concepts.md
```diff
@@ -0,0 +1,164 @@
...
+Let's make a short summary, following is all the core concepts used in ElasticDL:
+
+- Model
+- DenseParam
+- EmbeddingTable
+- Tensor
+
+## In-Memory Representation
+
+EmbeddingTable is initialized lazily in PS. Worker sends item ids to request embedding vectors from PS. If the ID is not found in PS, the corresponding embedding vector will be initialized and send back to the worker. For better query and increasing embedding vectors, we use a map data structure to represent EmbeddingTable.
```
brightcoder01: For better query and increasing embedding vectors => In order to query and insert embedding vectors efficiently

docs/designs/concepts.md
```diff
@@ -0,0 +1,164 @@
...
+}
+
+message EmbeddingTable {
+  string name = 1;
+  Tensor tensor = 2;
+  repeated int64 indices = 3;
+}
+
+message Model {
+  bool initialized = 1;
```
brightcoder01: Add comments for each property here?

docs/designs/concepts.md
brightcoder01: send back -> sent back

docs/designs/concepts.md
```diff
@@ -0,0 +1,164 @@
+# Concepts Design Doc
+
+This document describes core concepts used in ElasticDL, and discusses their representations.
+
+## Core Concepts
+
+Fundamentally, machine learning is using algorithms to extract information from raw data and represent it in some type of model.
```
brightcoder01: This sentence is a little confusing

docs/designs/concepts.md
```diff
@@ -0,0 +1,164 @@
...
+
+There are two kinds of parameters in a model, one is dense parameter, the other is sparse parameter. The sparse parameter is used by the embedding layer, so it is also called embedding table. The embedding table maps discrete ID *i* to embedding vector *vᵢ*.
+
+Both dense parameter and embedding table includes a tensor data field, and some other auxiliary fields, like name. Embedding table has an extra auxiliary field called indices.
+
+Tensor is used to represented a n-dim data. We support several data types, such as int8/int32/float32/float64.
```
brightcoder01: int64 is missing?

docs/designs/concepts.md
```diff
@@ -0,0 +1,164 @@
...
+
+Besides, we have two kinds of gradients, dense gradient and embedding table gradient. The components of dense gradient is the same with dense parameter, and the components of embedding table gradient is also the same with embedding table.
+
+In ElasticDL, only a large embedding layer is represented by an embedding table. For small embedding layer, a dense parameter is used, but its gradient is sparse. It has indices, which is actually an embedding table gradient.
```
brightcoder01: It has indices, which is actually an embedding table gradient. => It's the gradients for the embedding vectors corresponding to the indices?

docs/designs/concepts.md
lwpyr: Could fields or properties be more suitable?

QiJune: It means data structure. The data structure of dense gradient is the same with dense param.

docs/designs/concepts.md
lwpyr: This is a little bit confusing, is `For small embedding layer, a dense parameter is used` locally?

docs/designs/concepts.md
lwpyr: Referenced from Wikipedia
`Machine learning algorithms build a mathematical model based on sample data, known as "training data", in order to make predictions or decisions without being explicitly programmed to perform the task.`

docs/designs/concepts.md
QiJune: Done

docs/designs/concepts.md
QiJune: Remove it :)

docs/designs/concepts.md
QiJune: Done

docs/designs/concepts.md
QiJune: Rewrite the sentence.

docs/designs/concepts.md
```diff
@@ -0,0 +1,165 @@
...
+
+There are usually two kinds of parameters in a model, one is dense parameter, the other is sparse parameter. The sparse parameter is used by the embedding layer, which is also called embedding table. The embedding table maps discrete ID *i* to embedding vector *vᵢ*.
+
+Both dense parameter and embedding table includes a tensor data field, and some other auxiliary fields, like name. Embedding table has an extra auxiliary field called indices.
+
+Tensor is used to represented a n-dim data. We support several data types, such as int8/int32/int64/float32/float64.
+
+Besides, we have two kinds of gradients, dense gradient and embedding table gradient. The data structure of dense gradient is the same with dense parameter, and so is the embedding table gradient.
+
+In ElasticDL, only the parameter of large embedding layer is represented by an embedding table. For small embedding layer, a dense parameter is used, and its gradient also has an indices field. In this case, the data structure of gradient is not dense param, but an embedding table.
```
skydoorkai: so `The data structure of dense gradient is the same with dense parameter` is not true.
Need to add the exception there.

`and so is the embedding table gradient`. Make it clear, not also the same with dense parameter, but the same with embedding table.

docs/designs/concepts.md
```diff
@@ -0,0 +1,180 @@
...
+Please be aware that if some embedding table maps zero-based successive ID values to embedding vectors.  Such kind of embedding table, if not too large, could be represented by a dense parameters.  In the rest of the design, the concept embedding table refers to a sparse/general embedding table, where the ID might not be zero-based successive values, and even not an numerical value.
+
+Let's make a short summary, following is all the core concepts of ElasticDL include:
+
+- model = {dense parameter} + {embedding table}
+- dense parameter = tensor + name
+- embedding table = tensor + index + name
+- index = map[ID]address
+
+## In-Memory Representation
```
wangkuiyi: In-memory Representaiton => Tensor

docs/designs/concepts.md
```diff
@@ -0,0 +1,180 @@
...
+
+EmbeddingTable is initialized lazily in PS. Worker sends item ids to request embedding vectors from PS. If the ID is not found in PS, the corresponding embedding vector will be initialized and sent back to the worker. In order to query and insert embedding vectors efficiently, we use a map data structure to represent EmbeddingTable.
+
+```go
+type TensorDtype = int64
```
wangkuiyi: TensorDType => TensorElementType

docs/designs/concepts.md
```diff
@@ -0,0 +1,180 @@
...
+    Data   []byte
+    Length int64
+    Dtype  TensorDtype
+}
+
+type Tensor struct {
+    Content *Buffer
+    Dims    []int64
+    Indices []int64
+}
```
wangkuiyi: Do we need stride?

docs/designs/concepts.md
```diff
@@ -0,0 +1,180 @@
...
+type Buffer struct {
+    Data   []byte
+    Length int64
+    Dtype  TensorDtype
+}
+
+type Tensor struct {
+    Content *Buffer
+    Dims    []int64
+    Indices []int64
```
wangkuiyi: What are Indices?

docs/designs/concepts.md
```diff
@@ -0,0 +1,180 @@
...
+}
+
+type DenseParam struct {
```
wangkuiyi: DenseParam => DenseParameter

docs/designs/concepts.md
```diff
@@ -0,0 +1,180 @@
...
+    Name   string
+    Tensor Tensor
+}
+
+type Emdedding = Buffer // alias for better understanding
```
wangkuiyi: Don't use the alias, which usually implies that we are not clear with the naming.

docs/designs/concepts.md
```diff
@@ -0,0 +1,180 @@
...
+
+type EmbeddingTable struct {
+    Name        string
+    Dim         int64
+    Dtype       TensorDtype
+    Initializer string
+    Embeddings  map[int64]*Emdedding
```
wangkuiyi: Embedding => Tensor

docs/designs/concepts.md
```diff
@@ -0,0 +1,180 @@
...
+    Int16
+    Int32
+    Int64
+    Float16
+    Float32
+    Float64
+    Bool
+)
+
+type Buffer struct {
```
wangkuiyi: I don't see that we need type Buffer, do we?  If not necessary, let us follow Occam's Razor.

docs/designs/concepts.md
```diff
@@ -0,0 +1,180 @@
...
+- model = {dense parameter} + {embedding table}
+- dense parameter = tensor + name
+- embedding table = tensor + index + name
+- index = map[ID]address
+
+## In-Memory Representation
+
+EmbeddingTable is initialized lazily in PS. Worker sends item ids to request embedding vectors from PS. If the ID is not found in PS, the corresponding embedding vector will be initialized and sent back to the worker. In order to query and insert embedding vectors efficiently, we use a map data structure to represent EmbeddingTable.
+
+```go
```
wangkuiyi: I would suggest that we use protobuf instead of Go code in the design doc because protobuf is designed to be an IDL.

docs/designs/concepts.md
```diff
@@ -0,0 +1,180 @@
...
+    Dtype       TensorDtype
+    Initializer string
+    Embeddings  map[int64]*Emdedding
+}
+
+type Model struct {
+    Initialized     bool
+    Version         int32
+    Dtype           TensorDtype
+    DenseParams     map[string]*DenseParam
```
wangkuiyi: I agree to use map here in Model.  It seems that we don't need type DenseParameter={Tensor+name} any more.  Instead, we can have

```protobuf
syntax = "proto3";

message Model {
    map<string, Tensor> DenseParameters = 1,
}
```

wangkuiyi (changes requested):

docs/designs/concepts.md
QiJune: @wangkuiyi Here, I want to clarify the reason why we still need some extra in-memory data structure in Go, especially for embedding table. There are two reasons:

1.  Embedding table is usually a large dense tensor. In forward pass of embedding layer, worker will not pull the full table, but sends an array of item IDs to PS. Then the PS looks up the embeddig table, and sends corresponding embedding vectors back to worker. The worker will get a subset of the whole table. In the backward pass of embeddig layer, the gradients are also associated with IDs. Only IDs which are used in forward pass make contribution to the gradients.

TensorFlow introduce a data structure called [IndexedSlices](https://www.tensorflow.org/api_docs/python/tf/IndexedSlices) to represent the subset of a big dense tensor.
 The IndexedSlices has two filed, Tensor and Indices. Indices is a 1-dim array.

```
dense[indices[i],:] = tensor[i,:] 
```

2.  In our design, the embedding table is initialized lazily in PS. And the size of embedding table is also unknown. So we could not create a big dense tensor simplely. When looking IDs from workers, if the ID is not found in the table, the corresponding embedding vector will be created and initialized.  In order to query and insert embedding vectors efficiently in PS, we use a map data structure to represent EmbeddingTable.

In PS side, the embedding table parameter is a map data structure. In worker side, we pull the subset of the embedding table from PS, and push the gradients associated with IDs to PS.

We define a `IndexedSlices` message for the embedding table related RPC between worker and PS.

Following is the design:

```proto
message Tensor {
  repeated int64 dim = 1;
  bytes content = 2;
  TensorElementType dtype = 3;
}

message IndexedSlices {
  Tensor tensor = 1;
  repeated int64 indices = 2;
}

message Model {
  int32 version = 1; // model updated times
  map<string, Tensor> dense_parameters = 2;
  map<string, IndexedSlices> indexed_slices = 3;
}
```

For PS side, we propose following in-memory data structure in Go:

```go
import "elasticdl.org/elasticdl/pkg/proto"

type EmbeddingTable struct {
  Name             string
  Dim              int64
  Initializer      string
  EmbeddingVectors map[int64]*proto.Tensor
}

type Model struct {
  Version           int32
  InitStatus        bool
  DenseParameters   map[string]*proto.Tensor
  EmbeddingTables   map[string]*EmbeddingTable
}
```

docs/designs/concepts.md
```diff
@@ -0,0 +1,180 @@
...
+- model = {dense parameter} + {embedding table}
+- dense parameter = tensor + name
+- embedding table = {id, tensor} + name
+
+where the curly braces denote zero or more.
+
+It is notable that in practice, we use the rule
+
+- embedding table = tensor + index + name
+- index = map[ID]address
```
skydoorkai: embedding table = tensor + index + name
So this `index` is not ID but address? 
What is the relationship with IndexedSlices? Is `indices` in `IndexedSlices` is this `index`?
```
message IndexedSlices {
  Tensor tensor = 1;
  repeated int64 indices = 2;
}
```

docs/designs/concepts.md
QiJune: Please refer to the comment [here](https://github.com/sql-machine-learning/elasticdl/pull/1715/files#r376240095)

docs/designs/concepts.md
QiJune: Please refer to the comment [here](https://github.com/sql-machine-learning/elasticdl/pull/1715/files#r376240095).

After we reach a consensus, I will update the doc.

docs/designs/concepts.md
wangkuiyi: After the phone call, I got the point.

On the parameter server, we'd prefer to maintain each embedding table as a map from ID to embedding vectors.  With such a data structure, it is efficient to allocate memory for new embedding vectors.  On the contrary, we'd concatenate embedding vectors into protobuf messages for parameter pulling and gradient pushing.  We cannot use concatenated embedding vectors as the in-memory data structure on the PS, because allocating new embedding vectors involves resize the space of concatenated embedding vectors.

wangkuiyi:
```protobuf
message IndexedSlices {
  Tensor concat_embedding_vecs = 1;
  repeated uint64 ids = 2;
}
```

wangkuiyi (approved):

wangkuiyi:
I am approving this PR to dismiss my previous request for change. Please go on following the comments before merging. Thanks!

docs/designs/concepts.md
QiJune: Done

docs/designs/concepts.md
QiJune: Done, remove it

docs/designs/concepts.md
QiJune: Done

docs/designs/concepts.md
QiJune: Done, remove it

docs/designs/concepts.md
QiJune: Remove it, since we use `map<string, Tensor>` as DenseParameter

docs/designs/concepts.md
QiJune: Done

docs/designs/concepts.md
QiJune: Done

docs/designs/concepts.md
QiJune: Currently, we do not need stride.

The tensor is used for parameter pulling and gradient pushing. The paramter and gradient are all contious data.

terrytangyuan (approved):

PR #1719, opened 2020-02-09
Add concat_column in ElasticDL feature column

brightcoder01:
ConcatColumn is an option to implement the `CONCAT` transform function in the data analysis and transform #1664


elasticdl/python/elasticdl/feature_column/feature_column.py
```diff
@@ -185,3 +187,135 @@ def reset(self):
...
+            )
+            feature_tensors.append(ids_and_weights.id_tensor)
+
+        feature_tensors_with_offset = []
+        for index, offset in enumerate(self.accumulated_offsets):
+            feature_tensor = feature_tensors[index]
+            feature_tensor_with_offset = tf.SparseTensor(
+                indices=feature_tensor.indices,
+                values=tf.cast(
+                    tf.add(feature_tensor.values, offset), tf.int32
```
workingloong: Why cast to tf.int32 not tf.int64?

elasticdl/python/elasticdl/feature_column/feature_column.py
```diff
@@ -185,3 +187,135 @@ def reset(self):
...
+
+        return True
+
+    @property
+    def name(self):
+        feature_names = []
+        for categorical_column in self.categorical_columns:
+            feature_names.append(categorical_column.name)
+
+        return "_C_".join(sorted(feature_names))
```
workingloong, line 240: Maybe we can set an argument `key` in the `__init___` like others column class in tf.feature_columns.
https://github.com/tensorflow/tensorflow/blob/cf7fcf164c9846502b21cebb7d3d5ccf6cb626e8/tensorflow/python/feature_column/feature_column_v2.py#L2761-L2776

elasticdl/python/elasticdl/feature_column/feature_column.py
brightcoder01: Good suggestion, tf.int64 is better

elasticdl/python/elasticdl/feature_column/feature_column.py
brightcoder01: The 'key' means that the input can be the raw input. For this concat_column, we only accept the input of `CategoricalColumn` type

workingloong (approved):

PR #1699, opened 2020-02-04
Add a design to normalize the table schema

workingloong:
Fix #1648

PR #1720, opened 2020-02-10
Add ODPS UDF to normalize table with key-value column to wide table

workingloong:
Implement according to the design in PR #1699

### DIFF
--- a/tools/odps_table_tools/normalize_kv_udf.py
+++ b/tools/odps_table_tools/normalize_kv_udf.py
@@ -1,0 +1,54 @@
+# coding:utf-8
+# qinlong.wql
+from odps.udf import BaseUDTF
+
+
+def parse_kv_string_to_dict(
+    kvs_string, output_key_names, inter_kv_sep, intra_kv_sep
+):
+    """ Parse kv string to a dict like "k1:v1,k2:v2" to {k1:v1,k2:v2}
+    Args:
+        kvs_string: Key-value pairs string
+        output_key_names: All key names to be saved in the output dict
+        inter_kv_sep: Inter separator in the Key-value pair string
+        intra_kv_sep: Intra separator int Ker-value pairs string.
+
+    Returns:
+        dict
+    """
+    kv_dict = {}
+    kv_pairs = kvs_string.split(inter_kv_sep)
+    for kv in kv_pairs:
+        key_and_value = kv.split(intra_kv_sep)
+        if len(key_and_value) == 2:
+            kv_dict[key_and_value[0]] = key_and_value[1]
+    values = []
+    for name in output_key_names:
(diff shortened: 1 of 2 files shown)


wangkuiyi:
LGTM

tools/odps_table_tools/transform_kv_table.py
```diff
@@ -0,0 +1,318 @@
+import argparse
+import time
+
+import odps
+from odps import options, ODPS
+
+options.log_view_host = "http://logview.odps.aliyun-inc.com:8080"
```
terrytangyuan, line 7: This information can be exposed to public?

tools/odps_table_tools/transform_kv_table.py
```diff
@@ -0,0 +1,318 @@
...
+    try:
+        function = odps_entry.get_function(udf_function)
+        function.drop()
+    except odps.errors.NoSuchObject:
+        pass
+    finally:
+        print("Drop function is exists {}".format(udf_function))
+
+
+def flat_to_wide_table(
```
brightcoder01, line 155: Do we need make this code into a package? How do user use this code?

tools/odps_table_tools/transform_kv_table.py
workingloong: We can find the information in the document of aliyun. https://help.aliyun.com/document_detail/42745.html?spm=5176.10695662.1996646101.searchclickresult.4d975c33YqRrs8
So, I think it can be exposed to the public.

tools/odps_table_tools/normalize_kv_udf.py
```diff
@@ -0,0 +1,54 @@
+# coding:utf-8
+# qinlong.wql
```
terrytangyuan: Remove these comments

workingloong: Oh, Yes!

tools/odps_table_tools/transform_kv_table.py
workingloong: The user must use those codes on MAXCOMPUTE where the can submit PyODPS jobs. There maybe two ways:
1. The user can create a[ PyODPS node ](https://help.aliyun.com/knowledge_detail/121718.html?spm=a2c4g.11186631.2.1.32192d1ddR5hFz)in DataWorks. However, the user cannot install a python package on the PyODPS node, so they need to copy the UDF codes to PyODPS nodes. 
2. The user can execute by "python transform_kv_table.py --args" on the machine which can keep the connection with MAXCOMPUTE.

So, we cannot make these codes into a package because the user needs to upload the UDF codes to MAXCOMPUTE to create a UDF.

brightcoder01 (approved):

Issue #1722, opened 2020-02-11
Clean PS code according to the concepts document


QiJune:
We rethink the core concepts of parameter server, and put them to a [doc](https://github.com/sql-machine-learning/elasticdl/blob/develop/docs/designs/concepts.md). And we plan to refactor our codes according to the doc.

Most of the basic proto messages are changed. This will break a lots of unit tests now. We have to modified both Go PS code/Python PS code/Python worker code. Adding a single big PR to fix all the codes is a choice. But it brings a lots of risk, and it's hard to review.

We must change the code base step by step. In each step, we set up small PRs to the code base.

I propose a two-step solution.

Step 1

We add some new messages like IndexedSlices, and add some new fields to current messages. For example, the new Model proto mesages could be:

```proto
message Tensor {
  string name = 1;
  repeated int64 dim = 2;
  bytes content = 3;
  repeated int64 indices = 4;
  TensorDtype dtype = 5;
}

message IndexedSlices {
  Tensor concated_vectors = 1;
  repeated int64 ids = 2;
}

message Model {
  int32 version = 1;  // model updated times
  repeated Tensor param = 2;
  repeated EmbeddingTableInfo embedding_table_info = 3;
  map<string, Tensor> dense_parameters = 4;
  map<string, IndexedSlices> indexed_slices = 5;
}
```

In this way, the Python code is not broken. All current Python related unit tests are good. And we could modify our Go PS code according to the new design.

In this step, some proto messages are redundant. But we keeps the consistency of Python code. We could sends a lot of small PRs to refactor Go code.

Step 2.

We remove redundant fileds of some proto messages, and remove some redundant proto messages.

In this step, the Go PS code is not affected. We only clean the Python code base.

mhaoli:
Maybe we can split Step 2 into 2 steps, step 2 and step 3.

Step 2
Use `numpy.ndarray` and `tf.IndexedSlices` to replace `elasticdl.Tensor` data structure for gRPC services in python code. We can replace them concurrently for different gRPC services.

Step 3
Clean redundant proto messages.

QiJune:
Since Python PS will be replaced with Go PS. We should not put too much time to refactor Python code.

Step 1. 

- Add new fields or new proto messages if needed
- Refactor Go code
- Continue to develop Go PS features, such as checkpoint feature

Step 2. 
- Create a new `worker_gops.py` for only Go PS
- Add necessary util functions for `worker_gops.py`

Step 3.
[...]

PR #1724, opened 2020-02-11
modify proto message to support refactoring PS

QiJune:
For more details, please refer to #1722

### DIFF
--- a/elasticdl/proto/elasticdl.proto
+++ b/elasticdl/proto/elasticdl.proto
@@ -84,7 +84,7 @@ message Tensor {
 }
 
 message IndexedSlices {
-  Tensor concat_vectors = 1;
+  Tensor concat_tensors = 1;
   repeated int64 ids = 2;
 }
 
@@ -160,15 +160,6 @@ message PushGradientResponse {
   int32 model_version = 2;
 }
 
-// PS service
-service Pserver {
-  rpc pull_variable(PullVariableRequest) returns (PullVariableResponse);
-  rpc pull_embedding_vector(PullEmbeddingVectorRequest) returns (Tensor);
-  rpc push_model(Model) returns (google.protobuf.Empty);
-  rpc push_embedding_info(Model) returns (google.protobuf.Empty);
-  rpc push_gradient(PushGradientRequest) returns (PushGradientResponse);
-}
-
 message PullDenseParametersRequest {
   int32 version = 1;
 }
@@ -189,11 +180,17 @@ message PushGradientsResponse {
   int32 version = 2;
 }
 
-// GoPS service
-service GoPserver {
+// PS service
+service Pserver {
+  rpc pull_variable(PullVariableRequest) returns (PullVariableResponse);
+  rpc pull_embedding_vector(PullEmbeddingVectorRequest) returns (Tensor);
+  rpc push_model(Model) returns (google.protobuf.Empty);
+  rpc push_embedding_info(Model) returns (google.protobuf.Empty);
+  rpc push_gradient(PushGradientRequest) returns (PushGradientResponse);
+  // Following is the new rpc service described in concepts doc, after the
+  // refactoring work old rpc service will be removed safely.
   rpc pull_dense_parameters(PullDenseParametersRequest)
       returns (PullDenseParametersResponse);
   rpc pull_embedding_table(PullEmbeddingTableRequest) returns (Tensor);
-  rpc push_model(Model) returns (google.protobuf.Empty);
   rpc push_gradients(Model) returns (PushGradientsResponse);
 }
--- a/elasticdl/proto/elasticdl.proto
+++ b/elasticdl/proto/elasticdl.proto
@@ -83,6 +83,11 @@ message Tensor {
   TensorDtype dtype = 5;
 }
 
+message IndexedSlices {
+  Tensor concat_vectors = 1;
+  repeated int64 ids = 2;
+}
+
 message EmbeddingTableInfo {
   string name = 1;
   int64 dim = 2;
@@ -93,6 +98,8 @@ message Model {
   int32 version = 1;
   repeated Tensor param = 2;
   repeated EmbeddingTableInfo embedding_table_info = 3;
+  map<string, Tensor> dense_parameters = 4;
+  map<string, IndexedSlices> indexed_slices = 5;
 }
 
 message GetTaskRequest {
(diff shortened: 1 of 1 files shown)


mhaoli (approved):
Execellent!

lwpyr (approved):
Looks fine to me

elasticdl/proto/elasticdl.proto
```diff
@@ -83,16 +83,24 @@ message Tensor {
   TensorDtype dtype = 5;
 }
 
+message IndexedSlices {
+  Tensor concat_vectors = 1;
```
wangkuiyi: concat_vectors => concat_tensors

elasticdl/proto/elasticdl.proto
```diff
@@ -83,16 +83,24 @@ message Tensor {
...
+  repeated int64 ids = 2;
+}
+
 message EmbeddingTableInfo {
   string name = 1;
   int64 dim = 2;
   string initializer = 3;
+  TensorDtype dtype = 4;
```
wangkuiyi, line 96: I remember that @QiJune mentioned that we could reuse TensorFlow's protobuf messages. Why are we not doing it in this PR?

elasticdl/proto/elasticdl.proto
```diff
@@ -160,3 +168,32 @@ service Pserver {
...
+  repeated int64 ids = 2;
+}
+
+message PushGradientsResponse {
+  bool accepted = 1;
+  int32 version = 2;
+}
+
+// GoPS service
+service GoPserver {
```
wangkuiyi: The service is independent with programming languages.

wangkuiyi (changes requested):

elasticdl/proto/elasticdl.proto
QiJune: Done

elasticdl/proto/elasticdl.proto
QiJune: Done. Remove it.

And add the new rpc service described in concepts doc into current Pserver service. After the refactoring work is done, old rpc service will be remove safely.

elasticdl/proto/elasticdl.proto
QiJune: @wangkuiyi There are two things to clarify:

**1. Could we reuse TensorFlow's proto messages directly? or we could define it with our own**

I think the tensor proto message of Tensorflow and our new one in the concepts doc is almost the same.

Following is ours:

```proto
message Tensor {
  repeated int64 dim = 1;
  bytes content = 2;
  TensorDtype dtype = 3;
}
```

Following is Tensorflow's

```proto
message Tensor {
  DataType dtype = 1;
  TensorShapeProto tensor_shape = 2;
  int32 version_number = 3;
  // Serialized raw tensor content from either Tensor::AsProtoTensorContent or
  // memcpy in tensorflow::grpc::EncodeTensorToByteBuffer. This representation
  // can be used for all tensor types. The purpose of this representation is to
  // reduce serialization overhead during RPC call by avoiding serialization of
  // many repeated small items.
  bytes tensor_content = 4;

  // Type specific representations that make it easy to create tensor protos in
  // all languages.  Only the representation corresponding to "dtype" can
  // be set.  The values hold the flattened representation of the tensor in
  // row major order.

  // DT_FLOAT.
  repeated float float_val = 5 [packed = true];

  // DT_DOUBLE.
  repeated double double_val = 6 [packed = true];

  // DT_INT32, DT_INT16, DT_INT8, DT_UINT8.
  repeated int32 int_val = 7 [packed = true];
}
```
The bytes field could be used for all languages, and is more efficient during RPC.

The only big difference is that Tensorflow provides type specific representations that make it easy to create tensor protos in all languages. Maybe some language could not use bytes field conveniently. But Python/Go/C++ could work well with bytes filed. There is no need to add these fields to the tensor proto message.

So I proposal that we could use the proto mesage defined by ourselves.


We could reuse some in-memory data structure in Python, like `numpy.ndarray` and `tf.IndexedSlices`. Since our Tensor proto message and IndexedSlices proto message are one-to-one correspondence to `numpy.ndarray` and `tf.IndexedSlices`.

In Go, we could use the data structure defined in proto message directly. The in memory data structure is generated in `elasticdl.pb.go` already.

**2. How to refactor the code smoothly**

We want to refactor our code following the [concepts doc](https://github.com/sql-machine-learning/elasticdl/blob/develop/docs/designs/concepts.md).

Most of the basic proto messages are changed. This will break a lots of unit tests now. We have to modified a lot of code, inculding Go PS code/Python PS code/Python worker code. 

So one way is to add new messages like IndexedSlices, and add the new fields to current messages.

This is only a transition peroid. We will clean the proto messages after the refactoring is complete.

For example, the new Model proto mesages could be:

```proto
message Tensor {
  string name = 1;
  repeated int64 dim = 2;
  bytes content = 3;
  repeated int64 indices = 4;
  TensorDtype dtype = 5;
}

message IndexedSlices {
  Tensor concated_vectors = 1;
  repeated int64 ids = 2;
}

message Model {
  int32 version = 1;  // model updated times
  repeated Tensor param = 2;
  repeated EmbeddingTableInfo embedding_table_info = 3;
  map<string, Tensor> dense_parameters = 4;
  map<string, IndexedSlices> indexed_slices = 5;
}
```

And we also add the new rpc service to current `service Pserver` proto definition.

```proto
service Pserver {
  rpc pull_variable(PullVariableRequest) returns (PullVariableResponse);
  rpc pull_embedding_vector(PullEmbeddingVectorRequest) returns (Tensor);
  rpc push_model(Model) returns (google.protobuf.Empty);
  rpc push_embedding_info(Model) returns (google.protobuf.Empty);
  rpc push_gradient(PushGradientRequest) returns (PushGradientResponse);
  // Following is the new rpc service described in concepts doc, after the
  // refactoring work old rpc service will be removed safely.
  rpc pull_dense_parameters(PullDenseParametersRequest)
      returns (PullDenseParametersResponse);
  rpc pull_embedding_table(PullEmbeddingTableRequest) returns (Tensor);
  rpc push_gradients(Model) returns (PushGradientsResponse);
}
```

In this way, current code is not broken.

We could take this redundant proto message(which mixes legacy proto message and new proto message) as a start point.

Then, we could refactor our code by a lots of small PRs.

Besides, the refactoring work could also be done concurrently. For example, Python code refactoring and Go code refactoring could be done simultaneously.

skydoorkai: I agree that we use our own simplified message definition. The other reason is that The TensorFlow message includes many fields that ElasticDL will not use. If included, this may confuse future developers and users. We can add fields that are really needed in ElasticDL if new features should be added.

wangkuiyi (approved):

PR #1736, opened 2020-02-14
Support reusing the built image while submitting ElasticDL job

brightcoder01:


elasticdl/python/common/args.py
```diff
@@ -107,7 +107,13 @@ def add_common_params(parser):
         help="The repository for generated Docker images, if set, the image "
         "is also pushed to the repository",
     )
-    parser.add_argument("--image_base", help="Base Docker image.")
+    parser.add_argument("--image_base", default="", help="Base Docker image.")
+    parser.add_argument(
+        "--image_name",
+        default="",
+        help="The image for this job. If set, "
```
skydoorkai: help="If set, use this image instead of building one.",

elasticdl/python/elasticdl/api.py
```diff
@@ -31,39 +35,44 @@ def train(args):
...
-            build_arguments_from_parsed_result(
-                args, filter_args=["model_zoo", "cluster_spec", "worker_image"]
-            )
+    container_args = [
+        "-m",
+        "elasticdl.python.master.main",
+        "--worker_image",
+        image_name,
+        "--model_zoo",
+        _model_zoo_in_docker(model_zoo),
```
skydoorkai: If not building image, model_zoo should not set from `_model_zoo_in_docker`.

elasticdl/python/elasticdl/api.py
```diff
@@ -31,39 +35,44 @@ def train(args):
...
+        "--cluster_spec",
+        _cluster_spec_def_in_docker(args.cluster_spec),
```
skydoorkai, line 50: Also, if not building image, `cluster_spec` should not set from `_cluster_spec_def_in_docker `.

elasticdl/python/common/args.py
brightcoder01: Good suggestion, updated.

elasticdl/python/elasticdl/api.py
brightcoder01: Good suggestion, updated.

elasticdl/python/elasticdl/api.py
brightcoder01: Good suggestion, updated.

skydoorkai (approved):

skydoorkai:
Also need to merge into branch-0.2.

PR #1742, opened 2020-02-17
clean pull embedding vector rpc in Python

QiJune:


elasticdl/python/worker/worker.py
```diff
@@ -32,9 +32,11 @@
     Tensor,
     emplace_tensor_pb_from_ndarray,
     serialize_tensor,
-    tensor_pb_to_ndarray,
 )
-from elasticdl.python.common.tensor_utils import deduplicate_indexed_slices
+from elasticdl.python.common.tensor_utils import (
+    deduplicate_indexed_slices,
```
brightcoder01, line 37: I didn't see the usage of deduplicate_indexed_slices in this file

QiJune: The `deduplicate_indexed_slices ` is used at 

https://github.com/sql-machine-learning/elasticdl/blob/0a6858da862f6d3f15f41b2c4693078972f848c4/elasticdl/python/worker/worker.py#L513-L519

elasticdl/python/ps/servicer.py
```diff
@@ -78,16 +79,15 @@ def pull_variable(self, request, _):
         res.model_init_status = True
         return res
 
-    def pull_embedding_vector(self, request, _):
-        ret = elasticdl_pb2.Tensor()
+    def pull_embedding_table(self, request, _):
+        res = tensor_pb2.TensorProto()
```
workingloong: Maybe  "res" -> "result" ?

elasticdl/python/worker/worker.py
```diff
@@ -345,7 +347,7 @@ def get_model(self):
         self._model_version = max(self._model_versions_from_ps)
         self._timing.end_record_time("get_model")
 
-    def pull_embedding_vector(self, layer_name, embedding_ids):
+    def pull_embedding_table(self, layer_name, embedding_ids):
         """Pulls and returns embedding vectors ordered by the embedding ids."""
```
workingloong, line 351: Why do you rename `pull_embedding_table`? The `pull_embedding_vector` is more suitable for the docstring.

elasticdl/python/worker/worker.py
brightcoder01: I'm curious why the import is added in the change, but we can't see the usage in the new added change.
If the two methods are already used in the worker.py, they should be already imported.

elasticdl/python/ps/servicer.py
QiJune: Done

elasticdl/python/worker/worker.py
QiJune: It's actually pulling some embedding vectors from an embedding table. So maybe `pull_embedding_vectors` is more accurate.
Rename it.

skydoorkai (approved):

PR #1747, opened 2020-02-17
Clear session before cloning model to train.

workingloong:
The  `tf.keras.models.clone_model` will return a new model instance and the variable name of the new model will be added a `suffix`. For example:
```
import tensorflow as tf
def custom_model_with_embedding_layer():
    inputs = tf.keras.layers.Input(shape=(4,), name="x")
    embedding = tf.keras.layers.Embedding(300, 2)(inputs)
    outputs = tf.keras.layers.Dense(1)(embedding)
    return tf.keras.models.Model(inputs, outputs)


model = custom_model_with_embedding_layer()
print(model.inputs)
    
model_1 = tf.keras.models.clone_model(model)
print(model_1.inputs)
```

[<tf.Tensor 'x:0' shape=(None, 4) dtype=float32>]
[<tf.Tensor 'x_1:0' shape=(None, 4) dtype=float32>]

### DIFF
--- a/elasticdl/python/common/model_handler.py
+++ b/elasticdl/python/common/model_handler.py
@@ -202,6 +202,8 @@ class ParameterServerModelHandler(ModelHandler):
         """Replace the tf.keras.layers.Embedding layer in the model with
         an elasticdl.layers.Embedding layer in ParameterServerStrategy.
         """
+        # clear keras model session to avoid clutter from old models/layers.
+        tf.keras.backend.clear_session()
         if type(model) == tf.keras.Sequential or model._is_graph_network:
             model = self._clone_model_with_edl_embedding(model)
         else:


elasticdl/python/common/model_handler.py
```diff
@@ -202,6 +202,8 @@ def get_model_to_train(self, model):
         """Replace the tf.keras.layers.Embedding layer in the model with
         an elasticdl.layers.Embedding layer in ParameterServerStrategy.
         """
+        # clear keras model session to avoid clutter from old models/layers.
+        tf.keras.backend.clear_session()
```
skydoorkai, line 206: The old model is only defined before but not used. So the model definition will also have some side effects?

skydoorkai (approved):

brightcoder01 (approved):

elasticdl/python/common/model_handler.py
workingloong: The side effects may be that the model weights are created repeatedly and the memory is wasted.

PR #1749, opened 2020-02-18
clean push gradients rpc service in Python

QiJune:
Since we use `map` instead of `repeated` proto message to store gradients, the gradient must be unique in worker side.

This PR add a logic in worker to merge gradients of a same parameter.


workingloong:
LGTM

workingloong (approved):

PR #1752, opened 2020-02-18
clean report evaluation rpc service in Python

QiJune:


elasticdl/python/worker/worker.py
```diff
@@ -574,12 +572,9 @@ def report_evaluation_metrics(self, model_outputs, labels):
         req = elasticdl_pb2.ReportEvaluationMetricsRequest()
         for name, output in model_outputs.items():
             output = np.concatenate(output)
-            emplace_tensor_pb_from_ndarray(
-                req.model_outputs, output, name=name
-            )
+            req.model_outputs[name].CopyFrom(ndarray_to_pb(output))
```
brightcoder01: Will it bring one more memory copy?

skydoorkai (approved):

elasticdl/python/worker/worker.py
QiJune: Yes, you are right.

Remove it.

PR #1753, opened 2020-02-18
Reconstruct go ps: add embedding table definition

lwpyr:


elasticdl/pkg/commonnew/embedding_tabel.go
```diff
@@ -0,0 +1,70 @@
+package commonnew
```
QiJune: The file name `embedding_tabel.go` --> `embedding_table.go`

elasticdl/pkg/commonnew/embedding_tabel.go
```diff
@@ -0,0 +1,70 @@
...
+// EmbeddingTable struct
+type EmbeddingTable struct {
+	Dim              int64
+	Initializer      string
+	EmbeddingVectors map[int64]*Tensor
+	Dtype            types_go_proto.DataType
+}
+
+// NewIndexedSlices return proto.IndexedSlices
+func NewIndexedSlices(t *Tensor, indices []int64) *proto.IndexedSlices {
```
QiJune: Where is this function used?
It seems that it's a very thin wrapper, we could define a `proto.IndexedSlices` directly.

elasticdl/pkg/commonnew/embedding_tabel.go
```diff
@@ -0,0 +1,70 @@
...
+		if index == -1 {
+			continue
+		}
+		tensor.SetRow(int64(i), e.GetEmbeddingVector(index))
+	}
+	return tensor
+}
+
+// SetEmbeddingVectors sets (indices, value) pair to embedding vector
+func (e *EmbeddingTable) SetEmbeddingVectors(idxslice *proto.IndexedSlices) error {
```
QiJune: How about pass `indices` and `values` parameters directly?

`func (e *EmbeddingTable) SetEmbeddingVectors(indices []int64, vaules* Tensor)`

elasticdl/pkg/commonnew/embedding_tabel.go
```diff
@@ -0,0 +1,70 @@
+package commonnew
+
+import (
+	"elasticdl.org/elasticdl/pkg/proto"
+	"github.com/tensorflow/tensorflow/tensorflow/go/core/framework/types_go_proto"
+)
+
+// EmbeddingTable struct
+type EmbeddingTable struct {
```
QiJune: We need a function to serialize a `EmbeddingTable` instance to a `proto.IndexedSlices`. It will be used when making checkpoint in PS.

elasticdl/pkg/commonnew/embedding_tabel.go
```diff
@@ -0,0 +1,70 @@
...
+
+// NewEmbeddingTable creates an embedding table instance
+func NewEmbeddingTable(dim int64, initializer string, dtype types_go_proto.DataType) *EmbeddingTable {
+	var e = EmbeddingTable{
+		Dim:              dim,
+		Initializer:      initializer,
+		EmbeddingVectors: make(map[int64]*Tensor),
+		Dtype:            dtype,
+	}
+	return &e
```
QiJune:
```go
return &EmbeddigTable {...}
```

elasticdl/pkg/commonnew/embedding_tabel.go
```diff
@@ -0,0 +1,70 @@
...
+}
+
+// GetEmbeddingVectors returns COPYS of embedding vectors giving an array of indices
+func (e *EmbeddingTable) GetEmbeddingVectors(indices []int64) *Tensor {
+	dim := []int64{int64(len(indices)), e.Dim}
+	tensor := NewEmptyTensor(dim, e.Dtype)
+	for i, index := range indices {
+		if index == -1 {
+			continue
+		}
```
QiJune: What's the logic behind the code `if index == -1`?

elasticdl/pkg/commonnew/embedding_tabel.go
lwpyr: fixed

elasticdl/pkg/commonnew/embedding_tabel.go
lwpyr: fixed

elasticdl/pkg/commonnew/embedding_tabel.go
lwpyr: The code is writing in this way because. `EmbeddingTable` in memory is corresponding to `IndexedSlice` in proto, and It seems we don't need to unpack an `IndexedSlice` inside the Parameter server.  Any further questions could be raised.

elasticdl/pkg/commonnew/embedding_tabel.go
lwpyr: The logic here is, there will be a function `MergeIndexedSlice`, which will merge the same index column in-place, and the duplicate indices will be marked as -1, which indicates this should be ignored.

elasticdl/pkg/commonnew/embedding_tabel.go
lwpyr: This function is moved to Tensor. This is actually the Serializer function.

elasticdl/pkg/commonnew/embedding_tabel.go
lwpyr: added to tensor.go

lwpyr: for most of the time, we don't need to serialize all the data inside `EmbeddingTable` to an indxedSlice.
So we could use like this:
```
//given: indices []int64, table *EmbeddingTable
tensor := table.GetEmbeddingVectors(indices)
tensor.SerializerToIndexedSlice(indices)
```
`GetEmbeddingVectors` funtion will be independently used in optimizer.

elasticdl/pkg/commonnew/embedding_tabel.go
lwpyr: The logic here is, there will be a function `MergeIndexedSlice`, which will merge the same index column in-place, and the duplicate indices will be marked as -1, which indicates this should be ignored.

QiJune: @lw396285v It seem a little hack here.

And we do not to merge gradients in PS side. I propose a #1749 , this PR add a logic in worker in worker to merge gradients of a same parameter already. So the gradients in PS are always unique.

So, please remove the `if index == -1` here int this PR. If we really need it,  let's find a more elegant way.

lwpyr: fixed

elasticdl/pkg/commonnew/tensor.go
```diff
@@ -168,3 +169,12 @@ func (t *Tensor) SerializeTensor() *tensor_go_proto.TensorProto {
 		Dtype:         t.Dtype,
 	}
 }
+
+// SerializeToIndexedSlices return proto.IndexedSlices
+func (t *Tensor) SerializeToIndexedSlices(indices []int64) *proto.IndexedSlices {
```
QiJune, line 182: Need to check the Tensor dim and indices length here. If the dims do not match, then return a `nil` pointer.

elasticdl/pkg/commonnew/tensor.go
```diff
@@ -168,3 +169,12 @@ func (t *Tensor) SerializeTensor() *tensor_go_proto.TensorProto {
...
+	var i = proto.IndexedSlices{
+		ConcatTensors: t.SerializeToTensor(),
+		Ids:           indices,
+	}
+	return &i
```
QiJune:
```go
return &proto.IndexedSlices {...}
```

elasticdl/pkg/commonnew/embedding_table.go
```diff
@@ -0,0 +1,54 @@
...
+	tensor := NewEmptyTensor(dim, e.Dtype)
+	for i, index := range indices {
+		tensor.SetRow(int64(i), e.GetEmbeddingVector(index))
+	}
+	return tensor
+}
+
+// SetEmbeddingVectors sets (indices, value) pair to embedding vector
+func (e *EmbeddingTable) SetEmbeddingVectors(idxslice *proto.IndexedSlices) error {
+	idxsliceTensor := DeserializeFromTensorPB(idxslice.ConcatTensors)
```
QiJune, line 49: Need to check idxsliceTensor is null or not.

```go
if idxsliceTensor == nil:
    return fmt.Errorf("Embedding vectors dim not match")
```

lwpyr: fixed

elasticdl/pkg/commonnew/tensor.go
lwpyr: fixed

elasticdl/pkg/commonnew/tensor.go
lwpyr: fixed

elasticdl/pkg/commonnew/embedding_table.go
QiJune: Emm, I mean check the result after DeserializeFromTensorPB. Just like:

```go
idxsliceTensor := DeserializeFromTensorPB(idxslice.ConcatTensors)
if idxsliceTensor == nil {
    return fmt.Errorf("Embedding vectors dim not match")
}
for i, index := range idxslice.Ids {
    value := e.GetEmbeddingVector(index)
    copy(value.Buffer, idxsliceTensor.GetRow(int64(i)).Buffer)
}
return nil
```

elasticdl/pkg/commonnew/tensor.go
QiJune: It seems that it's still not fixed yet.

lwpyr: fixed now...

elasticdl/pkg/commonnew/embedding_table.go
lwpyr: ok, got it!

elasticdl/pkg/commonnew/tensor.go
```diff
@@ -151,8 +152,19 @@ func DeserializeTensorPB(pb *tensor_go_proto.TensorProto) *Tensor {
...
+func (t *Tensor) IsValid() bool {
+	if DimProduct(t.Dims) != int64(len(t.Buffer)/int(DtypeSize[t.Dtype])) {
+		return false
+	}
+	return true
+}
+
+// SerializeToTensor transforms tensor to pb
+func (t *Tensor) SerializeToTensor() *tensor_go_proto.TensorProto {
+	if !t.IsValid() {
```
QiJune: It is no need to to check here.

elasticdl/pkg/commonnew/tensor.go
QiJune: Sorry, I do not express it clearly.
It is saied to check the dim of Tensor and indices. The tensor must be a 2-dim tensor, and the row dim of the tensor  is equal to `len(indices)`.

QiJune (approved):
Excellent!

PR #1754, opened 2020-02-18
Add a callback to stop training by max_steps.

workingloong:
After the master receives a completed training task, we can add up the training steps (the count of mini-batches) in the task. If the total training steps > max_steps, we will set `callback_list.model.stop_training=True` and then clear `TaskDispatcher.todo` list. However, the master must wait that the tasks in`TaskDispatcher.doing` are completed and the training steps may be a little bigger than `max_steps`.


skydoorkai:
What if the training is loaded from a checkpoint? `_completed_steps` will be 0. Probably assign this value from model_version if loaded from a checkpoint.

workingloong:
> What if the training is loaded from a checkpoint? `_completed_steps` will be 0. Probably assign this value from model_version if loaded from a checkpoint.

Good suggestion. We can get the model_version from the checkpoint file.

elasticdl/python/common/save_utils.py
```diff
@@ -259,3 +259,11 @@ def restore_params_from_checkpoint(checkpoint_dir, shard_index, shard_num):
         parameters.embedding_params.update(embedding_tables)
         parameters.version = version
         return parameters
+
+    @staticmethod
+    def get_version_from_checkpoint(checkpoint_dir):
+        variable_shard_files = os.listdir(checkpoint_dir)
+        shard_file_path = os.path.join(checkpoint_dir, variable_shard_files[0])
```
brightcoder01, line 270: From this code, we only load the first file in variable_shard_files, what about other files?

elasticdl/python/master/master.py
```diff
@@ -127,6 +144,18 @@ def __init__(self, args):
         self._should_stop = False
         self._exit_code = 0
 
+    def _set_completed_steps_by_checkpoint(self, checkpoint_dir_for_init):
```
brightcoder01, line 150: When will we call this method?

elasticdl/python/common/save_utils.py
workingloong: Each shard file has the version and the versions in shard files are the same. So, we only need to load one file to get the version.

elasticdl/python/master/master.py
workingloong: We should call this method using `args.checkpoint_dir_for_init` after initializing callbacks. I have added the code to call it in the next commit.

elasticdl/python/master/master.py
```diff
@@ -127,6 +147,18 @@ def __init__(self, args):
...
+        if not checkpoint_dir_for_init:
+            return
+
+        CheckpointSaver.check_checkpoint_valid(checkpoint_dir_for_init)
```
brightcoder01: If check_checkpoint_valid returns false, we need return directly instead of executing the following logic

workingloong: Yes, it may be better to raise an error if checkpoint dir is not None and invalid.

brightcoder01 (approved):

elasticdl/python/master/master.py
skydoorkai: If not valid, return

workingloong: Yes, I have added the code to raise a `ValueError` if invalid.

PR #1755, opened 2020-02-18
Retry to read records from MAXCOMPUTE

workingloong:
The MAXCOMPUTE reader downloads records by https request and the timeout may happen sometimes. So it is better to retry if an error happens.

### DIFF
--- a/elasticdl/python/data/odps_io.py
+++ b/elasticdl/python/data/odps_io.py
@@ -243,8 +243,8 @@ class ODPSReader(object):
     def record_generator_with_retry(
         self, start, end, columns=None, max_retries=3
     ):
-        """Retry to read records from ODPS table by using to avoid
-        the network instability.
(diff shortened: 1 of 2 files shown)


elasticdl/python/data/odps_io.py
```diff
@@ -240,6 +240,29 @@ def read_batch(self, start, end, columns=None, max_retries=3):
                 time.sleep(5)
                 retry_count += 1
 
+    def record_generator_with_retry(
+        self, start, end, columns=None, max_retries=3
+    ):
+        """Retry to read records from ODPS table by using to avoid
```
skydoorkai: Wrap record_generator with retry to avoid ODPS table read failure due to network instability.

workingloong: Yeah!

brightcoder01 (approved):

PR #1756, opened 2020-02-18
Use go mod to manage packages

lwpyr:
Any one is welcome to modify the code, because I'm not quite familiar with the CI settings


QiJune (approved):
Excellent!

PR #1757, opened 2020-02-18
Remove the extra "/model_zoo" when copy the model zoo files. 

### LINKED ISSUE #1740: Copy the user defined model zoo files into root  in the docker image.
Sometimes, we need to define multiple files for a model. For example:
```
model_zoo
    |----mnist_model
        |----__init__.py
        |----mnist_model_functional.py
        |----custom_layer.py
```
We can import the module from `custom_layer.py` in `mnist_model_functional.py` using
`from model_zoo.custom_layer import CustomLayer`.  

However, we copy the `model_zoo` directory into `model_zoo/model_zoo` when we build the image. And the directory structure is
```
model_zoo
    |----model_zoo
        |----mnist_model
            |----__init__.py
            |----mnist_model_functional.py
            |----custom_layer.py
```
So, the user must use `from model_zoo.model_zoo.custom_layer import CustomLayer` to import.

workingloong:
Fix #1740

### DIFF
--- a/elasticdl/python/elasticdl/image_builder.py
+++ b/elasticdl/python/elasticdl/image_builder.py
@@ -139,7 +139,7 @@ def _create_dockerfile(
 ):
     HEAD = """
 FROM {BASE_IMAGE} as base
-ENV PYTHONPATH=/:/model_zoo
+ENV PYTHONPATH=/
 """
     if elasticdl:
         HEAD = """
@@ -154,14 +154,14 @@ COPY %s/elasticdl /elasticdl
 RUN pip install -r /elasticdl/requirements.txt \
   --extra-index-url="${EXTRA_PYPI_INDEX}"
 RUN make -f /elasticdl/Makefile
-COPY {MODEL_ZOO} /model_zoo/{MODEL_ZOO}
+COPY {MODEL_ZOO} /{MODEL_ZOO}
 """
     REMOTE_ZOO = """
 RUN pip install -r /elasticdl/requirements.txt \
   --extra-index-url="${EXTRA_PYPI_INDEX}"
 RUN make -f /elasticdl/Makefile
 RUN apt-get update && apt-get install -y git
-RUN git clone --recursive {MODEL_ZOO} /model_zoo
+RUN git clone --recursive {MODEL_ZOO} /
 """
     pr = urlparse(model_zoo)
     if not pr.path:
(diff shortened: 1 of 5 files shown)


terrytangyuan (approved):
LGTM as long as all existing examples pass

elasticdl/python/elasticdl/image_builder.py
```diff
@@ -154,14 +154,14 @@ def _create_dockerfile(
 RUN pip install -r /elasticdl/requirements.txt \
   --extra-index-url="${EXTRA_PYPI_INDEX}"
 RUN make -f /elasticdl/Makefile
-COPY {MODEL_ZOO} /model_zoo/{MODEL_ZOO}
+COPY {MODEL_ZOO} /{MODEL_ZOO}
```
skydoorkai, line 157: I think the correct way is that in image_builder, we copy `model_def` into `/model_zoo`. 
For example, on the client, `cifar10_functional_api` is in `some_dir`. The user set `model_zoo=some_dir`.
After image_builder, `cifar10_functional_api` is under `/model_zoo` in the docker image.

Then in master pod argument, set `model_zoo=/model_zoo`.

workingloong: If `cifar10_functional_api` is in `some_dir`, the user will write 
`from some_dir.cifar10_function_api.some_module import some_class` to import `some_class` from `some_module.py`. 
If `cifar10_functional_api` is under `/model_zoo` in the docker image, the import error will happen and the corret import code in the image is
`from model_zoo.cifar10_function_api.some_module import some_class`

scripts/client_test.sh
```diff
@@ -65,7 +65,7 @@ elif [[ "$JOB_TYPE" == "predict" ]]; then
       --image_base=elasticdl:ci \
       --model_zoo=model_zoo \
       --model_def=mnist_functional_api.mnist_functional_api.custom_model \
-      --checkpoint_dir_for_init=model_zoo/${MNIST_CKPT_DIR}/version-100 \
+      --checkpoint_dir_for_init=/${MNIST_CKPT_DIR}/version-100 \
```
brightcoder01, line 68: We can also update the example command in elasticdl_local.md and elasticdl_on_prem_cluster.md.
The model zoo parameter is set as follows, we need update them too
```
 --model_zoo=/model_zoo/model_zoo \
```

elasticdl/python/elasticdl/image_builder.py
skydoorkai: ok, then I assume the user will use the root directory of `some_dir` as python directory.

scripts/client_test.sh
workingloong: Yes!

PR #1760, opened 2020-02-19
add timeout mechanism

QiJune:
We will statistic average complete time of history sucessful tasks. If current doing task exceeds `2 * average time`, the worker who executes the doing task will be killed, and the task will be moved to todo list.


elasticdl/python/master/task_dispatcher.py
```diff
@@ -265,7 +266,7 @@ def report(self, request, success):
         task_id = request.task_id
         evaluation_task_completed = False
         with self._lock:
-            _, task = self._doing.pop(task_id, (-1, None))
+            _, task, get_time = self._doing.pop(task_id, (-1, None, -1))
```
workingloong: Rename `get_time` to `start_time` ?

elasticdl/python/master/servicer.py
```diff
@@ -19,6 +20,7 @@ def __init__(
         self._version = 0
 
         self._evaluation_service = evaluation_service
+        self._task_complete_times = [36000] * 20
```
workingloong: Is `20` the task count to calculate the average process time? If yes, it is more readable to use a constant such as `STATISTICS_TASK_QUEUE_SIZE = 20`

elasticdl/python/master/master.py
```diff
@@ -430,3 +436,21 @@ def _create_instance_manager(self, args):
...
+            cur_time = time.time()
+            threshold = (
+                2 * self.master_servicer.get_average_task_complete_time()
+            )
+            for task_id, (worker_id, task, get_time) in doing_tasks.items():
+                if (cur_time - get_time) > threshold:
+                    self.logger.info(
+                        "worker %d timeout, relaunch it" % worker_id
+                    )
+                    self.task_d.recover_tasks(worker_id)
```
workingloong: Do we need to set the recovery times for a task? IMO, the task should fail if the times to recover it due timeout exceed a threshold. Now, we will recover a bad task repeatedly.

QiJune: Good idea. It's another topic, and will be fixed in next PR.

elasticdl/python/master/servicer.py
QiJune: Done

elasticdl/python/master/task_dispatcher.py
QiJune: Done

elasticdl/python/master/servicer.py
```diff
@@ -67,7 +71,9 @@ def report_task_result(self, request, _):
             logger.warning("Worker reported error: " + request.err_message)
             self._task_d.report(request, False)
         else:
-            self._task_d.report(request, True)
+            complete_time = self._task_d.report(request, True)
+            self._task_complete_times.pop(0)
+            self._task_complete_times.append(complete_time)
```
skydoorkai: `self._task_complete_times` is from the last 20 completed tasks.
If training tasks take a long time, but the evaluation task can finish very quickly, and the worker gets 20 evaluation tasks in sequence. This will make the mean of `self._task_complete_times` very small, and timeout following training tasks.
Vice versa

QiJune: How about use all the historial task complete time?

skydoorkai: Or we need to distinguish `training` task and `eval` task.
`total_training_time`
`num_training_task`
`total_eval_time`
`num_eval_task`
```
if (num_training_task > threshold(such as 20) and
     current_training_task_running_time > total_training_time / num_training_task
):
  timeout the task
 ```

Comment (QiJune) [reply elasticdl/python/master/servicer.py]: Done

File: elasticdl/python/master/master.py [hunk 5]
```diff
@@ -430,3 +436,23 @@ def _create_instance_manager(self, args):
...
+        while True:
+            doing_tasks = self.task_d._doing.copy()
+            cur_time = time.time()
+            avg_time = self.master_servicer.get_average_task_complete_time()
+            for task_id, (worker_id, task, start_time) in doing_tasks.items():
+                if task.type in [
+                    elasticdl_pb2.TRAINING,
+                    elasticdl_pb2.EVALUATION,
+                ]:
+                    if (cur_time - start_time) > 3 * avg_time[task.type]:
```
skydoorkai, line 454: If a worker is in the process of a training task, there come some eval tasks, the worker will process these eval tasks first. After the eval tasks are done, the worker will continue the training task. Thus this training task may timeout if using `cur_time - start_time` to determine timeout. If the master has this worker's last report time (report_task_result or report_evaluation_metrics), using `cur_time - worker_last_report_time` will solve this problem.

QiJune: Done. Add an interface `get_worker_liveness_time`

elasticdl/python/master/servicer.py
```diff
@@ -83,3 +100,22 @@ def report_version(self, request, _):
...
+            return {
+                elasticdl_pb2.TRAINING: statistics.mean(
+                    self._task_complete_times[elasticdl_pb2.TRAINING]
+                ),
+                elasticdl_pb2.EVALUATION: statistics.mean(
+                    self._task_complete_times[elasticdl_pb2.EVALUATION]
+                ),
+            }
+
+    def get_worker_liveness_time(self, worker_id):
```
skydoorkai, line 120: `get_worker_liveness_time` may be called before `self._worker_liveness_time[worker_id]` is set.
Init `self._worker_liveness_time[worker_id]` in `get_task` if not already initialized.

skydoorkai: Also, we may need to update `self._worker_liveness_time` in `get_task`, as the worker may receive `WAIT` task and doing nothing for a while, then get an evaluation task. This task may timeout as `self._worker_liveness_time` is not updated for a long time.

QiJune: Done. Update the liveness time inside get_task

skydoorkai (approved):

PR #1761, opened 2020-02-19
Reconstruct go ps: add kernelnew package, add IndexedTensor

lwpyr:
* add kernelnew package
* add IndexedTensor definition

### DIFF
--- a/elasticdl/pkg/commonnew/tensor.go
+++ b/elasticdl/pkg/commonnew/tensor.go
@@ -134,6 +134,14 @@ func Slice(t *Tensor) interface{} {
 	return val.Interface()
 }
 
+// IsValid check tensor validity
+func (t *Tensor) IsValid() bool {
+	if DimProduct(t.Dims) != int64(len(t.Buffer)/int(DtypeSize[t.Dtype])) {
+		return false
+	}
+	return true
+}
+
 // DeserializeFromTensorPB transforms pb to tensor
 func DeserializeFromTensorPB(pb *tensor_go_proto.TensorProto) *Tensor {
 	pbDim := pb.GetTensorShape().GetDim()
@@ -150,14 +158,6 @@ func DeserializeFromTensorPB(pb *tensor_go_proto.TensorProto) *Tensor {
 		Dims:   dims,
 		Dtype:  pb.GetDtype(),
 	}
-}
-
-// IsValid check tensor validity
-func (t *Tensor) IsValid() bool {
-	if DimProduct(t.Dims) != int64(len(t.Buffer)/int(DtypeSize[t.Dtype])) {
-		return false
-	}
-	return true
 }
 
 // SerializeToTensor transforms tensor to pb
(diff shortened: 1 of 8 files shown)


lwpyr:
The kernel code is just modified from the original one, so this is  not a large PR.

elasticdl/pkg/commonnew/tensor.go
```diff
@@ -178,13 +178,35 @@ func (t *Tensor) SerializeToTensor() *tensor_go_proto.TensorProto {
 	}
 }
 
+// IndexedTensor : IndexedSlice in memory representation
+type IndexedTensor struct {
```
QiJune, line 182: We could just follow proto message definition, name it to `IndexedSlices`

```go
type IndexedSlices struct {
    ConcatTensors *Tensor
    IDs []int64
}
```

elasticdl/pkg/kernelnew/capi/kernel_api.cc
```diff
@@ -0,0 +1,63 @@
+#include "kernel_api.h"
+
+#include <cmath>
+#include <eigen3/Eigen/Dense>
+
+void AddTo(float* a, float* b, long long size) {
```
QiJune: AddTo is not used, please remove it.

elasticdl/pkg/kernelnew/kernel.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+	gradPtr := (*C.float)(unsafe.Pointer(&grad.Buffer[0]))
+	paramPtr := (*C.float)(unsafe.Pointer(&param.Buffer[0]))
+	length := len(grad.Buffer) / int(commonnew.DtypeSize[grad.Dtype])
+	C.SGD(gradPtr, paramPtr, C.float(lr), C.longlong(length))
+	return nil
+}
+
+// SparseSGD kernel
+func SparseSGD(grad *commonnew.IndexedTensor, param *commonnew.EmbeddingTable, lr float32) error {
```
QiJune, line 22: There are three cases:
- apply dense gradient to dense parameter
- apply sparse gradient to dense parameter. The gradient of native keras embedding layer is sparse, but the embedding parameter is dense.
- apply sparse gradient to sparse parameter 

So, we need an extra SGDKernel

elasticdl/pkg/kernelnew/kernel.go
```diff
@@ -0,0 +1,80 @@
...
+
+// SparseAdam kernel
+func SparseAdam(grad *commonnew.IndexedTensor, param *commonnew.EmbeddingTable, m *commonnew.EmbeddingTable, v *commonnew.EmbeddingTable, lr float32, step int64, beta1 float32, beta2 float32, epsilon float32, amsgrad bool, maxSquare *commonnew.EmbeddingTable) error {
+	if grad.Dims[1] != param.Dim {
+		return fmt.Errorf("grad width is not equal to embedding dim")
+	}
+	for i, index := range grad.Indices {
+		if index == -1 {
+			continue
+		}
```
QiJune: Please remove the code

```go
if index == -1 {
    continue
}
```

elasticdl/pkg/commonnew/tensor.go
lwpyr:
```
Tensor -> TensorProto
IndexedSlice -> IndexedSliceProto
```
could this be better?

QiJune: Since it's in memory representation, I think Proto is not needed.

QiJune: And here, we could use `tensorflow.TensorProto` and `proto.IndexedSlices` directly. 

Pros:
- no need to define extra in memory data structures, `Tensor`/`IndexedSlices`
- no need to define extra serialize/deserialize functions, `SerializeTensor`/`DeserializeTensor`/`SerializeIndexedSlices`/`DeserializeIndexedSlices`

Cons:
- tensorflow.TensorShapeProto is not convenient to use directly, we need to add two extra function, `SetDim`/`GetDim`

elasticdl/pkg/kernelnew/kernel.go
lwpyr: fixed

elasticdl/pkg/kernelnew/kernel.go
lwpyr: fixed

elasticdl/pkg/kernelnew/capi/kernel_api.cc
lwpyr: fixed

elasticdl/pkg/commonnew/tensor.go
QiJune: I find that it's not convenient to write codes without defining in memory data structures. So let's move on.

And for the struct name, How about

```go
type struct IndexedSlices {
    ConcatTensors *Tensor
    IDs []int64
} 
```

QiJune (approved):
Excellent!

PR #1762, opened 2020-02-19
Recover tasks when a pod fails

skydoorkai:
Before this PR, only killed pods will have their corresponding tasks recovered. If a pod fails due to some error inside the worker, its tasks will not be recovered, and the training will never stop due to the `doing` task set will remain non-empty.

### DIFF
--- a/elasticdl/python/master/k8s_instance_manager.py
+++ b/elasticdl/python/master/k8s_instance_manager.py
@@ -231,32 +231,33 @@ class InstanceManager(object):
         with self._lock:
             if pod_name in self._failed_pods:
                 return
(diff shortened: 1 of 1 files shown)


QiJune:
It seems that timeout mechanism could fix this issue.

If the worker failed due to some reason, when it reaches timeout thredshold, the worker will be killed by master. At this time, the task will be recovered.

skydoorkai:
> It seems that timeout mechanism could fix this issue.
> 
> If the worker failed due to some reason, when it reaches timeout thredshold, the worker will be killed by master. At this time, the task will be recovered.

Yes, timeout can handle this case. Probably just leave it to timeout.

PR #1767, opened 2020-02-21
Clean go package

lwpyr:


elasticdl/docker/scripts/install-go.bash
```diff
@@ -9,12 +9,4 @@ curl --silent ${GO_MIRROR_URL}/go1.13.4.linux-amd64.tar.gz | tar -C /usr/local -
 go env -w GO111MODULE=on
 go env -w GOPROXY=https://goproxy.io,direct
 
-go get github.com/golang/protobuf/protoc-gen-go
-go get golang.org/x/lint/golint
-go get golang.org/x/tools/cmd/goyacc
-go get golang.org/x/tools/cmd/cover
-go get github.com/mattn/goveralls
-go get github.com/rakyll/gotest
-go get github.com/wangkuiyi/goyaccfmt
```
QiJune, line 18: These go packages are some command line tools need to install. For example, golint is used in go code format. We should install them first.

lwpyr: yep

QiJune (approved):
Excellent!

PR #1768, opened 2020-02-21
Refine embedding table info

QiJune:
This PR does two things:
- rename `EmbeddingTableInfo` to `EmbeddingTableDesc` in Model messge
- set dtype to float32 when worker pushing embedding table desc. Otherwise, GO PS could not create an embedding table correctly.

### DIFF
--- a/elasticdl/pkg/ps/model.go
+++ b/elasticdl/pkg/ps/model.go
@@ -49,7 +49,7 @@ func (model *Model) SetEmbeddingTableInfo(info *proto.EmbeddingTableInfo) {
 
 // InitFromModelPB inits the model from model PB
 func (model *Model) InitFromModelPB(pb *proto.Model) error {
-	for _, v := range pb.EmbeddingTableInfo {
+	for _, v := range pb.EmbeddingTableInfos {
 		model.SetEmbeddingTableInfo(v)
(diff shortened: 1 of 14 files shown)


lwpyr (approved):
It looks like some modification over names, approved.

elasticdl/pkg/ps/model.go
```diff
@@ -49,7 +49,7 @@ func (model *Model) SetEmbeddingTableInfo(info *proto.EmbeddingTableInfo) {
 
 // InitFromModelPB inits the model from model PB
 func (model *Model) InitFromModelPB(pb *proto.Model) error {
-	for _, v := range pb.EmbeddingTableInfo {
+	for _, v := range pb.EmbeddingTableInfos {
```
skydoorkai, line 52: `Info` has no plural form.

QiJune: How about change `Info` to `Desc`? `Desc` is short for `Description`

skydoorkai: descriptions/attributes/properties/information. Hard to choose a better one.

We can keep to use `EmbeddingTableInfos`, since it can be understood.

PR #1772, opened 2020-02-22
add rwlock for embedding table

QiJune:
If there are multiple workers, the embedding table in PS will be read/write concurrently. Here, we add a rwlock to the embedding table. The lock is for the whole embedding table, which is a coarse grain lock.

In further, we may need a high performance concurrent map. [concurrent-map](https://github.com/orcaman/concurrent-map) is a potential choice.
> `concurrent-map` provides a high-performance solution to this by sharding the map with minimal time spent waiting for locks.

### DIFF
--- a/elasticdl/pkg/common/embedding_table.go
+++ b/elasticdl/pkg/common/embedding_table.go
@@ -2,6 +2,7 @@ package common
 
 import (
 	"github.com/tensorflow/tensorflow/tensorflow/go/core/framework/types_go_proto"
+	"sync"
 )
 
 // EmbeddingTable struct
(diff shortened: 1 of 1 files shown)


skydoorkai:
Do we really need this lock?  How does this lock impact performance?
Some worker may get a wrong lookup_embedding value when multiple workers are simultaneously looking up an embedding which is not yet initialized. This happens not very often, and most deep learning training can tolerate the wrong value in one step.

QiJune:
@skydoorkai Yes, we did. Since Go is very strict. Without this lock, there will be an error `fatal error: concurrent map read and map write`.

At first, new embedding vectors come in, and we have to use write lock. It will do a little harm to the performance. But as the training goes on, the worker will only read embedding vectors. We use RWLock here, it will not influence the performance.

skydoorkai (approved):

PR #1773, opened 2020-02-23
Add the switch to force use the kube config file while submitting the ElasticDL job.

brightcoder01:


elasticdl/python/common/args.py
```diff
@@ -261,6 +261,15 @@ def add_common_params(parser):
         "the job automatically, and users need to launch the job through "
         "command `kubectl create -f path_to_yaml_file`.",
     )
+    add_bool_param(
+        parser=parser,
+        name="--force_use_kube_config_file",
```
skydoorkai, line 266: `use_kube_config_file` should be enough.

brightcoder01: if use_kube_config_file == False，the k8s client will also load the kube config file if the pod is not running in a K8S environment.
IMO, adding `force` may be clear

elasticdl/python/common/k8s_client.py
```diff
@@ -67,14 +68,23 @@ def __init__(
                 Used as pod name prefix and value for "elasticdl" label.
             event_callback: If not None, an event watcher will be created and
                 events passed to the callback.
+            force_use_kube_config_file: If true, force load the cluster config
+                from ~/.kube/config file. Otherwise, if it's in a process
+                running in a K8S environment, it loads the incluster config,
+                if not, load the kube config file.
```
workingloong: "force load the cluster config" -> "force to load the cluster config"

"load the kube config file."  -> "loads the kube config file."

skydoorkai (approved):

elasticdl/python/common/k8s_client.py
brightcoder01: Good suggestion, updated.

PR #1774, opened 2020-02-24
fix worker crash when master stops before worker pods (#1766)

skydoorkai:
* fix worker crash when master stops before worker pods

* add comment


QiJune (approved):
Excellent!

PR #1777, opened 2020-02-24
Add initializer

lwpyr:
add some initializer

### DIFF
--- a/elasticdl/pkg/common/initializer.go
+++ b/elasticdl/pkg/common/initializer.go
@@ -70,9 +70,10 @@ func Constant(n interface{}) Initializer{
 }
 
 // RandomNorm return a normal Initializer
-func RandomNorm(mean float64, std float64) Initializer{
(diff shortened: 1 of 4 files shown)


elasticdl/pkg/common/initializer.go
```diff
@@ -0,0 +1,111 @@
...
+		length := int(DimProduct(t.Dims))
+		for i := 0; i<length; i++ {
+			byteSet(t.Buffer, i, n)
+		}
+		return nil
+	}
+}
+
+// RandomNorm return a normal Initializer
+func RandomNorm(mean float64, std float64) Initializer{
```
QiJune: If there is no seed, all the embedding vectors will be initialized with a same value sequence.
It seems that we should add a `seed` parameter. If the seed is provide, we use it. It the seed is not provided, we use a random seed.

QiJune (approved):

PR #1780, opened 2020-02-25
Catch FileExistsError when make directory for checkpoint version.

workingloong:


elasticdl/python/common/save_utils.py
```diff
@@ -108,7 +108,8 @@ def _get_checkpoint_file(
         checkpoint_version_dir = os.path.join(
             checkpoint_dir, "version-%s" % str(version)
         )
-        os.makedirs(checkpoint_version_dir, exist_ok=True)
+        with contextlib.suppress(FileExistsError):
+            os.makedirs(checkpoint_version_dir, exist_ok=True)
```
QiJune, line 112: Since we already set `exist_ok` to true, I am curious when this bug will be triggered?

QiJune (approved):

PR #1781, opened 2020-02-25
Add some checkpoint util functions in Go PS

QiJune:


elasticdl/pkg/ps/checkpoint_test.go
```diff
@@ -0,0 +1,58 @@
...
+
+import (
+	"elasticdl.org/elasticdl/common"
+	"github.com/stretchr/testify/assert"
+	"os"
+	"path"
+	"testing"
+)
+
+func TestCheckPoint(t *testing.T) {
```
skydoorkai, line 11: Add another test that a model has M bucket, and saved to a directory. Then N bucket (N != M) is used to load the model from the directory.
For example, M = 2, N = 3. In this way, more code paths such as `MergeIndexedSlices` are tested.

QiJune: Done

skydoorkai (approved):

PR #1782, opened 2020-02-25
Use k8s_client to replace the Client in validate_job_status.py.

workingloong:
The Client in validate_job_status.py has code duplication with `k8s_client.Client` and deviation from DRY (Do not repeat yourself) principle.

### DIFF
--- a/scripts/setup_k8s_env.sh
+++ b/scripts/setup_k8s_env.sh
@@ -1,5 +1,6 @@
 #!/bin/bash
 
+sudo pip3 install tensorflow
 # install Kuberneters Python client to validate job status
 sudo pip3 install kubernetes
 export MINIKUBE_WANTUPDATENOTIFICATION=false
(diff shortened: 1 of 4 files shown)


scripts/setup_k8s_env.sh
```diff
@@ -1,5 +1,6 @@
 #!/bin/bash
 
+sudo pip3 install tensorflow
```
terrytangyuan, line 3: Why is this needed for status validation?

workingloong: Because   the module https://github.com/sql-machine-learning/elasticdl/blob/492fe94a377ff15cb67468fefbc99f89a16147a2/elasticdl/python/common/k8s_client.py#L13 in `elasticdl.python.common.k8s_client.Client`  imports Tensorflow and I import `elasticdl.python.common.k8s_client.Client` in `validate_job_status.py`.

workingloong: I find dependency packages for `elasticdl.python.common.k8s_client.Client` is too many. Tensorflow, odps and others. So, it is more simple to use an indenpent `Client` for `validate_job_status.py`. I will close the PR.

PR #1787, opened 2020-02-29
Set initalized True after restoring parameters

workingloong:
The `init_status` has been renamed to `initialized`


QiJune (approved):
Excellent!

PR #1790, opened 2020-03-02
Allow to pass service name for initializing FTLib consensus

terrytangyuan:
Part of https://github.com/sql-machine-learning/elasticdl/issues/1710.

### DIFF
--- a/elasticdl/python/collective_ops/communicator.py
+++ b/elasticdl/python/collective_ops/communicator.py
@@ -1,3 +1,5 @@
+import socket
+
 from elasticdl.python.common.constants import CollectiveCommunicatorStatus
 from elasticdl.python.common.log_utils import default_logger as logger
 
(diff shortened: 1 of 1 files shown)


elasticdl/python/collective_ops/communicator.py
```diff
@@ -16,10 +18,15 @@
 
 
 class CollectiveCommunicator(object):
-    def __init__(self):
+    def __init__(self, svc_name=None):
```
skydoorkai, line 21: What is this `svc_name`? Current worker's service name, or FtLib service?

terrytangyuan: FTLib k8s service. I’ll work on adding that service in a follow-up PR

skydoorkai: `getaddrinfo` of this `svc_name` will return a list of all worker ip info?

terrytangyuan: Yes, this way the consensus method knows what addresses are known.

skydoorkai (approved):

PR #1795, opened 2020-03-04
use multi-stage dockerfile

skydoorkai:
Using one Dockerfile for all.
To build dev:
```
docker build \
    --target dev \
    -t elasticdl:dev \
    -f elasticdl/docker/Dockerfile .
```

To build ci:
```
docker build \
    --target ci \
    -t elasticdl:ci \
    -f elasticdl/docker/Dockerfile .
```

To build ElasticDL image for end-user use, which does not include building/dev tools:
```
docker build \
    -t elasticdl:cpu \
    -f elasticdl/docker/Dockerfile .
```
Better to use DOCKER_BUILDKIT in building.


elasticdl/docker/Dockerfile
```diff
@@ -1,18 +1,108 @@
...
-RUN apt-get update && apt-get install -y unzip curl git software-properties-common libeigen3-dev
-
-COPY elasticdl/requirements.txt /requirements.txt
-ARG EXTRA_PYPI_INDEX=https://pypi.org/simple
-RUN pip install -r /requirements.txt --extra-index-url=${EXTRA_PYPI_INDEX}
+# Copy GO PS
+ARG GO_PS=/root/go/bin/main
+RUN mkdir -p /usr/local/go/bin
+COPY --from=builder  ${GO_PS} /usr/local/go/bin
+ENV PATH /usr/local/go/bin::$PATH
```
QiJune: `ENV PATH /usr/local/go/bin:$PATH`

elasticdl/docker/Dockerfile
```diff
@@ -1,18 +1,108 @@
...
+RUN rm -rf /root/.keras/datasets
+
+# Stage 3: dev
+FROM data as dev
+
+# Remove /elasticdl in dev
+RUN cd / && rm -rf /elasticdl && \
+    rm /elasticdl-develop-py3-none-any.whl
+
+# Stage 4: ci
```
QiJune: It seems that dev and ci are very thin wrapper. Could we just has two stage, dev and release?
- The dev install necessary dependencies, and generate Go binary and python wheel package.
- The release just install the Go binary and python wheel package, and copy model zoo.

skydoorkai: ci is used in ci testing in [client_test.sh]( https://github.com/sql-machine-learning/elasticdl/blob/develop/scripts/client_test.sh). It needs the source code for image building.

So we need at least 3 images:
1. dev that including all building/development dependencies. Also some sample data(maybe move sample data to ci).
2. ci to including both dev and source code and sample model zoo.
3. release for the end-user that does not need those dependencies and sample data. Does it need sample model zoo?

If dev does not include sample data, we can reduce to 3 stages.
1. dev with Go binary and python wheel.
2. release to install binary/wheel from dev.
3. ci, based on dev, add sample model zoo and sample data.

elasticdl/README.md
```diff
@@ -44,12 +47,13 @@ docker run --rm -u $(id -u):$(id -g) -it \
 
 ## Continuous Integration Docker Image
 
-Continuous integration docker image contains everything from the development docker image and the ElasticDL source code. It is  used to run continuous integration with the latest version of the source code. In repo's root directory, run the following command:
+Continuous integration docker image contains everything from the development docker image , processed demo data in RecordIO format and the ElasticDL source code. It is  used to run continuous integration with the latest version of the source code. In repo's root directory, run the following command:
```
terrytangyuan: remove unnecessary space before comma and "It is  used"

terrytangyuan (commented):
Multi-stage build requires Docker 17.05 or higher. Are you sure we want to drop support for older versions of Docker? It would also be good to add a note on this in the doc if we really want to do this.

elasticdl/README.md
skydoorkai: Fixed.

skydoorkai:
> Multi-stage build requires Docker 17.05 or higher. Are you sure we want to drop support for older versions of Docker? It would also be good to add a note on this in the doc if we really want to do this.


Add a note in the doc.

terrytangyuan (approved):

PR #1806, opened 2020-03-09
Use keras callback called by worker to schedule learning rate.

### LINKED ISSUE #1797: support learning rate scheduler in worker instead of PS
Before, the learning rate scheduler is written in Python, and used in Python PS.

We have Go PS now, and it's not convenient to call python module in Go directly. We are trying to support learning rate scheduler in worker, and report the learning rate to PS.

workingloong:
Fix #1797 
On the train begin of each batch, the worker schedules the learning rate and sends it to PS with gradients.


elasticdl/python/elasticdl/callbacks.py
```diff
@@ -90,3 +92,45 @@ def on_task_end(self, task, logs=None):
             self._completed_steps += task_batch_count
             if self._completed_steps > self._max_steps:
                 self.model.stop_training = True
+
+
+class LearningRateScheduler(tf.keras.callbacks.Callback):
+    """Stop training if the training steps exceed the maximum.
```
skydoorkai: Wrong comment for the function.

elasticdl/python/ps/servicer.py
```diff
@@ -249,3 +244,12 @@ def _save_params_to_checkpoint_if_needed(self):
                 shard_index=self._ps_id,
                 shard_num=self._num_ps_pods,
             )
+
+    def _set_optimizer_learning_rate(self, learning_rate):
+        if learning_rate is None:
```
skydoorkai: What is the default value for a float in proto? 0?

elasticdl/python/ps/servicer.py
```diff
@@ -249,3 +244,12 @@ def _save_params_to_checkpoint_if_needed(self):
...
+            return
+
+        if self._use_wrap_opt:
+            self._optimizer.set_learning_rate(learning_rate)
```
QiJune, line 253:
```python
def set_learning_rate(self, learning_rate):
        K.set_value(self._opt.lr, K.get_value(learning_rate))
```

It seems that there is no need to add a `if/else` here.

workingloong: The ` self._optimizer` maybe an `OptimizerWrapper` instance or `Optimizer` instance, so it needs `if/else`.

elasticdl/python/ps/servicer.py
workingloong: For numeric types, the default value is zero and 0.0 for a float. https://developers.google.com/protocol-buffers/docs/proto3#default
And I will fix it using `if learning_rate == 0.0:`

elasticdl/python/worker/worker.py
```diff
@@ -576,6 +577,7 @@ def report_gradient_to_ps(self, grads):
         for ps_id in range(self._ps_num):
             req = reqs[ps_id]
             req.version = self._model_versions_from_ps[ps_id]
+            req.learning_rate = K.get_value(self._model.optimizer.lr)
```
QiJune, line 580: One question, do we need to propose a concept: `chief worker`? The learning rate is only pushed by the chief worker to PS.

I am not very sure, we could merge this PR first.

QiJune (approved):
Excellent!

elasticdl/python/worker/worker.py
workingloong: The model version which the chief worker gets maybe not continuous. So, it cannot schedule the learning rate precisely. The learning rate may overlap if the workers get the same model version. Here, we may need a trade-off.

PR #1807, opened 2020-03-09
add Momentum optimizer for Go PS

QiJune:


elasticdl/pkg/ps/optimizer.go
```diff
@@ -86,6 +86,57 @@ func (opt *SGDOptimizer) InitOptimizer(pb *proto.Model) error {
...
+	}
+	opt.IndexedKernel = func(grad *common.IndexedSlices, param *common.Tensor, name string) error {
+		v := opt.v.GetDenseParameter(name)
+		return kernel.IndexedMomentum(grad, param, v, opt.mu, opt.nesterov, opt.GetLR())
+	}
+	return &opt
+}
+
+// InitOptimizer set v non-embedding of MomentumOptimizer
+func (opt *MomentumOptimizer) InitOptimizer(pb *proto.Model) error {
```
terrytangyuan: So this never returns error?

QiJune: Yes, I will clean the unnecessary return value of `error`

skydoorkai (approved):

PR #1815, opened 2020-03-10
Set timeout for channel_ready_future and retry to connect with 3 times.

### LINKED ISSUE #1814: Worker hangs when connects to PS.
The worker will wait the grpc channel of PS is ready using `channel_ready_future`. We should set timeout and retry.
image

workingloong:
Fix #1814
If the PS instance is launched after the worker, the result of the `grpc.channel_ready_future` may always wait.

### DIFF
--- a/elasticdl/python/worker/main.py
+++ b/elasticdl/python/worker/main.py
@@ -1,4 +1,5 @@
 import grpc
+import concurrent
 
 from elasticdl.python.common import log_utils
 from elasticdl.python.common.args import parse_worker_args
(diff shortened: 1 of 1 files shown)


elasticdl/python/worker/main.py
```diff
@@ -24,12 +25,27 @@ def main():
             channel = build_channel(addr)
 
             # Wait the channel is ready by a Future object.
-            grpc.channel_ready_future(channel).result()
-            logger.info(
-                "grpc channel %s to connect pod %s is ready"
-                % (addr, addr.split(".")[0])
-            )
-            ps_channels.append(channel)
+            connect_succeed = False
```
terrytangyuan: Succeeded

elasticdl/python/worker/main.py
```diff
@@ -24,12 +25,27 @@ def main():
...
+            for i in range(3):
```
terrytangyuan: Move this 3 and timeout=60 to constants in this file?

elasticdl/python/worker/main.py
```diff
@@ -24,12 +25,27 @@ def main():
...
+                try:
```
terrytangyuan, line 31: Also would be good explaining why this is needed. Why not simply increasing the timeout instead of retrying with timeout=60?

elasticdl/python/worker/main.py
workingloong: Yeah

elasticdl/python/worker/main.py
workingloong: The worker may start before PS and the future will be waiting even if the PS starts. So, the worker can retry to connect with the PS after it starts.

skydoorkai: `grpc.channel_ready_future` is waiting until the GRPC server channel is ready. It is ok for PS to launch after worker pods.
Why need to add timeout and retry? Is this for the case that when PS pods may fail to run,  worker pods can exit?

workingloong: In many experiments, I find some workers hang and always wait GRPC channel is ready. After I delete the worker, the new worker succeeds to connect with PS.

skydoorkai (approved):

PR #1818, opened 2020-03-11
Use a lock for `get` and `set` of EmbeddingTable

### LINKED ISSUE #1808: Exception calling application: dictionary changed size during iteration
The exception occasionally happens when the worker reports gradients to ps.
image

workingloong:
Fix #1808 
The parameter dictionary may be updated by other threads during iteration to save it into a checkpoint file.

### DIFF
--- a/elasticdl/python/ps/embedding_table.py
+++ b/elasticdl/python/ps/embedding_table.py
@@ -52,21 +52,20 @@ class EmbeddingTable(object):
         if len(indices) == 0:
             return None
         values = []
-        with self._lock:
-            for i in indices:
-                value = self.embedding_vectors.get(i, None)
-                if value is None:
+        for i in indices:
+            value = self.embedding_vectors.get(i, None)
+            if value is None:
+                with self._lock:
                     value = self.initializer(shape=(self.dim,)).numpy()
                     self.embedding_vectors[i] = value
-                values.append(value)
+            values.append(value)
         return np.stack(values)
 
     def set(self, indices, values):
         # TODO(qijun) need to add a RWLock in Sync-SGD
-        with self._lock:
-            for index, i in enumerate(indices):
-                embedding_vector = values[index]
-                self.embedding_vectors[i] = embedding_vector
+        for index, i in enumerate(indices):
+            embedding_vector = values[index]
+            self.embedding_vectors[i] = embedding_vector
 
     def clear(self):
         self.embedding_vectors.clear()
@@ -93,13 +92,12 @@ class EmbeddingTable(object):
 
     def get_table_size(self):
         """Get the element count of an embedding table"""
-        with self._lock:
-            if len(self.embedding_vectors) > 0:
-                element_size = list(self.embedding_vectors.values())[
-                    0
-                ].itemsize
-                size = self.dim * len(self.embedding_vectors) * element_size
-                return size
+        if len(self.embedding_vectors) > 0:
+            element_size = list(self.embedding_vectors.values())[
+                0
+            ].itemsize
+            size = self.dim * len(self.embedding_vectors) * element_size
+            return size
         return 0
 
     def debug_info(self):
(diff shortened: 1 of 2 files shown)


elasticdl/python/ps/parameters.py
```diff
@@ -175,10 +175,13 @@ def to_model_pb(self):
...
         model_pb.version = self.version
-        for name, var in self.non_embedding_params.items():
+        non_embedding_params = self.non_embedding_params.copy()
+        for name, var in non_embedding_params.items():
             serialize_ndarray(var.numpy(), model_pb.dense_parameters[name])
+        del non_embedding_params
 
-        for name, embedding_table in self.embedding_params.items():
+        embedding_params = self.embedding_params.copy()
+        for name, embedding_table in embedding_params.items():
```
skydoorkai: Maybe this embedding_param iteration is not the cause of the bug, as `self.embedding_params` has a fixed number (number of embedding tables) after initialization.

It is probably in `embedding_table.to_indexed_slices(),` a few lines later, as when iterating through the embedding vectos, new vectors may be added by other threads.

workingloong: Yes. It is more suitable to copy `embedding_table.embedding_vectors` before iteration or use a lock ?

skydoorkai: embedding_table.embedding_vectors may be too long for copy. If lock does not affect training efficiency when checkpoint is not saved, better to use a lock.

workingloong: In PS with Python, the `optimizer_wrapper` has used a lock to update the embedding table. https://github.com/sql-machine-learning/elasticdl/blob/e695ba7b33f01383798b28825fab408dae4e1327/elasticdl/python/ps/optimizer_wrapper.py#L155-L157

So, it does not affect training efficiency to use a lock for `set` and `get` in `EmbeddingTable`. I will use a lock in the next commit.

elasticdl/python/ps/embedding_table.py
```diff
@@ -44,34 +46,38 @@ def __init__(self, name, dim=None, initializer=None, is_slot=False):
...
-        for i in indices:
-            value = self.embedding_vectors.get(i, None)
-            if value is None:
-                value = self.initializer(shape=(self.dim,)).numpy()
-                self.embedding_vectors[i] = value
-            values.append(value)
+        with self._lock:
+            for i in indices:
+                value = self.embedding_vectors.get(i, None)
+                if value is None:
```
skydoorkai: Only adding embedding vector and iteration need the lock. Reading or updating existing vectors does not.
So Move the lock here after `if value is None:` for adding. 

Also, `def set(self, indices, values)` below  does not need the lock, as it does not add any vectors.

QiJune: We need a RWLock here. Please refer to the implementation in Go PS. https://github.com/sql-machine-learning/elasticdl/blob/develop/elasticdl/pkg/common/embedding_table.go#L14

The lock is used at `GetEmbeddingVector ` func and `ToIndexedSlices ` func.

workingloong: If we don't use lock the `def set(self, indices, values)`, Reading and writing the `embedding_vector` are simultaneous which may cause inconsistency.

QiJune: @workingloong  Yes. I am not sure if the reading/writeing concurrently will hurt the model accuracy. It seems that it's a tradeoff here, training speed and model accuracy. We could do some expriments.

skydoorkai: In async-SGD, reading/writing inconsistency does not matter, as the async-SGD itself has a staleness issue.
In sync-SGD, there won't be any reading/writing inconsistency.

workingloong: OK, we may need to do some experiments to view training speed and model accuracy and then to add the lock. Now, I only use the lock for iteration and set new vectors.

skydoorkai (approved):

PR #1819, opened 2020-03-11
support learning rate scheduler in Go PS

QiJune:
This PR add learning rate scheduler in Go PS. And this PR also add a new message `PushGradientsRequest`, since its not very reasonable to put a learning rate field in a `Model`.

### DIFF
--- a/elasticdl/pkg/ps/server.go
+++ b/elasticdl/pkg/ps/server.go
@@ -162,12 +162,12 @@ func (s *Server) PullEmbeddingVectors(ctx context.Context, in *proto.PullEmbeddi
 func (s *Server) PushGradients(ctx context.Context, in *proto.PushGradientsRequest) (*proto.PushGradientsResponse, error) {
 	// TODO: only support async now
 	var lrMultiplier = float32(1.0)
-	if s.lrStalenessModulation && s.Model.Version > in.Version {
-		staleness := s.Model.Version - in.Version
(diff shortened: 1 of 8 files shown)


elasticdl/pkg/ps/server.go
```diff
@@ -159,14 +159,22 @@ func (s *Server) PullEmbeddingVectors(ctx context.Context, in *proto.PullEmbeddi
...
 	// TODO: only support async now
 	var lrMultiplier = float32(1.0)
-	if s.lrStalenessModulation && s.Model.Version > in.Version {
-		staleness := s.Model.Version - in.Version
+	if s.lrStalenessModulation && s.Model.Version > in.Gradients.Version {
+		staleness := s.Model.Version - in.Gradients.Version
 		lrMultiplier = lrMultiplier / float32(staleness)
 	}
-	err := s.Opt.ApplyGradients(in, s.Model, lrMultiplier)
+	s.Opt.SetLR(in.LearningRate)
```
skydoorkai: When multiple workers are PushGradients to PS, `s.Opt.lr` may be overwritten among threads.
Probably add a lr argument in `s.Opt.ApplyGradients`. If set (>0), use it. Otherwise, use the default lr value in s.Opt.

QiJune: Great idea. Fix it.

skydoorkai (approved):

PR #1821, opened 2020-03-11
Support lr_staleness_modulation in python PS

skydoorkai:
This PR only works with learning_rate_scheduler so workers can send learning_rate together with gradients.

Modulating learning_rate defined in the optimizer is not implemented in this PR.


QiJune (approved):
Excellent!

PR #1822, opened 2020-03-12
support adagrad optimizer in Go PS

QiJune:


elasticdl/pkg/ps/optimizer.go
```diff
@@ -235,8 +281,9 @@ const (
 )
 
 var optArgumentsMap = map[string][]string{
-	"SGD":  []string{optArgLR, optArgMomentum, optArgNesterov},
-	"Adam": []string{optArgLR, optArgBeta1, optArgBeta2, optArgEpsilon, optArgAmsgrad},
+	"SGD":     []string{optArgLR, optArgMomentum, optArgNesterov},
```
skydoorkai, line 284: MomentumOptimizer missing?

QiJune: The paraments MomentumOptimizer need are incuded in SGD already.  `optArgLR, optArgMomentum, optArgNesterov`

skydoorkai (approved):

PR #1825, opened 2020-03-12
Support VolumeMount with subPath

brightcoder01:
Use the following configuration
```
--volume='claim_name=glusterfs-elasticdl,mount_path=/assets,sub_path=elasticdl/20200312/vocab;claim_name=glusterfs-elasticdl,mount_path=/work_dir' \
```

We will mount the directory `elasticdl/20200312/vocab` from glusterfs into the POD as the local directory '/assets'. In this way, we can pass the vocabulary from different data without updating the Python code of the model. The model inside the POD only see the local path '/assets'

What's more, in the example above, there are two mounts using the PVC glusterfs-elasticdl, we only create just one volume instead of two(in the previous implementation) in the generated YAML.


elasticdl/python/common/k8s_volume.py
```diff
@@ -16,32 +22,54 @@ def parse_volume_and_mount(volume_conf, pod_name):
         volumes (List): a Python list contains k8s volumes.
         volume_mounts (List): a Python list contains k8s volume mounts.
     """
-    volumes = []
+    volumes_cache = {}
```
terrytangyuan: Why is this called cache?

elasticdl/python/common/k8s_volume.py
```diff
@@ -16,32 +22,54 @@ def parse_volume_and_mount(volume_conf, pod_name):
...
+    num_volumes = 0
+    for volume_dict in volume_dicts:
         if "claim_name" in volume_dict:
-            pvc_volume_source = client.V1PersistentVolumeClaimVolumeSource(
-                claim_name=volume_dict["claim_name"], read_only=False
-            )
-            volume = client.V1Volume(
-                name=volume_name, persistent_volume_claim=pvc_volume_source
-            )
+            volume = volumes_cache["pvc"].get(volume_dict["claim_name"], None)
```
workingloong: An example may be needed to explain when the volume isn't None?

brightcoder01: We can check the example in the PR description: 
```
--volume='claim_name=glusterfs-elasticdl,mount_path=/assets,sub_path=elasticdl/20200312/vocab;claim_name=glusterfs-elasticdl,mount_path=/work_dir' \
```

brightcoder01: Will add this example into the comments.

elasticdl/python/common/k8s_volume.py
brightcoder01: It's used to store the built volume while parsing the volume configuration string. We use it to avoid creating the volume with the same name and type.
Maybe renaming it volumes_per_type will be more suitable.

terrytangyuan (approved):

PR #1828, opened 2020-03-12
Add barrier implementation and update logic on broadcast src rank

terrytangyuan:
* Added barrier implementation that was missing earlier
* Update the logic on broadcast src rank since this is handled by FTLib and we only need to broadcast from root worker for now.

### DIFF
--- a/elasticdl/python/worker/worker.py
+++ b/elasticdl/python/worker/worker.py
@@ -748,32 +748,23 @@ class Worker(object):
     def _get_local_model_params(self):
         return self._non_embed_vars
 
-    # TODO: Implement master gRPC service to select a worker
-    # to be used for broadcast model parameters from.
     @staticmethod
-    def _get_ip_of_broadcast_root_worker():
-        return 1
+    def _get_rank_of_broadcast_src_worker():
+        return 0
 
-    @staticmethod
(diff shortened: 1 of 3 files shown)


elasticdl/python/worker/worker.py
```diff
@@ -748,40 +747,31 @@ def _update_local_model_params(self, model_params):
...
     def _broadcast_model_params(self):
         status = self._collective_communicator.barrier()
         if status == CollectiveCommunicatorStatus.FAILED:
             self.logger.warning("Failed to perform barrier operation")
             return False
-        broadcast_root_worker_ip = self._get_ip_of_broadcast_root_worker()
-        this_worker_ip = self._get_ip_of_this_worker()
-        is_broadcast_root_worker = this_worker_ip == broadcast_root_worker_ip
+        broadcast_root_worker_ip = self._get_rank_of_broadcast_src_worker()
+        this_worker_ip = self._get_rank_of_this_worker()
```
skydoorkai: broadcast_root_worker_ip, broadcast_root_worker_ip --> rename to *_rank

terrytangyuan: Thanks. Fixed.

skydoorkai (approved):

PR #1832, opened 2020-03-13
Rename concat_column to concatenated_categorical_column.

brightcoder01:
Add more description and unit tests.

### DIFF
--- a/elasticdl/python/elasticdl/feature_column/feature_column.py
+++ b/elasticdl/python/elasticdl/feature_column/feature_column.py
@@ -228,6 +228,54 @@ def concatenated_categorical_column(categorical_columns):
     to avoid this conflict. The id range of the concatenated column is
     [0, num_bucket_0 + num_bucket_1 + ... + num_bucket_n].
 
+    Example
+
+    ```python
+    id = categorical_column_with_identity("id", num_buckets=32)
+    work_class = categorical_column_with_vocabulary_list(
+        "work_class",
+        vocabulary_list=[
+            "Private",
+            "Self-emp-not-inc",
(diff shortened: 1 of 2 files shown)


elasticdl/python/elasticdl/feature_column/feature_column.py
```diff
@@ -210,7 +210,80 @@ def embedding_weight_name(self):
...
+
+    Use this when you have many categorical columns and want to map all
+    to embedding. It's recommended to concatenate the sparse id tensors
+    from the categorical columns above to one sparse id tensor using this
+    API, and then map the combined SparseTensor to Embedding. Because
+    there is much overhead to create embedding variables for each
+    categorical column, the model size will be huge. Using this way, we
+    can reduce the model size.
+
+    The id range of source categorical columns are [0, num_buckets_0],
```
workingloong: "The id range of " -> "The output integer id range of"

workingloong: Why is it `[0, num_buckets_0]` not `[0, num_buckets_0)`

brightcoder01: > Why is it `[0, num_buckets_0]` not `[0, num_buckets_0)`

Good catch, will update it.

terrytangyuan (approved):
LGTM once comments are addressed

elasticdl/python/elasticdl/feature_column/feature_column.py
brightcoder01: > "The id range of " -> "The output integer id range of"

Updated to `The output id range of`

PR #1842, opened 2020-03-17
Add RoundIdentity layer to transform numeric values to integer ids

### LINKED ISSUE #1841: Implement a keras layer to round numeric values and truncate values with a max value.
For numeric features, we may want to convert the values to integer id by rounding and then make embedding with those ids.  Though we can implement the function by using bucketing with boundaries where 1 interval，it is tedious to set the boundaries and inefficient to transform if the max value is big.

Example：
inputs: [[1.2], [4.1], [6.7], [100.1]] 
max_value: 50
default_value: 50
the outputs: [[1], [4], [7], [50]]

workingloong:
Fix #1841


elasticdl_preprocessing/layers/round_identity.py
```diff
@@ -0,0 +1,71 @@
+import tensorflow as tf
+from tensorflow.python.ops.ragged import ragged_functional_ops, ragged_tensor
+
+
+class RoundIdentity(tf.keras.layers.Layer):
+    """Implements numeric feature rounding with a max value.
```
brightcoder01: I think this layer is casting a numeric feature into a categorical feature.

elasticdl_preprocessing/layers/round_identity.py
```diff
@@ -0,0 +1,71 @@
...
+        inp = np.asarray([[1.2], [1.6], [0.2], [3.1], [4.9]])
+        layer(inputs)
+        [[1], [2], [0], [3], [5]]
+    ```
+
+    Arguments:
+        num_buckets: Range of inputs and outputs is `[0, num_buckets)`.
+        **kwargs: Keyword arguments to construct a layer.
+
+    Input shape: A numeric tensor of shape
```
brightcoder01: Tensor and SparseTensor

elasticdl_preprocessing/layers/round_identity.py
workingloong: How do you define the categorical feature? I think the categorical feature is the type fo source feature. The layer is a tool to process the numeric feature although the outputs are integer ids.

elasticdl_preprocessing/layers/round_identity.py
workingloong: Done

elasticdl_preprocessing/layers/round_identity.py
brightcoder01: Maybe: cast a numeric feature into a d discretized/categorical value.
IMO, categorical value is discretized and enumerable. After executing this layer, the value become enumerable: [0, 1, 2, 3, ..., MAX_VALUE]

workingloong: Cast a numeric feature into a discrete integer value?

brightcoder01: discretized integer value?

brightcoder01 (approved):

elasticdl_preprocessing/layers/round_identity.py
workingloong: "discrete" is an adj and the code spell checker in VScode hints that "discretized" is an unknown word.

PR #1845, opened 2020-03-17
Add the ToNumber preprocess layer.

brightcoder01:


terrytangyuan (commented):
Is it really necessary to add this to elasticdl? The preprocessing layers seem so simple and I think they should be in examples instead.

brightcoder01:
> Is it really necessary to add this to elasticdl? The preprocessing layers seem so simple and I think they should be in examples instead.

Good question.
The layers is useful for the code_gen of SQLFlow transform expression and can be imported directly in the generated code. If not, I'm afraid we should generate this piece of code multiple times.

terrytangyuan:
@brightcoder01 Sure but maybe this can be moved to examples instead of a module/folder?

elasticdl_preprocessing/tests/to_number_test.py
```diff
@@ -0,0 +1,78 @@
+import unittest
+
+import numpy as np
+import tensorflow as tf
+
+from elasticdl_preprocessing.layers.to_number import ToNumber
+
+
+def sparse_tensor_equal(sp_a, sp_b):
```
workingloong: We may use the method in multiple unit tests of preprocessing layers. Can we move it to a single file like `tests/test_utils.py` ?

brightcoder01: Good idea.

brightcoder01:
> @brightcoder01 Sure but maybe this can be moved to examples instead of a module/folder?

Good suggestion. We will add a model example for all the preprocessing layers after submitting all the layers.

terrytangyuan (approved):

PR #1848, opened 2020-03-17
refine odps reader

QiJune:
User could insert customized data processig logic with API, `parallel_read_records`.

A custom data reader example is as following:

```python
class CustomParallelODPSDataReader(ODPSDataReader):
    def __init__(self, **kwargs):
        ODPSDataReader.__init__(self, **kwargs)

    def read_records(self, task):
        shard_size = (task.end - task.start) // 4
        record_gen = self.parallel_record_records(
            task=task,
            num_processes=4,
            shard_size=shard_size,
            transform_fn=py_parse_data,
        )
        for record in record_gen:
            yield record

    @property
    def records_output_types(self):
        return tf.string
```

In an internal experimet, it will decrease the training time from 4h52min to 3h15min.


elasticdl/requirements.txt
```diff
@@ -2,4 +2,5 @@ grpcio-tools
 kubernetes==10.1.0
 docker
 pyrecordio>=0.0.6
+Cython
```
QiJune, line 5: PyODPS will compile some C code if Cython is installed.
Please refer to http://pyodps.alibaba.net/pyodps-docs/zh-CN/latest/base-tables.html#tunnel

elasticdl/python/data/odps_io.py
```diff
@@ -102,146 +103,106 @@ def __init__(
...
+        return len(self._shards)
+
+    def get_records(self):
+        data = self._result_queue.get()
+        self._put_index()
+        return data
+
+    def stop(self):
+        for q in self._index_queues:
+            q.put((None, None))
```
QiJune, line 159: I find that there is no need for the main process to wait worker processes finishing. The `join` method is time-consuming.

After the main process is done, the daemon child processes will exit automatically.

elasticdl/python/data/odps_io.py
```diff
@@ -102,146 +103,106 @@ def __init__(
...
+        self._result_queue = Queue()
+        self._index_queues = []
+        self._workers = []
+
+        self._shards = []
+        self._shard_idx = 0
+        self._worker_idx = 0
+
+        for i in range(self._num_processes):
+            index_queue = Queue()
```
chunyang-wen, line 137: It is better we give a predefined size queue size to avoid OOM.

elasticdl/python/data/odps_io.py
chunyang-wen: We can use a new class  instead of `None`. For exmaple:

```
class ExitSignal(object): pass
```

elasticdl/python/data/odps_io.py
```diff
@@ -102,146 +103,106 @@ def __init__(
...
+
+        self._create_shards(shards, shard_size)
+        for i in range(2 * self._num_processes):
+            self._put_index()
+
+    def get_shards_count(self):
+        return len(self._shards)
+
+    def get_records(self):
+        data = self._result_queue.get()
```
chunyang-wen, line 153: I am not sure which way is better. `get` or `put` is a blocking call. So the worker has no chance to react to the environment. Previously we add the `timeout` parameter and when an exception is thrown, we test whether we should exit without further reading any data. And then retry.

workingloong (approved):

elasticdl/python/data/odps_io.py
QiJune: This is the queue for index of a task, which stores several (start, end) pairs. So it occupies very little memory.

elasticdl/python/data/odps_io.py
QiJune: There are two steps when reading data from ODPS:
1. call  `record_generator_with_retry` to get data and put it to result queue, we have `timeout` in `record_generator_with_retry` function already.
2. get data from the result queue

The result queue is for sharing local data between multi process. It's okay to use `get`/`put` API.

PR #1853, opened 2020-03-18
Recover the task to call training end callback into todo list

workingloong:
Now, the master cannot recover the task to call train_end_callback if the task fails.

### DIFF
--- a/elasticdl/python/master/task_dispatcher.py
+++ b/elasticdl/python/master/task_dispatcher.py
@@ -301,7 +301,8 @@ class _TaskDispatcher(object):
                 # TODO: keep count of retries.
                 logger.warning("Task %d of %s failed " % (task_id, task.type))
                 if task.type in [
-                    elasticdl_pb2.TRAINING, elasticdl_pb2.TRAIN_END_CALLBACK
+                    elasticdl_pb2.TRAINING,
+                    elasticdl_pb2.TRAIN_END_CALLBACK,
                 ]:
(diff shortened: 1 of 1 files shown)


skydoorkai (approved):

elasticdl/python/master/task_dispatcher.py
```diff
@@ -318,7 +322,7 @@ def report(self, request, success):
             if evaluation_task_completed:
                 self._evaluation_service.complete_task()
 
-            if self._callbacks_list.model.stop_training:
+            if success and self._callbacks_list.model.stop_training:
```
brightcoder01, line 325: Why do we check success here?
If the task always fail, will it enter an infinite loop?

workingloong: If we don't check success, the master will clear the todo and cannot retry the tasks in todo list. To avoid an infinite loop, we have added a TODO here.
https://github.com/sql-machine-learning/elasticdl/blob/a7be465c73e27872db0dae70c1ffa5296970cee6/elasticdl/python/master/task_dispatcher.py#L301-L305

We may need to implement the TODO now.

terrytangyuan (approved):

elasticdl/python/master/task_dispatcher.py
workingloong: I have implemented the TO in #1854

PR #1854, opened 2020-03-18
Set max retries for a task.

workingloong:
Set max retries for a task to avoid an infinite loop.

### DIFF
--- a/elasticdl/python/master/task_dispatcher.py
+++ b/elasticdl/python/master/task_dispatcher.py
@@ -11,6 +11,8 @@ from elasticdl.proto import elasticdl_pb2
 from elasticdl.python.common.constants import TaskExecCounterKey
 from elasticdl.python.common.log_utils import default_logger as logger
 
+MAX_TASK_RETRIES = 3
+
 
 class _Task(object):
     """Internal representation of a task"""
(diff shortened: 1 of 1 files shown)


elasticdl/python/master/task_dispatcher.py
```diff
@@ -328,6 +331,17 @@ def report(self, request, success):
 
         return (time.time() - start_time), task, worker_id
 
+    def check_exceed_max_task_retries(self, task):
+        self._task_retry_count.setdefault(task, 1)
```
brightcoder01, line 338: Remove the key 'task' if it's already success. Otherwise the memory will keep increasing if the task number is huge.

elasticdl/python/master/task_dispatcher.py
```diff
@@ -11,6 +11,8 @@
 from elasticdl.python.common.constants import TaskExecCounterKey
 from elasticdl.python.common.log_utils import default_logger as logger
 
+MAX_TASK_RETRIES = 3
```
terrytangyuan: Start variable name with underscore if it's only used internally

elasticdl/python/master/task_dispatcher.py
workingloong: Done!

elasticdl/python/master/task_dispatcher.py
workingloong: Yeah! Done

skydoorkai (approved):

PR #1862, opened 2020-03-20
Layer to convert tensor to ragged tensor dropping the ignored value

### LINKED ISSUE #1861: Support convert Tensor to RaggedTensor dropping the ignored value.
The format of the sequence is usually a string with separator such as the sentence in a text with whitespace separator. And the number of words in each sentence is not the same. We can split the sentences to a `tf.RaggedTensor`. Then we may want to drop some values like the empty string.

For example:
```python
data = np.array([["Hello world"], [""], ["Welcome to ElasticDL"]])
```
1. Split strings to 
```python
data = tf.RaggedTensor([["Hello", "world"], [""], ["Welcome", "to", "ElasticDL"]])
```
2. Drop the empty string to
```python
data = tf.RaggedTensor([["Hello", "world"], [], ["Welcome", "to", "ElasticDL"]])
```
3. Transform strings in the ragged tensor to integer ids by `Lookup` or `Hashing`
```python
data = tf.RaggedTensor([[0, 1], [], [2, 3, 4]])
```
4. Convert it to `Tensor` with padding by `tf.RaggedTensor.to_tensor(shape=(None, 3))`
```python
data = data.to_tensor(shape=(None,3))
tf.Tensor([[0, 1, 0], [0, 0 ,0], [2, 3, 4]])
```

workingloong:
Fix #1861

### DIFF
--- a/elasticdl_preprocessing/layers/to_ragged.py
+++ b/elasticdl_preprocessing/layers/to_ragged.py
@@ -13,16 +13,16 @@ class ToRagged(tf.keras.layers.Layer):
     Example (Integer):
     ```python
         layer = ToRagged()
-        inp = tf.constant([[1], [-1], [4]], tf.int64)
-        out = layer(inp)
+        input_tensor = tf.constant([[1], [-1], [4]], tf.int64)
+        out = layer(input_tensor)
         [[1], [], [4]]
     ```
 
     Example (String):
     ```python
-        ayer = ToRagged()
-        inp = tf.constant([["1,2,3"], ["4,5"], [""]])
-        out = layer(inp)
+        layer = ToRagged()
+        input_tensor = tf.constant([["1", "2", "3"], ["4", "5"], [""]])
+        out = layer(input_tensor)
     ```
     The expected output is `[["1", "2", "3"], ["4", "5"], []]`
 
@@ -52,9 +52,6 @@ class ToRagged(tf.keras.layers.Layer):
 
         if isinstance(inputs, tf.Tensor):
             inputs = tf.RaggedTensor.from_tensor(inputs)
-
-        if inputs.dtype == tf.string:
-            inputs = tf.strings.split(inputs, sep=self.sep).values
 
         ignore_value = self._get_ignore_value(inputs.dtype)
         if ignore_value is None:
--- a/elasticdl_preprocessing/layers/to_ragged.py
+++ b/elasticdl_preprocessing/layers/to_ragged.py
@@ -42,6 +42,15 @@ class ToRagged(tf.keras.layers.Layer):
         self.ignore_value = ignore_value
 
     def call(self, inputs):
+        if (
+            not isinstance(inputs, tf.Tensor) or
+            not isinstance(inputs, tf.RaggedTensor)
+        ):
+            raise TypeError(
+                "The inputs must be a Tensor or RaggedTensor and "
+                "the type of inputs is {}".format(type(inputs))
+            )
+
         if isinstance(inputs, tf.Tensor):
             inputs = tf.RaggedTensor.from_tensor(inputs)
 
@@ -51,8 +60,11 @@ class ToRagged(tf.keras.layers.Layer):
                 inputs = tf.strings.split(inputs, sep=self.sep).values
             elif inputs.dtype.is_integer:
                 self.ignore_value = -1
-        self.ignore_value = tf.cast(self.ignore_value, inputs.dtype)
 
-        return tf.ragged.boolean_mask(
-            inputs, tf.not_equal(inputs, self.ignore_value)
-        )
+        if self.ignore_value is not None:
+            self.ignore_value = tf.cast(self.ignore_value, inputs.dtype)
+            return tf.ragged.boolean_mask(
+                inputs, tf.not_equal(inputs, self.ignore_value)
+            )
+
+        return inputs
--- a/elasticdl_preprocessing/layers/to_ragged.py
+++ b/elasticdl_preprocessing/layers/to_ragged.py
@@ -44,7 +44,7 @@ class ToRagged(tf.keras.layers.Layer):
         self.ignore_value = ignore_value
 
     def call(self, inputs):
-        if not isinstance(inputs, tf.Tensor) or not isinstance(
+        if not isinstance(inputs, tf.Tensor) and not isinstance(
             inputs, tf.RaggedTensor
         ):
             raise TypeError(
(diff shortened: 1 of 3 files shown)


elasticdl_preprocessing/layers/to_ragged.py
```diff
@@ -0,0 +1,58 @@
...
+        inp = tf.constant([["1,2,3"], ["4,5"], [""]])
+        out = layer(inp)
+        [["1", "2", "3"], ["4", "5"], []]
+    ```
+
+    Arguments:
+        sep: Valid if the input's dtype is string.
+        ignore_value: Entries in inputs equal to this value will be
+            absent from the output `RaggedTensor`. If `None`, default value of
+            input's dtype will be used ('' for `str`, -1 for `int`).
```
terrytangyuan, line 33: I had similar comment on this in the other PR

elasticdl_preprocessing/layers/to_ragged.py
```diff
@@ -0,0 +1,58 @@
...
+
+    Output shape: An `RaggedTensor` with the same shape as inputs
+    """
+
+    def __init__(self, sep=_COMMA_SEP, ignore_value=None):
+        super(ToRagged, self).__init__()
+        self.sep = sep
+        self.ignore_value = ignore_value
+
+    def call(self, inputs):
```
brightcoder01, line 46: What will happen if inputs is a SparseTensor?

elasticdl_preprocessing/layers/to_ragged.py
```diff
@@ -0,0 +1,58 @@
...
+        if isinstance(inputs, tf.Tensor):
+            inputs = tf.RaggedTensor.from_tensor(inputs)
+
+        if self.ignore_value is None:
```
brightcoder01: Same comments in the ToSparse layer

elasticdl_preprocessing/layers/to_ragged.py
```diff
@@ -0,0 +1,58 @@
+import tensorflow as tf
+
+_COMMA_SEP = ","
+
+
+class ToRagged(tf.keras.layers.Layer):
+    """Converts a `Tensor` or `RaggedTensor` to a `RaggedTensor`,
+    dropping ignore_value cells. If the input's dtype is string, split
+    the string elements to convert the input to `RaggedTensor` firstly.
```
brightcoder01, line 9: What's the TF version requirement for this layer?

elasticdl_preprocessing/layers/to_ragged.py
```diff
@@ -0,0 +1,58 @@
...
+        [[1], [], [4]]
+    ```
+
+    Example (String):
+    ```python
+        ayer = ToRagged()
+        inp = tf.constant([["1,2,3"], ["4,5"], [""]])
+        out = layer(inp)
+    ```
+    The expected output is `[["1", "2", "3"], ["4", "5"], []]`
```
brightcoder01, line 27: For string input, the rank of output will increase 1
For int input, the rank of output won't change.
It will be confusing to the users.

elasticdl_preprocessing/layers/to_ragged.py
workingloong: It may raise an unknown error. We need to check the input type and raise an error is the type is not `Tensor` or `RaggedTensor`.

elasticdl_preprocessing/layers/to_ragged.py
workingloong: So, is the expected output for string input `[[["1", "2", "3"]], [["4", "5"]], [[]]]`?

elasticdl_preprocessing/layers/to_ragged.py
workingloong: Above TF 2.0.0 and I have added a note.

elasticdl_preprocessing/layers/to_ragged.py
```diff
@@ -0,0 +1,72 @@
...
+    """
+
+    def __init__(self, sep=_COMMA_SEP, ignore_value=None):
+        super(ToRagged, self).__init__()
+        self.sep = sep
+        self.ignore_value = ignore_value
+
+    def call(self, inputs):
+        if not isinstance(inputs, tf.Tensor) and not isinstance(
+            inputs, tf.RaggedTensor
```
chunyang-wen:
```python
if not isinstance(inputs, (tf.Tensor, tf.RaggedTensor):
    pass
```

elasticdl_preprocessing/layers/to_ragged.py
```diff
@@ -0,0 +1,72 @@
...
+        if isinstance(inputs, tf.Tensor):
+            inputs = tf.RaggedTensor.from_tensor(inputs)
+
+        ignore_value = self.ignore_value
+        if ignore_value is None:
+            if inputs.dtype == tf.string:
+                ignore_value = ""
+                inputs = tf.strings.split(inputs, sep=self.sep).values
+            elif inputs.dtype.is_integer:
+                ignore_value = -1
```
chunyang-wen: what about other `dtype`? It seems that it is not complete. And you can add a process function for `ignore_value` instead of testing and setting it everywhere. All you have to do is to gather enough information and return a valid `ignore_value`.

elasticdl_preprocessing/tests/to_ragged_test.py
```diff
@@ -0,0 +1,34 @@
...
+import tensorflow as tf
+
+from elasticdl_preprocessing.layers.to_ragged import ToRagged
+from elasticdl_preprocessing.tests.test_utils import ragged_tensor_equal
+
+
+class ToRaggedTest(unittest.TestCase):
+    def test_dense_to_ragged(self):
+        layer = ToRagged()
+        inp = tf.constant([[1], [-1], [4]], tf.int64)
```
chunyang-wen: Use abbreviations when they are commonly used, otherwise, you are making it hard to read the code. inp for input?

elasticdl_preprocessing/tests/to_ragged_test.py
```diff
@@ -0,0 +1,34 @@
...
+        out = layer(inp)
+        expected_out = tf.ragged.constant([[1], [], [4]], tf.int64)
+        self.assertTrue(ragged_tensor_equal(out, expected_out))
```
chunyang-wen, line 15: I find that `RaggedTensor` has a member function `to_list`. You can convert the tensor to a numpy array and use a normal `assertListEqual` to compare them. I think this function will give more information when anything is wrong. BTW, I have not check the code for `ragged_tensor_equal`.

elasticdl_preprocessing/layers/to_ragged.py
workingloong: Yeah, Done!

elasticdl_preprocessing/tests/to_ragged_test.py
workingloong: `ragged_tensor_equal` checks the dtypes and contents of `RaggedTensor`. It is more simple to check the contents by `to_list` than the implementation in `ragged_tensor_equal`. I will use `to_list` to compare. Thank you for the suggestion!

brightcoder01 (approved):

PR #1863, opened 2020-03-20
Discretization layer to bucket data into discrete ranges

workingloong:


brightcoder01 (approved):

elasticdl_preprocessing/layers/discretization.py
```diff
@@ -0,0 +1,59 @@
...
+import tensorflow as tf
+from tensorflow.python.ops import math_ops
+
+
+class Discretization(tf.keras.layers.Layer):
+    """Buckets data into discrete ranges.
+
+    TensorFlow 2.2 has developed `tf.keras.layers.preprocessing.Discretization`
+    but not released it yet. So the layer is a simple temporary version.
+    https://github.com/tensorflow/tensorflow/blob/r2.2/tensorflow/python/keras/layers/preprocessing/discretization.py
```
terrytangyuan: Cannot you use it via `tensorflow.python.xxxx`, which would be better than copy-pasting code?

workingloong: Good suggestion!

PR #1865, opened 2020-03-21
The metadata sample from SQLFlow transform expression and two corresponding models with keras preprocessing layers and feature columns.

brightcoder01:


model_zoo/census_model_sqlflow/wide_and_deep/wide_deep_functional_keras.py
```diff
@@ -0,0 +1,299 @@
...
+        input_layers[schema_info.name] = tf.keras.layers.Input(
+            name=schema_info.name, shape=(1,), dtype=schema_info.dtype
+        )
+
+    return input_layers
+
+
+# Build the transform logic from the metadata in feature_configs.py.
+def transform(inputs):
+    transformed = inputs.copy()
```
workingloong, line 83: Why do you copy `inputs` to transformed not `transformed = inputs`?

model_zoo/census_model_sqlflow/wide_and_deep/transform_ops.py
```diff
@@ -0,0 +1,74 @@
...
+    BUCKETIZE = (2,)
+    LOOKUP = (3,)
+    EMBEDDING = (4,)
+    CONCAT = (5,)
+    ARRAY = 6
+
+
+class TransformOp(object):
+    def __init__(self, name, input, output):
+        self.name = name
```
workingloong, line 15: Can the `name` and `output` be optional? We can set a default value according to input for `name` and `output`. For example, the `name` and `output` can be the same as `input` because the `input` is unique.

model_zoo/census_model_sqlflow/wide_and_deep/transform_ops.py
```diff
@@ -0,0 +1,74 @@
...
+        # The output name
+        self.output = output
+        # The type of the TransformOp.
+        # The value is one of `TransformOpType`.
+        self.op_type = None
+
+
+class Vocabularize(TransformOp):
+    def __init__(
+        self, name, input, output, vocabulary_list=None, vocabulary_file=None
```
workingloong, line 35: We can merge `vocabulary_list` and `vocabulary_file` to a single argument `vocabulary`. Because the `IndexLookup` only has `vocabulary`.
https://github.com/sql-machine-learning/elasticdl/blob/402de98127621733ee2f1d8b2238a8ea681e0691/elasticdl_preprocessing/layers/index_lookup.py#L33-L35

model_zoo/census_model_sqlflow/wide_and_deep/transform_ops.py
```diff
@@ -0,0 +1,74 @@
...
+        # The input name or a list of input names
+        self.input = input
+        # The output name
+        self.output = output
+        # The type of the TransformOp.
+        # The value is one of `TransformOpType`.
+        self.op_type = None
+
+
+class Vocabularize(TransformOp):
```
workingloong, line 25: Add a docstring to explain which Keras layer or feature column the transformation operator use

model_zoo/census_model_sqlflow/wide_and_deep/transform_ops.py
```diff
@@ -0,0 +1,74 @@
...
+        self, name, input, output, vocabulary_list=None, vocabulary_file=None
+    ):
+        super(Vocabularize, self).__init__(name, input, output)
+        self.op_type = TransformOpType.LOOKUP
+        self.vocabulary_list = vocabulary_list
+        self.vocabulary_file = vocabulary_file
+
+
+class Bucketize(TransformOp):
+    def __init__(self, name, input, output, num_buckets=None, boundaries=None):
```
workingloong, line 51: `boundaries` may be enough and we can infer the `num_buckets` according to `boundaries`. I wonder that the argument names in Keras preprocessing layers and feature columns are different. For example, the name is `bins` in `Discretization` 
https://github.com/sql-machine-learning/elasticdl/blob/402de98127621733ee2f1d8b2238a8ea681e0691/elasticdl_preprocessing/layers/discretization.py#L20-L23
and  is `boundaries` in `bucketized_column`.
```python
tf.feature_column.bucketized_column(
    source_column, boundaries
)
```

model_zoo/census_model_sqlflow/wide_and_deep/wide_deep_functional_keras.py
```diff
@@ -0,0 +1,299 @@
...
+
+    for feature_transform_info in FEATURE_TRANSFORM_INFO_EXECUTE_ARRAY:
+        if feature_transform_info.op_type == TransformOpType.HASH:
```
workingloong, line 86: We can add a `transform` method to `TransformOp` and the `transform` method can use Keras preprocessing layers or feature columns to transform the inputs. Then, we don't need to decide using which layer by `if`.

model_zoo/census_model_sqlflow/wide_and_deep/transform_ops.py
```diff
@@ -0,0 +1,74 @@
+from enum import Enum
+
+
+class TransformOpType(Enum):
```
typhoonzero, line 4: Seems transform ops will also be used in SQLFlow in the future, so can `TransformOp` be an independent python package so that we can install it by `pip install` or will this be replaced by Keras transform layers?

brightcoder01: Yes, the `TransformOp` will be in the SQLFlow code. It's used in the code_generator component.
I just want to describe the data structure and choose to write here.

model_zoo/census_model_sqlflow/wide_and_deep/transform_ops.py
brightcoder01: If we are using feature column, we may differentiate whether it's a vocabulary list or file.

model_zoo/census_model_sqlflow/wide_and_deep/transform_ops.py
brightcoder01: I prefer to decouple input and output name. 
For one input, we may apply two different transform actions and get two different outputs.
BTW, the output name is auto-generated.

model_zoo/census_model_sqlflow/wide_and_deep/wide_deep_functional_keras.py
brightcoder01: If we don't add copy() here. 
We will update the input layers directly.
1. The value of the existed key will be updated: such as `age: <tensorflow.python.framework.sparse_tensor.SparseTensor object at 0x13b237a90>`
2. More keys will be added in the input layers such as age_bucketize.

And then the updated input layers will be fed into the keras.Model(inputs=inputs, xxx), the input layers for the model is not the origin input features.

model_zoo/census_model_sqlflow/wide_and_deep/transform_ops.py
brightcoder01: My original opinion is that each TransformOP contains the fields from different source
1. User provide in the SQLFlow `COLUMN` expression. User may provide `num_buckets` in the SQLFlow statement.
2. Parameters we infer from data analysis
3. Parameters by walking through the Transform Graph.

model_zoo/census_model_sqlflow/wide_and_deep/transform_ops.py
brightcoder01: Good idea.

model_zoo/census_model_sqlflow/wide_and_deep/transform_ops.py
workingloong: I wonder whether we will provide users with those `TransformOP`s to develop preprocessing logic in their model definition. If it is, the required arguments are as less as possible. So, it may be better to auto-generate `output` and `name`.

model_zoo/census_model_sqlflow/wide_and_deep/transform_ops.py
workingloong: We can differentiate it by the type of `vocabulary`.

model_zoo/census_model_sqlflow/wide_and_deep/transform_ops.py
workingloong: Yeah. I also think how users utilize those TransformOPs to develop their models. So, it need to make it user-friendly.

brightcoder01: I see, we can add `required` keyword for the necessary parameters.
What's more, the TransformOp in this python file is a sample describing the data structure which has the same fields in GO. For E2E ML on SQLFlow, there is no code containing the `TransformOp` in python file.

model_zoo/census_model_sqlflow/wide_and_deep/transform_ops.py
brightcoder01: In my original opinion, I won't publish this `TransformOP` to users.
If you are planning to do so, you can simplify based on this.

model_zoo/census_model_sqlflow/wide_and_deep/transform_ops.py
brightcoder01: Just like the information in other comments, the TransformOp in this python file is a sample describing the data structure which has the same fields in GO. It may be clearer if we differentiate whether it's a list of file name in the code generator component.

model_zoo/census_model_sqlflow/wide_and_deep/wide_deep_functional_keras.py
brightcoder01: Good idea, you can try adding the code according to this idea in the next PR? It maybe useful if we expose the `TransformOP` structure and generic `transform` function to users.

For code_generator, we are focusing on `def transform_from_code_gen(inputs)` function.

workingloong (approved):

PR #1868, opened 2020-03-23
Only check PS channels when not under allreduce strategy

terrytangyuan:
This `TimeoutError` happens even under allreduce strategy. This PR fixes this so we only check PS channels when not under allreduce strategy.

```
  File "/usr/lib/python3.6/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/lib/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/elasticdl/python/worker/main.py", line 63, in <module>
    main()
  File "/elasticdl/python/worker/main.py", line 50, in main
    % addr.split(".")[0]
TimeoutError: Time out to connect pod elasticdl-test-train-ps-0 with 3 retries
```

### DIFF
--- a/elasticdl/python/worker/main.py
+++ b/elasticdl/python/worker/main.py
@@ -2,7 +2,6 @@ import grpc
 
 from elasticdl.python.common import log_utils
 from elasticdl.python.common.args import parse_worker_args
-from elasticdl.python.common.constants import DistributionStrategy
 from elasticdl.python.common.grpc_utils import build_channel
 from elasticdl.python.worker.worker import Worker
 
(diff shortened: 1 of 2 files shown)


elasticdl/python/worker/main.py
```diff
@@ -19,7 +20,10 @@ def main():
     master_channel = build_channel(args.master_addr)
 
     ps_channels = []
-    if args.ps_addrs:
+    if (
+        args.ps_addrs
```
skydoorkai: Why `DistributionStrategy.ALLREDUCE` still has `ps_addrs`? `ps_addrs` should set to `` for ALLREDUCE.

terrytangyuan: Thanks. PR updated to set it to empty string instead.

skydoorkai (approved):

PR #1873, opened 2020-03-24
Fix small typo

tonyyang-svail:


terrytangyuan (approved):

docs/designs/data_transform.md
```diff
@@ -66,13 +66,13 @@ SELECT * FROM new_plates TO PREDICT price USING a_model;
...
 ```
 
 Or, if she wants standardization instead of normalization, she could write the following.
 
 ```sql
-SELECT * FROM new_plate TO PREDICT price USING a_model COLUMN standardize(size)
+SELECT * FROM new_plates TO PREDICT price USING a_model COLUMN standardize(size)
 ```
 
 How if the table new_plate doesn't have the field used to train the model? How if there is no field named `size`, but a field `diameter`. We should allow users to use the latter as well.
```
brightcoder01: How about update `new_plate` in this sentence as well? :)

brightcoder01 (approved):

PR #1878, opened 2020-03-27
support mixed priority of worker pods

QiJune:
This PR aims to support mixed priority of worker pods. For example, one user wants to launch 10 worker pods, and sets  worker pod priority to `high=0.5`, the master will create 5 worker pods with high priority and 5 worker pods with low priority.

This mechanism ensures that an elasticdl job could keep running, even when most worker pods with low priority are preempted.

### DIFF
--- a/elasticdl/python/master/k8s_instance_manager.py
+++ b/elasticdl/python/master/k8s_instance_manager.py
@@ -12,13 +12,22 @@ _SERVICE_ADDR_SEP = ","
 def _parse_worker_pod_priority(num_workers, worker_pod_priority):
     res = {}
     if isinstance(worker_pod_priority, str) and "high=" in worker_pod_priority:
-        fraction = float(worker_pod_priority.split("=")[1])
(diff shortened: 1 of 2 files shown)


elasticdl/python/master/k8s_instance_manager.py
```diff
@@ -9,6 +9,22 @@
 _SERVICE_ADDR_SEP = ","
 
 
+def _parse_worker_pod_priority(num_workers, worker_pod_priority):
+    res = {}
+    if isinstance(worker_pod_priority, str) and "high=" in worker_pod_priority:
+        fraction = float(worker_pod_priority.split("=")[1])
```
skydoorkai: This line may cause ValueError exception if the argument is not in the correct form.
use try to catch exception. If caught, log a warning and use a low priority for all pods.

QiJune: Done

skydoorkai (approved):

PR #1879, opened 2020-03-27
expose termination_grace_period_seconds of worker pod

QiJune:
The worker pods with low priority of an elasticdl job could be preempted usually. The default value of terminationGracePeriodSeconds is 30s, which is too long for most senarios. Since worker pods is stateless, it could be killed immediately by setting terminationGracePeriodSeconds to 1s.

### DIFF
--- a/elasticdl/python/common/args.py
+++ b/elasticdl/python/common/args.py
@@ -154,6 +154,12 @@ def add_common_params(parser):
         "default to worker_resource_request",
     )
     parser.add_argument(
+        "--worker_termination_period",
+        type=int,
+        default=30,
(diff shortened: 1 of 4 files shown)


skydoorkai:
We can just set all worker pod terminationGracePeriodSeconds to 1s.
K8s terminationGracePeriodSeconds is used in case the pod needs to do some extra steps (like saving something) before get killed. 
ElasticDL is fault-tolerant and does not need the extra seconds in termination.

QiJune:
@skydoorkai  Done, set all worker pod terminationGracePeriodSeconds to 1s.

skydoorkai (approved):

PR #1881, opened 2020-03-30
add debug_params arg for benchmark experiments

QiJune:
In benchmark experiments, we need to create a control group case that running ElasticDL jobs without relaunching mechanism. It is designed for compare with the study group.

This PR add a debug_params for ElasticDL, it supports key-value pair config. So we could add any needed params conveniently.

### DIFF
--- a/elasticdl/python/common/args.py
+++ b/elasticdl/python/common/args.py
@@ -281,12 +281,13 @@ def add_common_params(parser):
         help="True for Go-based PS, False for Python-based PS",
     )
     parser.add_argument(
-        "--debug_params",
+        "--aux_params",
         type=str,
         default="",
-        help="Only for debug. The debug parameters in a string separated "
+        help="Auxiliary parameters for misc purposes such as debugging."
+        "The auxiliary parameters in a string separated "
         'by semi-colon used to debug , e.g. "param1=1; param2=2" '
-        "Supported debug parameters: disable_relaunch",
+        "Supported auxiliary parameters: disable_relaunch",
     )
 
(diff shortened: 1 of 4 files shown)


elasticdl/python/common/args.py
```diff
@@ -280,6 +280,14 @@ def add_common_params(parser):
         default=False,
         help="True for Go-based PS, False for Python-based PS",
     )
+    parser.add_argument(
+        "--debug_params",
```
skydoorkai: `aux_params` is better.
auxiliary parameters for misc purposes such as debugging.

QiJune: Done

skydoorkai (approved):

elasticdl/python/master/master.py
```diff
@@ -455,6 +455,11 @@ def _create_instance_manager(self, args):
             for key in env_dict:
                 env.append(V1EnvVar(name=key, value=env_dict[key]))
 
+            kwargs = get_dict_from_params_str(args.aux_params)
+            disable_relaunch = (
+                kwargs.get("disable_relaunch", False) if kwargs else False
```
chunyang-wen: disable_relaunch = kwargs.get("disable_relance", False)

chunyang-wen (changes requested):

elasticdl/python/master/master.py
QiJune: Done

chunyang-wen (approved):

Issue #1885, opened 2020-03-30
The interface/bridge between generated transform code and model definition from model zoo.


brightcoder01:
The model definition in the model zoo is a python function or a python class. It has some input parameters.

The transform python code is built upon feature column api or keras preprocessing layers. And it can be generated from the `COLUMN` expression in SQLFlow.

We will get the outputs from transform code and pass them as the input parameters of the model definition. And then we can get the complete model and it's ready to train.

In the original design of model zoo, the input parameters of model definition are a list of feature columns. This solution can't cover the transform logic built upon keras preprocess layers.
**Our goal is to find an interface/bridge design to cover the transform logic using both feature columns and keras preprocessing layers.**

##### Functional Model
`Tensors` is a good choice to bridge the transform code and model definition. The output of the transform logic for one `COLUMN` expression is one tensor, and two expressions output two tensors.  Just like the following SQL statement:
```SQL
SELECT *
FROM census_income
TO TRAIN WideAndDeepClassifier
COLUMN
    COMBINE(
        EMBEDDING(CONCAT(VOCABULARIZE(workclass), BUCKETIZE(capital_gain, num_buckets=5), BUCKETIZE(capital_loss, num_buckets=5), BUCKTIZE(hours_per_week, num_buckets=6)) AS group_1, 8),
        EMBEDDING(CONCAT(HASH(education), HASH(occupation), VOCABULARIZE(martial_status), VOCABULARIZE(relationship)) AS group_2, 8),
        EMBEDDING(CONCAT(BUCKETIZE(age, num_buckets=5), HASH(native_country), VOCABULARIZE(race), VOCABULARIZE(sex)) AS group_3, 8)
    ) FOR deep_embedding,
    COMBINE (
        EMBEDDING(group1, 1),
        EMBEDDING(group2, 1)
    ) FOR wide_embedding
LABEL label
```
It will output two tensors: `deep_embedding` and `wide_embedding`. We can check the sample code in #1887 

##### Subclass Model
The model have only one method `def call(self, inputs)` to describe the forward_pass logic. We cannot infer how many tensors to input into the model from this API. So we propose to add a decorator `@model_input_names` on the model class to inject one property `_model_input_names` - a list of string to indicate how many tensors the model accepts and their names. Please check #1920

brightcoder01:
Already fixed in PR #1920

PR #1887, opened 2020-03-31
Add the sample model to describe that the interface between transform code and model definition is tensors.

brightcoder01:
Relate #1885 

Model definition from model zoo
```python
def wide_and_deep_classifier(input_layers, wide_embedding, deep_embedding):
```

Generated transform python code
```python
def transform_from_code_gen(source_inputs):
```

### DIFF
--- a/elasticdl_preprocessing/layers/concatenate_with_offset.py
+++ b/elasticdl_preprocessing/layers/concatenate_with_offset.py
@@ -35,11 +35,10 @@ class ConcatenateWithOffset(tf.keras.layers.Concatenate):
         self.axis = axis
 
     def call(self, inputs):
-        return (
-            self._call_with_none_offsets(inputs)
-            if self.offsets is None
-            else self._call_with_offsets(inputs)
-        )
+        if self.offsets is None:
+            return self._call_with_none_offsets(inputs)
+
+        return self._call_with_offsets(inputs)
 
     def _call_with_offsets(self, inputs):
         ids_with_offset = []
--- a/elasticdl_preprocessing/layers/concatenate_with_offset.py
+++ b/elasticdl_preprocessing/layers/concatenate_with_offset.py
@@ -35,6 +35,13 @@ class ConcatenateWithOffset(tf.keras.layers.Concatenate):
         self.axis = axis
 
     def call(self, inputs):
+        return (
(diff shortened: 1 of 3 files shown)


model_zoo/census_model_sqlflow/wide_and_deep/wide_deep_functional_tensor_interface_fc.py
```diff
@@ -0,0 +1,315 @@
...
+        )
+
+    return input_layers
+
+
+# Build the transform logic from the metadata in feature_configs.py.
+def transform(inputs):
+    feature_column_dict = {}
+
+    for feature_transform_info in FEATURE_TRANSFORM_INFO_EXECUTE_ARRAY:
```
typhoonzero, line 76: Is `FEATURE_TRANSFORM_INFO_EXECUTE_ARRAY` some data only used for testing? In real cases, does `FEATURE_TRANSFORM_INFO_EXECUTE_ARRAY` should be generated by the `COLUMN` clause? Also the `TRANSFORM_OUTPUTS`.

typhoonzero (approved):
LGTM++

elasticdl_preprocessing/layers/concatenate_with_offset.py
```diff
@@ -35,6 +35,13 @@ def __init__(self, offsets, axis=-1):
         self.axis = axis
 
     def call(self, inputs):
+        return (
+            self._call_with_none_offsets(inputs)
+            if self.offsets is None
+            else self._call_with_offsets(inputs)
```
chunyang-wen: It is better not to write complex clauses in a single line.

brightcoder01: Good suggestion, updated.

model_zoo/census_model_sqlflow/wide_and_deep/wide_deep_functional_tensor_interface_fc.py
brightcoder01: FEATURE_TRANSFORM_INFO_EXECUTE_ARRAY and TRANSFORM_OUTPUTS won't exist in the generated code.
I think it's the intermediate object during code generation. It should be exist in GO.

PR #1892, opened 2020-04-03
Add elasticdl_preprocessing layers in package data

workingloong:


setup.py
```diff
@@ -28,6 +28,7 @@
             "Makefile",
             "requirements.txt",
             "pkg/kernel/capi/*",
+            "elasticdl_preprocessing/layers/*",
```
brightcoder01: Why not elasticdl_preprocessing/*

workingloong: "package_data" can only contain files. Here, we can use "package_dir", https://docs.python.org/2/distutils/setupscript.html#listing-whole-packages

brightcoder01 (approved):

brightcoder01 (commented):
Will accept if the CI pass.

workingloong:
There is `__init__.py` in the elasticdl_preprocessing directory and the python files in elasticdl_preprocessing will be packaged when building a wheel. So, we don't need the configuration.

PR #1894, opened 2020-04-07
Rethink end-to-end feature derivation syntax

wangkuiyi:
We used to think that each COLUMN clause maps to a TensorFlow feature column API call.  However, feature column API is TF-only, and indeed, future versions of TF might not longer support them.

Considering that we want this end-to-end ML work portable to PyTorch, XGBoost, and other ML toolkit, how about we abstract the model as a Python function with one or more input parameters: 

```python
def forward_pass_of_a_model(x, y, z=[0]):
    return do_something_with(x,y,z)
```

We hope that given only one COLUMN clause, but maybe more than one AS sub-clauses, each AS maps to a named Python function parameter like x, y, or z.

For example, the following COLUMN clause

```sql
COLUMN embedding(...) AS x, embedding(...) AS y, embedding(...) AS z
```
 maps the outputs of the three embedding layers to x, y, and z respectively.

And the following clause

```sql
COLUMN embedding(...), embedding(...), embedding(...)
```

maps the three outputs to x, y, and z respectively too, but by their positions.

The following case

```sql
COLUMN embedding(...), embedding(...)
```

maps the outputs of the two embeddings to x and y, and leaves z to its default value `[0]`.

### DIFF
--- a/docs/designs/data_transform.md
+++ b/docs/designs/data_transform.md
@@ -199,14 +199,30 @@ SELECT *
 FROM census_income
 TO TRAIN WideAndDeepClassifier
 COLUMN
-    EMBEDDING(CONCAT(VOCABULARIZE(workclass), BUCKETIZE(capital_gain, num_buckets=5), BUCKETIZE(capital_loss, num_buckets=5), BUCKTIZE(hours_per_week, num_buckets=6)) AS group_1, 8),
-    EMBEDDING(CONCAT(HASH(education), HASH(occupation), VOCABULARIZE(martial_status), VOCABULARIZE(relationship)) AS group_2, 8),
-    EMBEDDING(CONCAT(BUCKETIZE(age, num_buckets=5), HASH(native_country), VOCABULARIZE(race), VOCABULARIZE(sex)) AS group_3, 8)
-    FOR deep_embeddings
-COLUMN
+  pan(
+    EMBEDDING(
+        CONCAT(VOCABULARIZE(workclass), 
+               BUCKETIZE(capital_gain, num_buckets=5), 
+               BUCKETIZE(capital_loss, num_buckets=5), 
+               BUCKETIZE(hours_per_week, num_buckets=6)) AS group_1, 
+        8),
+    EMBEDDING(
+        CONCAT(HASH(education), 
+               HASH(occupation), 
+               VOCABULARIZE(martial_status), 
+               VOCABULARIZE(relationship)) AS group_2, 
+        8),
+    EMBEDDING(
+        CONCAT(BUCKETIZE(age, num_buckets=5), 
+               HASH(native_country), 
+               VOCABULARIZE(race), 
+               VOCABULARIZE(sex)) AS group_3, 
+        8)
+    ) AS deep_embeddings
+  pan(
     EMBEDDING(group1, 1),
     EMBEDDING(group2, 1)
-    FOR wide_embeddings
+  ) AS wide_embeddings
 LABEL label
 ```
 


wangkuiyi:
We might need to change the current grammar rule

```yacc
column_clause
: COLUMN columns 				{ $$ = map[string]ExprList{"feature_columns" : $2} }
| COLUMN columns FOR IDENT 			{ $$ = map[string]ExprList{$4 : $2} }
| column_clause COLUMN columns FOR IDENT 	{ $$[$5] = $3 }
;
```

into something like

```yacc
column_clause
: COLUMN columns
;

columns
: expr {}
: expr AS IDENT {}
: columns "," expr {}
: columns "," expr AS INDENT {}
;
```

```sql
COLUMN embedding(...) AS x, embedding(...)
```

typhoonzero:
- We use fields specified by `SELECT` clause as model inputs (including the label), e.g. `SELECT a,b,c,label .... COLUMN embedding(hash(b),128,"sum") ...` , column a and c will generate pre-processing layers automatically by feature derivation, and the input fields looks like `numeric(a),embedding(hash(b)), numeric(c)`. In fact some columns like enum columns can be directly translated into an "embedding" column. So using a list by order can not implement this feature.
- One input field (one column) can be used as several input features, e.g. `user_id` feature can be used as an embedding feature, then as a part of a cross feature.

brightcoder01:
After further consideration, users may be confusing about how many tensor the `COLUMN` expression outputs. Please check the following sample clause, the style is the same, but the expected outputs are different:

COLUMN age, STANDARDIZE(capital_gain), EMBEDDING(HASH(workclass)) => User expects the output is one tensor.

COLUMN PAD(EMBEDDING(…), EMBEDDING(…)), PAD(EMBEDDING(…), EMBEDDING(…)) => User expected the output is two tensors.

docs/designs/data_transform.md
```diff
@@ -199,14 +199,30 @@ SELECT *
...
+               HASH(native_country), 
+               VOCABULARIZE(race), 
+               VOCABULARIZE(sex)) AS group_3, 
+        8)
+    ) AS deep_embeddings
+  pan(
     EMBEDDING(group1, 1),
     EMBEDDING(group2, 1)
-    FOR wide_embeddings
+  ) AS wide_embeddings
```
brightcoder01, line 225: ### Problem

After further consideration, users may be confused about how many tensors the `COLUMN` expression outputs. Please check the following sample clauses, the styles of these two clause are the same, but the expected outputs are different:

User expects the output is one tensor.
```
COLUMN age, STANDARDIZE(capital_gain), EMBEDDING(HASH(workclass))
```

User expected the output contains two tensors.
```
COLUMN PAD(EMBEDDING(…), EMBEDDING(…)), PAD(EMBEDDING(…), EMBEDDING(…))
```

### Proposal

##### One COLUMN clause
If we choose one `COLUMN` clause and want to output multiple tensors from it, we need separate this clause into multiple parts using a kine of markers such as `AS` keyword, each part outputs one tensor.
For example:

The following clause outputs two tensors.
```
COLUMN PAD(EMBEDDING(…), EMBEDDING(…)) AS wide_embeddings, PAD(EMBEDDING(…), EMBEDDING(…)) AS deep_embeddings
```

If we don't use keyword `AS` to separate them, the following clause output one tensor.
```
COLUMN PAD(EMBEDDING(…), EMBEDDING(…)), PAD(EMBEDDING(…), EMBEDDING(…))
```

##### Multiple COLUMN clauses

Keep the original design and choose multiple COLUMN clause. One COLUMN clause outputs one tensor.
```
COLUMN EMBEDDING(...), EMBEDDING(...) FOR wide_embeddings,
COLUMN EMBEDDING(...), EMBEDDING(...) FOR deep_embeddings
```

brightcoder01: After further discussion with more SQLFlow teammates, we choose this following syntax design: **keep one COLUMN clause**; If we want to output multiple tensors in this clause, **use keyword `FOR` to indicate the output tensor name**. Please check the following example:

```
COLUMN
  pan(
    EMBEDDING(
        CONCAT(VOCABULARIZE(workclass), 
               BUCKETIZE(capital_gain, num_buckets=5), 
               BUCKETIZE(capital_loss, num_buckets=5), 
               BUCKETIZE(hours_per_week, num_buckets=6)) AS group_1, 
        8),
    EMBEDDING(
        CONCAT(HASH(education), 
               HASH(occupation), 
               VOCABULARIZE(martial_status), 
               VOCABULARIZE(relationship)) AS group_2, 
        8),
    EMBEDDING(
        CONCAT(BUCKETIZE(age, num_buckets=5), 
               HASH(native_country), 
               VOCABULARIZE(race), 
               VOCABULARIZE(sex)) AS group_3, 
        8)
    ) FOR deep_embeddings,
  pan(
    EMBEDDING(group1, 1),
    EMBEDDING(group2, 1)
  ) FOR wide_embeddings
LABEL label
```

In the COLUMN clause, we use two keyword `AS` and `FOR`. Why don't we merge them into one - `AS`? Because they have different meanings.
1. `AS` is the alias of a part of transform logic. The alias can be shared by multiple transform function call.
2. `FOR` means this part of transform logic will output a tensor. The IDENT after `FOR` is the tensor name.

workingloong: what's mean of `pan`?

brightcoder01: > what's mean of `pan`?

`pan` is a temporary name which means a transform function to merge multiple tensors into one. We need figure out another more accurate name to describe the transform logic.

brightcoder01:
Close this PR for discussion.
We have already gotten the [conclusion](https://github.com/sql-machine-learning/elasticdl/pull/1894#discussion_r406187759) about the COLUMN syntax. The updated syntax for COLUMN clause is included in #1900

PR #1900, opened 2020-04-07
Add COLUMN expression compilation workflow in data_transform design doc.

brightcoder01:


docs/designs/data_transform.md
```diff
@@ -210,64 +224,101 @@ COLUMN
...
 
-SQLFlow will generate the analysis SQL to calculate the statistical value. For this clause `COLUMN NORMALIZE(capital_gain), STANDARDIZE(age)`, the corresponding analysis SQL is as follows:
+#### Parse COLUMN Expression into AST
+
+Given a COLUMN expression, syntax parsing is naturally the first step. We will parse one COLUMN expression to one AST. For the SQL statement containing two column expressions above, we will parse them into two ASTs. Please check the following figure of the AST for the example SQL.  
+![column_clause_ast](../images/column_clause_ast.png)
+
+#### Convert the AST into a DAG of Transform Flow
+
+To convert ASTs into transform flow, we will do a one-on-one mapping at first.
```
workingloong: `transform flow` -> `transform flows`
`a one-on-one` -> `an one-to-one`

docs/designs/data_transform.md
```diff
@@ -210,64 +224,101 @@ COLUMN
...
+SELECT DISTINCT(x)
+FROM data_table;
+```
 
-The model definition in model zoo is a Python class derived from `tf.keras.Model`. It is compatible with various feature input. Please check the [sample model](https://github.com/sql-machine-learning/models/blob/1b6f5eaabd57b8ca682e60c1dde4cd4ec1053bf9/sqlflow_models/dnnclassifier.py#L3).  
+HASH(x)
 
-The transform code generated from `COLUMN` clause specifies how to convert the SELECT result into model inputs in the form of tensors.
+```SQL
+SELECT (COUNT(DISTINCT(x)) * 3) AS x_hash_bucket_size
```
workingloong, line 298: Explain why we need to multiply by 3.

docs/designs/data_transform.md
```diff
@@ -210,64 +224,103 @@ COLUMN
...
+#### Parse COLUMN Expression into AST
+
+Given a COLUMN expression, syntax parsing is naturally the first step. We will parse one COLUMN expression to one AST. For the SQL statement containing two column expressions above, we will parse them into two ASTs. Please check the following figure of the AST for the example SQL.  
+![column_clause_ast](../images/column_clause_ast.png)
+
+#### Convert the AST into a DAG of Transform Flow
+
+To convert ASTs into transform flow, we will do a one-on-one mapping at first.
+
+1. Transform Function(AST) to TransformOP(Transform Flow)
```
typhoonzero: Maybe describe when to do "Derive the Parameters of TransformOP" in this list?

docs/designs/data_transform.md
```diff
@@ -210,64 +224,101 @@ COLUMN
...
-### Generate Analysis SQL From SQLFlow Statement
+### COLUMN Expression Compilation Workflow
 
-SQLFlow will generate the analysis SQL to calculate the statistical value. For this clause `COLUMN NORMALIZE(capital_gain), STANDARDIZE(age)`, the corresponding analysis SQL is as follows:
+#### Parse COLUMN Expression into AST
+
+Given a COLUMN expression, syntax parsing is naturally the first step. We will parse one COLUMN expression to one AST. For the SQL statement containing two column expressions above, we will parse them into two ASTs. Please check the following figure of the AST for the example SQL.  
+![column_clause_ast](../images/column_clause_ast.png)
+
+#### Convert the AST into a DAG of Transform Flow
```
wangkuiyi, line 238: I am afraid that SQLFlow extended syntax parser doesn't generate AST. 
Would this fact affect the design here?

docs/designs/data_transform.md
```diff
@@ -210,64 +224,103 @@ COLUMN
...
+We can make typological sort on the DAG and get an ordered list. Then we generate the python code according to this order. A TransformOP node will generate a line of python code. A line of python code is an api call to [feature column](https://www.tensorflow.org/versions/r2.2/api_docs/python/tf/feature_column) or [keras preprocessing layer](https://github.com/tensorflow/community/pull/188) from TensorFlow native API and our extensions in [elasticdl_preprocessing](https://github.com/sql-machine-learning/elasticdl/tree/develop/elasticdl_preprocessing). Please check the sample code generated from the example SQL: [keras preprocessing layer version](https://github.com/sql-machine-learning/elasticdl/blob/84bf8026565df81521ffdfe55d854428fb1156d4/model_zoo/census_model_sqlflow/wide_and_deep/wide_deep_functional_tensor_interface_keras.py#L129-L228) and [feature column version](https://github.com/sql-machine-learning/elasticdl/blob/84bf8026565df81521ffdfe55d854428fb1156d4/model_zoo/census_model_sqlflow/wide_and_deep/wide_deep_functional_tensor_interface_fc.py#L130-L243).
```
typhoonzero, line 312: We assume to use the order from `SELECT ...` statement now since it's more readable and simple.

docs/designs/data_transform.md
```diff
@@ -210,64 +224,103 @@ COLUMN
...
 
-|            Name             |                      Feature Column Template                                   |      Analyzer      |
-|:---------------------------:|:------------------------------------------------------------------------------:|:------------------:|
-|       STANDARDIZE(x)        | numeric_column({var_name}, normalizer_fn=lambda x : x - {mean} / {std})        |    MEAN, STDDEV    |
-|        NORMALIZE(x)         | numeric_column({var_name}, normalizer_fn=lambda x : x - {min} / {max} - {min}) |      MAX, MIN      |
-|           LOG(x)            | numeric_column({var_name}, normalizer_fn=lambda x : tf.math.log(x))            |         N/A        |
-| BUCKETIZE(x, num_buckets=y) | bucketized_column({var_name}, boundaries={percentiles})                        |     PERCENTILE     |
+For subclass model, please check the [discussion](https://github.com/sql-machine-learning/elasticdl/issues/1867).
```
typhoonzero: Describe how to implement for subclass model here? not refer to an issue.

docs/designs/data_transform.md
brightcoder01: Good idea, add the notes here.

docs/designs/data_transform.md
brightcoder01: Good catch, updated.

docs/designs/data_transform.md
brightcoder01: I think it won't affect the design.
What we want is that we can convert the `COLUMN` clauses into the transform flow DAG here. Maybe we can try to convert the IR into the DAG.

docs/designs/data_transform.md
brightcoder01: I'm afraid not.

The order of the python code is the same with the order of the TransformOP execution. One TransformOP (not an item from the source table) node generate a line of python code. 

The order from SELECT ... statement is the order of items from the source table. And we won't generate the code according to this information.

docs/designs/data_transform.md
brightcoder01: > Maybe describe when to do "Derive the Parameters of TransformOP" in this list?

Good idea. I'll add a note to mention Derive the parameter of TransformOP from data analysis and talk about it in detail in the next steps.

workingloong (approved):

docs/designs/data_transform.md
brightcoder01: Good suggestion, add the content about how to implement subclass model.

PR #1901, opened 2020-04-08
Set create_pod method to public

workingloong:
During data transformation, we need to use k8s client to launch a pod. We can reuse `create_pod` in `k8s_client` to create a pod object and launch it. So, it is better to set `create_pod` to public.


QiJune (approved):
Excellent!

PR #1906, opened 2020-04-10
Add design to preprocess feature inputs using analysis results.

workingloong:

PR #1907, opened 2020-04-13
Add a util to get analysis result from the environment

workingloong:
According to the design #1906, we will set the statistics into the environments of the training pod. So we need to develop an util to get statistics from the environment.


elasticdl_preprocessing/utils/analyzer_utils.py
```diff
@@ -0,0 +1,168 @@
...
+    the environment.
+
+    Args:
+        feature_name: String, feature name or column name in a table
+        default_value: Float.
+
+    Return:
+        Float
+    """
+    env_name = feature_name + "_min"
```
brightcoder01: We can name it feature_name + "__min", because the feature name may contain '_', so we use '__' to separate feature name and analyze result key.

elasticdl_preprocessing/utils/analyzer_utils.py
```diff
@@ -0,0 +1,168 @@
...
+
+    Return:
+        Float
+    """
+    env_name = feature_name + "_max"
+    max_value = os.getenv(env_name, None)
+    if max_value is None:
+        return default_value
+    else:
+        return float(max_value)
```
brightcoder01, line 45: What will happen if max_value can't be converted to float?

elasticdl_preprocessing/utils/analyzer_utils.py
```diff
@@ -0,0 +1,168 @@
...
+
+    Return:
+        List with strings.
+    """
+    env_name = feature_name + "_vocab_path"
+    vocabulary_path = os.getenv(env_name, None)
+    if vocabulary_path is None:
+        return default_value
+    else:
+        return _read_vocabulary_from_file(vocabulary_path)
```
brightcoder01: How about returning file path directly?

workingloong: Returning a vocabulary list is more convenient for users to test the model. If return a file path, the users need to mock a file with vocabulary when they test the  model.

elasticdl_preprocessing/utils/analyzer_utils.py
workingloong: Can you give an example to show the difference between "_" and ""?

elasticdl_preprocessing/utils/analyzer_utils.py
workingloong: The `max_value` is not a numeric value if it cannot be cast to float. I think it should raise an error if the `max_value` is not a numeric value.

elasticdl_preprocessing/utils/analyzer_utils.py
brightcoder01: User can pass either a vocabulary list or a file name.
I means that we don't convert it into a list if it's a file name, let the model or op to do this work.

workingloong: Good Suggestion!

elasticdl_preprocessing/utils/analyzer_utils.py
workingloong: We can use `"_" + feature_name + "_min"` to separate them.

elasticdl_preprocessing/utils/analyzer_utils.py
```diff
@@ -0,0 +1,146 @@
...
+    the environment.
+
+    Args:
+        feature_name: String, feature name or column name in a table
+        default_value: Integer.
+
+    Return:
+        Integer.
+    """
+    env_name = "_" + feature_name + "_count"
```
brightcoder01: "_" + feature_name + "_distinct_count"

elasticdl_preprocessing/utils/analyzer_utils.py
```diff
@@ -0,0 +1,146 @@
...
+    """
+    env_name = "_" + feature_name + "_max"
+    max_value = os.getenv(env_name, None)
+    if max_value is None:
+        return default_value
+    else:
+        return float(max_value)
+
+
+def get_avg(feature_name, default_value):
```
brightcoder01, line 48: get_mean?

brightcoder01 (approved):
Approve. We can address the left comments.

elasticdl_preprocessing/utils/analyzer_utils.py
workingloong: "AVG" is the keyword in SQL but `mean` is not. So I think `get_avg` is more suitable.

Issue #1867, opened 2020-03-23
Create subclass model with preprocessing layers


workingloong:
In #1856, we can generate the codes to implement the preprocessing logic according to COLUMN expressions in SQLFlow. The issue describes that how we define Keras subclass model with the generated preprocessing codes.

We can save the generated codes by "code_gen" into "preprocessing.py"
```python
import numpy as np
import tensorflow as tf

from elasticdl_preprocessing.layers.hashing import Hashing
from elasticdl_preprocessing.layers.concatenate_with_offset import (
    ConcatenateWithOffset
)
def get_input_layers(feature_names):
    input_layers = {}
    for name in feature_names:
        input_layers[name] = tf.keras.Input(name=name, shape=(1, ), dtype=tf.string)
    return input_layers

def preprocess_fn(input_layers):
    output = {}
    for name, input_layer in input_layers.items():
        output[name] = Hashing(num_bins=100)(input_layer)
    concat = ConcatenateWithOffset(offsets=[0,100,200])(output.values())
    embedding = tf.keras.layers.Embedding(300,4)(concat)
    return embedding

def custom_preprocessing_model(feature_names):
    input_layers = get_input_layers(feature_names)
    transform_out = preprocess_fn(input_layers)
    return tf.keras.Model(inputs=input_layers, outputs=transform_out)
```

The model definition in "model_zoo.py" is:
```python
class DNNModel(tf.keras.Model):
    def __init__(self, preprocessing_model=None):
        super(DNNModel, self).__init__()
        self.preprocessing_model = preprocessing_model
        self.dense1 = tf.keras.layers.Dense(3, activation=tf.nn.relu)
        self.dense2 = tf.keras.layers.Dense(2, activation=tf.nn.softmax)

    def call(self, inputs):
        if self.preprocessing_model:
            inputs = self.preprocessing_model(inputs)
        x = self.dense1(inputs)
        return self.dense2(x)
```
The model_zoo.py may be a module from a model package.

We also can generate the code to train the subclass model with the preprocessing model and save the codes in python model file to execute in ElasticDL.
```python
from model_zoo import DNNModel
from preprocessing import custom_preprocessing_model

def custom_model()
    feature_names = ["f1", "f2", "f3"]
    preprocessing_model = custom_preprocessing_model(feature_names)
    dnn_model = DNNModel(preprocessing_model)
    return dnn_model

def loss():
    ...

def optimizer():
    ...

if __name__ == "__main__":
[...]

PR #1915, opened 2020-04-14
The bridge between generated transform code and subclass model definition in model zoo.

brightcoder01:
Relate #1867

### DIFF
--- a/model_zoo/census_model_sqlflow/wide_and_deep/wide_deep_subclass_keras.py
+++ b/model_zoo/census_model_sqlflow/wide_and_deep/wide_deep_subclass_keras.py
@@ -42,6 +42,8 @@ class WideAndDeepClassifier(tf.keras.Model):
 
     # TODO: Add a decorator here to describe the InputSpec
     # of `inputs`. It contains two tensors in this model.
+    # The decorator should be able to inject some metadata of the InputSpec
+    # and model zoo can extract this information from it.
     def call(self, inputs):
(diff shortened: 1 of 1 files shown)


typhoonzero (approved):

model_zoo/census_model_sqlflow/wide_and_deep/wide_deep_subclass_keras.py
```diff
@@ -0,0 +1,244 @@
+import tensorflow as tf
+
+from elasticdl.python.elasticdl.callbacks import LearningRateScheduler
```
typhoonzero: Can we have an example that does not import `elasticdl` so that we can test run this python script on any environment that does not have elasticdl installation

workingloong (approved):

model_zoo/census_model_sqlflow/wide_and_deep/wide_deep_subclass_keras.py
brightcoder01: Good idea, I'll remove "import elasticdl" in this code example.
BTW, we still need import elasticdl_preprocessing library.

brightcoder01: For SparseEmbedding layer, we will move it from elasticdl package to elasticdl_preprocessing package in the next PR.

PR #1916, opened 2020-04-16
Add LogRound layer to convert numeric values to integer IDs

### LINKED ISSUE #1910: support LogRound preprocessing layer
We could use implement it based on Discretization layer with a default bin [0, 2, 4, 8, ...]

workingloong:
Fix #1910

### DIFF
--- a/elasticdl_preprocessing/layers/log_round.py
+++ b/elasticdl_preprocessing/layers/log_round.py
@@ -14,7 +14,7 @@ def log(x, base):
 
 
 class LogRound(tf.keras.layers.Layer):
-    """Cast a numeric feature into a discrete integer value by
+    """Cast a numeric value into a discrete integer value by
     `round(log(x))`.
 
 
--- a/elasticdl_preprocessing/layers/log_round.py
+++ b/elasticdl_preprocessing/layers/log_round.py
@@ -5,6 +5,10 @@ from tensorflow.python.ops.ragged import ragged_functional_ops, ragged_tensor
 def log(x, base):
     x = tf.cast(x, tf.float64)
     numerator = tf.math.log(x)
+
+    if base is None:
+        return numerator
+
     denominator = tf.math.log(tf.constant(base, dtype=numerator.dtype))
     return numerator / denominator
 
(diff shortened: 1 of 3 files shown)


elasticdl_preprocessing/layers/log_round.py
```diff
@@ -0,0 +1,81 @@
...
+        **kwargs: Keyword arguments to construct a layer.
+
+    Input shape: A numeric `Tensor`, `SparseTensor` or `RaggedTensor` of shape
+        `[batch_size, d1, ..., dm]`
+
+    Output shape: An int64 tensor of shape `[batch_size, d1, ..., dm]`
+
+    """
+
+    def __init__(self, num_bins, default_value=0, base=None):
```
brightcoder01, line 40: I think we need a `offset` or `bias` parameter.

elasticdl_preprocessing/tests/log_round_test.py
```diff
@@ -0,0 +1,48 @@
...
+import tensorflow as tf
+
+from elasticdl_preprocessing.layers import LogRound
+from elasticdl_preprocessing.tests.test_utils import (
+    ragged_tensor_equal,
+    sparse_tensor_equal,
+)
+
+
+class RoundIdentityTest(unittest.TestCase):
```
brightcoder01: RoundIdentityTest?

elasticdl_preprocessing/layers/log_round.py
workingloong: Can you give an use case to explain why we need it?

brightcoder01: We will get minus infinite if we pass `0` as input or an exception will be thrown?

workingloong: If the input is zero or a minus, the output of `tf.math.log` is -inf. Then, `tf.where` will replace it with zero to output.
```python
values = tf.where(
            tf.logical_or(values < 0, values >= num_bins),
            x=tf.fill(dims=tf.shape(values), value=default_value),
            y=values,
        )
```
And I have added the test case with input `0` in the unit test.

brightcoder01: We can omit the bias parameter at this time. If user needs, we can add it later.

brightcoder01 (approved):

PR #1917, opened 2020-04-16
Move SparseEmbedding into elasticdl_preprocessing.

workingloong:
After moving the SparseEmbedding layer into elasticdl_preprocessing, users only need to import elasticdl_preprocessing to develop their models.


QiJune (approved):
Excellent!

brightcoder01 (approved):

PR #1920, opened 2020-04-22
Add decorator to declare the input tensor names of keras subclass model.

brightcoder01:
The keras subclass model has the core method to describe the forward pass logic: `def call(self, inputs)`. We cannot get how many tensors the model accepts just from the parameter `inputs`.

As a result, we add decorator on the model class to indicate how many tensors the model accept as input and their names. The tensor names are used to parameter binding for `COLUMN` clause.

1. If ML specialist develop the model logic and verify it using `model.fit`, the decorator is not necessary
2. If ML specialist run the machine learning pipeline locally using SQLFlow / run model unit test using SQLFlow / submit model into SQLFlow model zoo, ML specialist need add the decorator.


model_zoo/census_model_sqlflow/wide_and_deep/decorators.py
```diff
@@ -0,0 +1,12 @@
+def model_input_name(input_names):
```
workingloong: Can we place the decorator into a common util?

brightcoder01: Yes, I think we can put it into a separate python package.
The ML specialist want to commit the model into SQLFlow model zoo. He need add this decorator.

brightcoder01: We can move it to the folder `elasticdl_preprocessing` at first.

elasticdl_preprocessing/utils/decorators.py
```diff
@@ -0,0 +1,12 @@
+def model_input_name(input_names):
```
typhoonzero: Maybe use a "variadic function": https://en.wikipedia.org/wiki/Variadic_function ?

brightcoder01: Good idea, updated.

typhoonzero (approved):
LGTM

Issue #1926, opened 2020-04-24
Fill the missing column with a default value of the specific dtype while creating a dataset for a MaxCompute table.


workingloong:
The column schemas in a MaxCompute table may be different, such as int, float, and string. The output tensor dtype of dataset should be the same as the column schemas. What's more, some columns may contain missing values. For example:

| age (int) | education (string) | height (float) | label  (int)|
| --- | --------- | ---------| -----|
| 40 | HS-grad | 173.5 | 0 |
|  | Master | 184.2 | 1 |
| 56 |  | 169.4 | 0 |

Here, we discuss how to output tensors using `tf.data.Dataset` with the same dtype as column schema dtype and fill the missing value using a default value.

PyODPS tunnel will convert the records to the corresponding type in python according to the column schema. It will return `None` for the missing column. For the above table, PyODPS tunnel will returns 
```python
[[40, "HS-grad", 173.5, 0], [None, "master", 184.2, 1], [56, None, 169.4, 0]]
```

### Solution 1: Create a dataset with column schemas and fill missing values in `transform_fn` in`DataReader`.

https://github.com/sql-machine-learning/elasticdl/blob/eefcb31cb148a44884643d24e9db9d8a4c230856/elasticdl/python/data/odps_io.py#L352-L355

We can custom a `transform_fn` to fill missing value which is None in the records. Then we can create a dataset using `from_generator` like:
```python
dataset = tf.data.Dataset.from_generator(
    gen, [tf.int64, tf.float, tf.float64, tf.int64]
)
```

### Solution 2: Convert the data to string and cast data by column schema in dataset_fn.

We can force to cast data in records to string and convert None to "None". Then we can create a dataset like
```python
dataset = tf.data.Dataset.from_generator(
    gen, tf.string
)
```

Then in `dataset_fn`, we need to cast string values to column schema dtype like.
https://github.com/sql-machine-learning/elasticdl/blob/eefcb31cb148a44884643d24e9db9d8a4c230856/model_zoo/odps_iris_dnn_model/odps_iris_dnn_model.py#L32-L36

Before converting dtype, we need to use `tf.where` to replace "None" with the default value like
```
tf.where(tf.equal(tensor, "None"), tensor, default_value)
```

PR #1927, opened 2020-04-24
Add column dtype in metadata

workingloong:
In the #1926, we may need the column dtypes to create tensor with the same dtype as column schema in a MaxCompute table.

### DIFF
--- a/elasticdl/python/data/reader/odps_reader.py
+++ b/elasticdl/python/data/reader/odps_reader.py
@@ -59,9 +59,10 @@ class ODPSDataReader(AbstractDataReader):
             self._metadata.column_names = (
                 reader._odps_table.schema.names if columns is None else columns
             )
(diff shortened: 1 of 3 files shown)


elasticdl/python/data/reader/odps_reader.py
```diff
@@ -59,6 +59,9 @@ def read_records(self, task):
             self._metadata.column_names = (
                 reader._odps_table.schema.names if columns is None else columns
             )
+            column_dtypes = []
+            for column_name in self._metadata.column_names:
```
brightcoder01: self._metadata.column_names = [reader._odps_table.schema[column_name] for column_name in self._metadata.column_names]

terrytangyuan (approved):

brightcoder01 (approved):

Issue #1233, opened 2019-09-24
Some workers are spinning while waiting for other workers finish tasks at the end of one epoch.


brightcoder01:
In this cluster of N workers, at the end of one epoch, M workers finished their tasks, the other N - M workers are still working. These M workers will get the task of WAIT type to wait until the whole epoch finishes. In this stage, these M workers will spinning in the loop and one cpu will be very busy.

chunyang-wen:
It is safe to keep them online but it will waste computing resources. Maybe the master should have better control of this situation.

+ Deallocate redundant resources
+ If there are still new tasks coming up due to the failure of workers, the master should start enough new workers in order to finish the tasks.

chunyang-wen:
cc @ywskycn

brightcoder01:
The worker will sleep(5) while receiving the WAIT task. Please check the following code snippet.

https://github.com/sql-machine-learning/elasticdl/blob/3b5196ff071474a2e3a2df9aaea2e5e042bb3bbe/elasticdl/python/worker/task_data_service.py#L86-L90

PR #1933, opened 2020-04-29
Workers wait 5s if receiving a WAIT task.

workingloong:
There are too many requests if the worker does not sleep.
Relate #1233


elasticdl/python/worker/task_data_service.py
```diff
@@ -205,6 +205,7 @@ def _gen(self):
                 if task.type == elasticdl_pb2.WAIT:
                     self._pending_dataset = True
                     logger.info("No tasks for now, maybe more later")
+                    time.sleep(5)
```
wangkuiyi, line 210: Need a comment here to explain why we need to wait here.

QiJune (approved):
Excellent!

PR #1935, opened 2020-04-30
Catch the exception when using k8s API to get pods.

workingloong:
There are network jitter and raise an exception when getting pods.

### DIFF
--- a/elasticdl/python/common/k8s_job_monitor.py
+++ b/elasticdl/python/common/k8s_job_monitor.py
@@ -1,5 +1,5 @@
 import time
-
+import traceback
 from elasticdl.python.common.constants import PodStatus
 from elasticdl.python.common.k8s_client import Client
 from elasticdl.python.common.log_utils import default_logger as logger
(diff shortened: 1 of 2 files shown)


elasticdl/python/common/k8s_job_monitor.py
```diff
@@ -145,38 +150,42 @@ def monitor_status(self):
...
         while True:
-            master_pod = self.client.get_master_pod()
-            if master_pod is None:
-                retry_num += 1
-                if retry_num > MAX_READ_POD_RETRIES:
-                    logger.error(
-                        "{} Not Found".format(
-                            self.client.get_master_pod_name()
+            try:
+                master_pod = self.client.get_master_pod()
```
brightcoder01, line 156: If we can get the pod object, `retry_num` need be reset to 0.

elasticdl/python/common/k8s_job_monitor.py
```diff
@@ -43,24 +44,28 @@ def monitor_status(self):
         pod_succeeded = False
 
         while True:
-            pod = self.client.get_pod(self.pod_name)
-            if pod is None:
-                retry_num += 1
-                if retry_num > MAX_READ_POD_RETRIES:
-                    logger.error("{} Not Found".format(self.pod_name))
+            try:
+                pod = self.client.get_pod(self.pod_name)
```
brightcoder01, line 49: Will `self.client.get_pod(self.pod_name)` throw exception.

elasticdl/python/common/k8s_job_monitor.py
```diff
@@ -43,24 +44,28 @@ def monitor_status(self):
...
-                break
-            else:
-                time.sleep(30)
+                elif pod.status.phase == PodStatus.FAILED:
+                    logger.info(self.client.get_pod_log(self.pod_name))
+                    break
+                else:
+                    time.sleep(30)
+            except Exception:
+                traceback.print_exc()
```
brightcoder01: Should we record the trace in the log?

elasticdl/python/common/k8s_job_monitor.py
workingloong: Yes, it will. But it will not break although it throws exception.

elasticdl/python/common/k8s_job_monitor.py
workingloong: The log is unused for users, so I will remove the code.

elasticdl/python/common/k8s_job_monitor.py
```diff
@@ -43,24 +43,29 @@ def monitor_status(self):
...
-                logger.info(self.client.get_pod_log(self.pod_name))
-                break
-            else:
-                time.sleep(30)
+                elif pod.status.phase == PodStatus.FAILED:
+                    logger.info(self.client.get_pod_log(self.pod_name))
+                    break
+                else:
+                    time.sleep(30)
+            except Exception:
```
terrytangyuan: This exception is too broad though. Perhaps you only want to capture exceptions like `client.api_client.ApiException`?

workingloong: Yes! It is `client.api_client.ApiException`.

terrytangyuan (approved):

PR #1936, opened 2020-05-06
Add a fileHandler into logger to write the logs to a file

workingloong:
Except outputing the logs to stdout, we also need to write the logs to a file. Then, the log system can collect logs from the file.

### DIFF
--- a/elasticdl/python/common/log_utils.py
+++ b/elasticdl/python/common/log_utils.py
@@ -2,29 +2,50 @@ import logging
 import sys
 import typing
 
-_DEFAULT_LOGGER = "elastic.logger"
+LOG_FILE = "/home/admin/logs/elasticdl.log"
+
+_DEFAULT_LOGGER = "elasticdl.logger"
+_STDOUT_LOGGER = "elasticdl_stdout.logger"
 
 _DEFAULT_FORMATTER = logging.Formatter(
     "[%(asctime)s] [%(levelname)s] "
     "[%(filename)s:%(lineno)d:%(funcName)s] %(message)s"
 )
 
-_ch = logging.StreamHandler(stream=sys.stderr)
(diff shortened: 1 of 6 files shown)


elasticdl/python/common/log_utils.py
```diff
@@ -2,29 +2,50 @@
 import sys
 import typing
 
-_DEFAULT_LOGGER = "elastic.logger"
+LOG_FILE = "/home/admin/logs/elasticdl.log"
```
terrytangyuan, line 5: This is only internal path. Others may not have this dir.

elasticdl/python/worker/worker.py
```diff
@@ -21,7 +21,10 @@
     scatter_embedding_vector,
     string_to_id,
 )
-from elasticdl.python.common.log_utils import get_logger
+from elasticdl.python.common.log_utils import (
+    get_logger,
+    LOG_FILE,
```
terrytangyuan, line 26: You can just use `LOG_FILE` inside `log_utils` and provide it as the default value for argument `filename`.

elasticdl/python/common/log_utils.py
```diff
@@ -2,29 +2,50 @@
...
+def get_file_hander(filename):
+    file_handler = logging.FileHandler(
+        filename, mode='a', encoding="UTF-8", delay=True
+    )
+    file_handler.setFormatter(_DEFAULT_FORMATTER)
+    return file_handler
+
+
+def get_logger(
+    name, level="INFO", filename=None, handlers=None, update=False
```
terrytangyuan, line 31: filename -> file_name

workingloong:
The file handler cannot redirect the trace stack into the file. We need to use the redirection command in Linux like ">> log.txt 2>&1". So, I close the PR.

PR #1937, opened 2020-05-07
Use /bin/bash to execute the python command

workingloong:
Sometimes, we need to redirect the pod logs to a file in the disk and we can use bash redirection
command like ">> log.txt 2>&1". So we need to use /bin/bash in the command of the pod container. Then the args of the container is ["-c", "{COMMAND}"]

### DIFF
--- a/elasticdl/python/common/args.py
+++ b/elasticdl/python/common/args.py
@@ -289,6 +289,12 @@ def add_common_params(parser):
         'by semi-colon used to debug , e.g. "param1=1; param2=2" '
         "Supported auxiliary parameters: disable_relaunch",
     )
+    parser.add_argument(
+        "--log_file_path",
+        type=str,
+        default="",
+        help="The path to save logs",
+    )
 
 
 def add_train_params(parser):
@@ -721,6 +727,7 @@ def build_arguments_from_parsed_result(args, filter_args=None):
         such as ["--foo", "3", "--bar", False]
     """
     items = vars(args).items()
+    items = filter(lambda item: item[1] != "", items)
     if filter_args:
         items = filter(lambda item: item[0] not in filter_args, items)
 
--- a/elasticdl/python/common/args.py
+++ b/elasticdl/python/common/args.py
@@ -290,7 +290,10 @@ def add_common_params(parser):
         "Supported auxiliary parameters: disable_relaunch",
     )
     parser.add_argument(
-        "--log_file_path", type=str, default="", help="The path to save logs",
+        "--log_file_path",
+        type=str,
+        default="",
+        help="The path to save logs (e.g. stdout, stderr)",
     )
 
 
--- a/elasticdl/python/common/args.py
+++ b/elasticdl/python/common/args.py
@@ -290,10 +290,7 @@ def add_common_params(parser):
         "Supported auxiliary parameters: disable_relaunch",
     )
     parser.add_argument(
-        "--log_file_path",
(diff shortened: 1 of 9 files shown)


elasticdl/python/master/master.py
```diff
@@ -462,15 +475,15 @@ def _create_instance_manager(self, args):
                 self.task_d,
                 job_name=args.job_name,
                 image_name=args.worker_image,
-                worker_command=worker_command,
+                worker_command=pod_command,
```
terrytangyuan: pod_command -> container_command

elasticdl/python/master/master.py
```diff
@@ -449,6 +453,15 @@ def _create_instance_manager(self, args):
                     "--num_minibatches_per_task",
                     str(args.num_minibatches_per_task),
                 ]
+                ps_args = wrap_python_args_with_string(ps_args)
+                ps_args.insert(0, ps_client_command)
+
+            if args.log_file_path:
+                worker_args.append(">> {} 2>&1".format(args.log_file_path))
+                ps_args.append(">> {} 2>&1".format(args.log_file_path))
```
terrytangyuan: Refactor string to const

elasticdl/python/elasticdl/api.py
```diff
@@ -182,6 +179,16 @@ def _submit_job(image_name, client_args, container_args):
         force_use_kube_config_file=client_args.force_use_kube_config_file,
     )
 
+    container_args = wrap_python_args_with_string(container_args)
+
+    master_client_command = "python -m elasticdl.python.master.main"
```
terrytangyuan, line 184: This appears in multiple places as well

elasticdl/python/common/args.py
```diff
@@ -734,3 +737,38 @@ def _str_ignore_none(s):
         "--" + k if i % 2 == 0 else k for i, k in enumerate(arguments)
     ]
     return arguments
+
+
+def wrap_python_args_with_string(args):
```
terrytangyuan, line 745: Better to add simple test cases

elasticdl/python/common/args.py
```diff
@@ -289,6 +289,9 @@ def add_common_params(parser):
         'by semi-colon used to debug , e.g. "param1=1; param2=2" '
         "Supported auxiliary parameters: disable_relaunch",
     )
+    parser.add_argument(
+        "--log_file_path", type=str, default="", help="The path to save logs",
```
terrytangyuan: What type of logs? e.g. stdout, stderr

elasticdl/python/elasticdl/api.py
workingloong: It only appears here. Others are `python -m elasticdl.python.ps.main` and `python -m elasticdl.python.worker.main` which are responsible to start the ps process and the worker process.

terrytangyuan: You are correct. I was looking at the old code in the diff where this appears in multiple places.

elasticdl/python/common/args.py
workingloong: Yeah!

elasticdl/python/common/args.py
workingloong: The stdout and stderr logs.  I have added the comment in the help.

terrytangyuan (approved):

PR #1938, opened 2020-05-08
Add log_loss_steps to control the loss log frequency.

workingloong:
The log count is too much if the worker logs the loss for each batch.  We can refer to [tf.estimator.RunConfig](https://www.tensorflow.org/api_docs/python/tf/estimator/RunConfig) to control the frequency.

### DIFF
--- a/elasticdl/python/worker/worker.py
+++ b/elasticdl/python/worker/worker.py
@@ -130,6 +130,7 @@ class Worker(object):
         self._worker_id = args.worker_id
         self._job_type = args.job_type
         self._minibatch_size = args.minibatch_size
+        self._log_loss_steps = args.log_loss_steps
         (
             model_inst,
             self._dataset_fn,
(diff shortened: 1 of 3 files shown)


elasticdl/python/worker/worker.py
```diff
@@ -876,8 +877,14 @@ def _process_minibatch(
                 *accepted, min_model_version, loss = self._run_training_task(
                     features, labels
                 )
+                if min_model_version % self._log_loss_steps == 0:
```
skydoorkai: ` min_model_version` is the max model version from ps. So sometimes it will skip some steps. For example, if log_loss_steps=100, we may not get ` min_model_version==100`.

workingloong: I have fixed that each worker print loss if its `model_version` > `log_loss_steps * log_loss_count`. `log_loss_count` add 1 if the worker printes loss.

skydoorkai: If this worker is relaunched in the middle of the training, or a checkpoint is loaded, `self._log_loss_count` is 0, and `min_model_version` is very large. The log will be printed every time.
Need to set `self._log_loss_count` accordingly, such as `int(min_model_version/self._log_loss_steps)`

workingloong: Yeah!

skydoorkai (approved):

PR #1945, opened 2020-05-12
Create env name template for analysis result

workingloong:


elasticdl_preprocessing/utils/analyzer_utils.py
```diff
@@ -5,6 +5,16 @@
 import os
 
 
+class AnalysisEnvTemplate:
```
terrytangyuan: Put this in a separate file? Also is this only used internally? If so, you can start the name with underscore.

workingloong: Yes, it is better to put it into a separate file.

terrytangyuan (approved):

PR #1949, opened 2020-05-25
Raise a ValueError when the divisor is 0.

workingloong:
The output will be nan if the divisor is 0, so we should raise an error if the divisor is 0.


skydoorkai (approved):
LGTM

brightcoder01 (approved):

PR #1952, opened 2020-05-27
Cast value to float test before operating normalization 

workingloong:
If the `x.type` is tf.int32 or tf.int64, the `tf.cast` will convert the float to int. It will make loss of precision.


skydoorkai (approved):
gltm

PR #1955, opened 2020-05-30
Convert DNN outputs to a logit

workingloong:
We should convert DNN outputs to a logit like 
https://github.com/shenweichen/DeepCTR/blob/9e6be3f28d5b213e8596b8f79e68efd6d5fd21ba/deepctr/models/wdl.py#L50-L53


QiJune (approved):
Excellent！

PR #1956, opened 2020-06-03
ConcatenateWithOffset return inputs directly if the inputs is not a list

workingloong:
`ConcatenateWithOffset` support a single tensor or a list with only one tensor. If it extends `tf.keras.layers.Concatenate`, the method `build` of `tf.keras.layers.Concatenate` will check whether the input shape > 2


skydoorkai (approved):
LGTM

PR #1984, opened 2020-06-10
pack go ps binary to elasticdl python wheel

QiJune:
This PR packs `elasticdl_ps` binary into the elasticdl python wheel package.

Apart from installing elasticdl python package, we have to add `/usr/local/lib/python3.6/dist-packages/elasticdl/go/bin` to system path.


elasticdl/docker/Dockerfile
```diff
@@ -60,6 +60,7 @@ RUN cd /elasticdl && \
     go install ./...
 
 # Create elasticdl package
+RUN cp /root/go/bin/main /elasticdl/elasticdl/pkg/ps_main
```
workingloong: How can we generate the `/root/go/bin/main`?

QiJune: It's built by `go install ./...`
The `main` binary is installed at `$GOPATH/bin` directory

wangkuiyi (approved):

PR #1989, opened 2020-06-11
refine go ps directory

QiJune:
Add a cmd directory, named `start_ps`.  As a result, the Go PS binary name is also changed to `start_ps`.


elasticdl/go/cmd/start_ps/main.go
```diff
@@ -1,13 +1,27 @@
...
 
 import (
-	"elasticdl.org/elasticdl/common"
-	"elasticdl.org/elasticdl/ps"
 	"flag"
 	"fmt"
 	"log"
 	"os"
 	"time"
+
```
workingloong, line 22: Is the empty line needed?

elasticdl/go/pkg/common/tensor.go
```diff
@@ -1,13 +1,27 @@
...
+
 package common
 
 import (
-	"elasticdl.org/elasticdl/proto"
 	"fmt"
+	"reflect"
+	"unsafe"
+
+	"elasticdl.org/elasticdl/pkg/proto"
```
workingloong, line 21: "elasticdl.org/elasticdl/go/pkg/proto"?

elasticdl/go/cmd/start_ps/main.go
QiJune: It's formated by visual code automatically.

elasticdl/go/pkg/common/tensor.go
QiJune: Here is the reference for go project structure, https://github.com/golang-standards/project-layout.  A project has pkg and cmd.

We build the go project by a Dockerfile of elasticdl dev.
```
cp -r /elasticdl/elasticdl/go/ ${ELASTICDLPATH}
cd $ELASTICDLPATH
go mod init elasticdl.org/elasticdl
go mod tidy
go install ./...
```

The `go` directory is copied and renamed to a new directory, and we use go module to init a new module in this new directory, named `elasticdl.org/elasticdl`

So the import path is `elasticdl.org/elasticdl/pkg/proto`

workingloong:
LGTM

workingloong (approved):
LGTM

wangkuiyi:
> Add a cmd directory, named `start_ps`. As a result, the Go PS binary name is also changed to `start_ps`.

How about the name `elasticdl_ps`, or simply `ps`? @QiJune

PR #1997, opened 2020-06-12
refine elasticdl:ci dockerfile for integration test

QiJune:
We need to make running integration test in Travis is the same as running in local.

We need a Dockerfile like this:

```
FROM tensorflow
RUN pip install elasticdl 
COPY model_zoo /model_zoo
```

This PR will do the first step, we make a Dockerfile for ci like following:

```
FROM tensorflow

COPY --from=dev /elasticdl-develop-py3-none-any.whl /
RUN python -m pip -q install /elasticdl-develop-py3-none-any.whl \
        --extra-index-url=${EXTRA_PYPI_INDEX} \
    && rm /elasticdl-develop-py3-none-any.whl

// to be simplified in the next step
COPY --from=dev ${GO_PS} /usr/local/go/bin

COPY model_zoo /model_zoo

// to be simplified in the next step
COPY --from=dev /data /data
```

So, in the integration test, we could directly pass "elasticdl:ci" to `image_name`  to launch a training job.

### DIFF
--- a/elasticdl/docker/build_all.sh
+++ b/elasticdl/docker/build_all.sh
@@ -18,13 +18,14 @@ TF_VERSION=2.1.0
 
 if [[ ! -d .git ]]; then
     echo "We must run this script at the root of the source tree."
+    exit 1
 fi
 
 if [[ $# -eq 1 && $1 == "-gpu" ]]; then
     base_image="tensorflow/tensorflow:${TF_VERSION}-gpu-py3"
     echo "To support CUDA; all images are from " $base_image
-    dev_image="elasticdl:gpudev"
-    ci_image="elasticdl:gpuci"
+    dev_image="elasticdl:dev-gpu"
+    ci_image="elasticdl:ci-gpu"
 
 else
     base_image="tensorflow/tensorflow:${TF_VERSION}-py3"
--- a/elasticdl/docker/build_all.sh
+++ b/elasticdl/docker/build_all.sh
@@ -1,27 +1,36 @@
 #!/bin/bash
+# Copyright 2020 The SQLFlow Authors. All rights reserved.
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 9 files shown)


elasticdl/docker/build_all.sh
```diff
@@ -1,28 +1,37 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
 
 TF_VERSION=2.1.0
 
 if [[ ! -d .git ]]; then
     echo "We must run this script at the root of the source tree."
-    exit -1
```
wangkuiyi, line 7: It seems that we cannot remove the `exit` command?

elasticdl/docker/build_all.sh
```diff
@@ -1,28 +1,37 @@
...
 fi
 
 if [[ $# -eq 1 && $1 == "-gpu" ]]; then
     base_image="tensorflow/tensorflow:${TF_VERSION}-gpu-py3"
     echo "To support CUDA; all images are from " $base_image
-    image="elasticdl:gpu"
     dev_image="elasticdl:gpudev"
```
wangkuiyi: elasticdl:dev-gpu
elasticdl:ci-gpu

elasticdl/python/common/save_utils.py
```diff
@@ -1,3 +1,16 @@
+# Copyright 2020 The SQLFlow Authors. All rights reserved.
```
wangkuiyi, line 1: It seems that in the next PR, we should remove the filename from `save_utils.py` into `checkpoint.py`.

elasticdl/docker/build_all.sh
QiJune: Yes, fix it. 

Add `exit 1`

elasticdl/docker/build_all.sh
QiJune: Done.

elasticdl/python/common/save_utils.py
QiJune: Exactly, `checkpoint.py` is more meaningful than `save_utils.py`

wangkuiyi (approved):

PR #2005, opened 2020-06-12
WIP Reduce stages in .travis.yml

wangkuiyi:
Since each of the two stages has only one job, and each job builds the Docker image, I am trying to merge them into one job.

Also, following https://ops.tips/blog/update-travis-ci-docker/, I am trying to upgrade Docker in Travis VMs so we can enable `DOCKER_BUILDKIT`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: minimal
 
-dist: bionic
+dist: focal
 
 addons:
   apt:
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,8 +27,8 @@ jobs:
     - stage: tests
       script:
         - set -e
-        - apt-get update -y
-        - apt install --only-upgrade docker-ce -y
+        - sudo apt-get update -y
+        - sudo apt-get install --only-upgrade docker-ce -y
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -1,16 +1,17 @@
...
   apt:
     packages:
-      - docker-ce
       - python3-pip
       - python3-setuptools
       - clang-format
 
 install:
+  # Install the most recent version of Docker to use DOCKER_BUILDKIT.
+  - bash scripts/travis/install-docker.sh
```
terrytangyuan, line 14: What's the benefit of this change compared to the original approach?

wangkuiyi: I am trying to install a more recent version of Docker so we can use  DOCKER_BUILDKIT.

wangkuiyi:
it seems that we can install the most recent version of Docker in Travis CI VMs, but it doesn't work with resolving the base image when we run `docker build`.

```
#3 [internal] load metadata for docker.io/tensorflow/tensorflow:2.1.1
#3 ERROR: docker.io/tensorflow/tensorflow:2.1.1 not found
```

PR #2011, opened 2020-06-13
[Discussion Only] Enumerate all the options of organizing directory structures

brightcoder01:
## Principles
- Baseline: Don't bring conflict or potential bug.
- Make the directory clean and simple. Avoid making it too deep.
- Make all the packages decoupled.

## Comparison Detail
- elasticdl0: Cannot work. `Server` and `Client` cannot include the modules in common;
- elasticdl1: Potential conflict. If user pip install `elasticdl1_client` and `elasticdl1_server` on the same machine. The `site-packages/elasticdl1_common` folder will be overrided by the later `pip install`. It will bright potential bug.
- elasticdl2: It works. But the directory structure is deep. For example. the path of `client.py` is `elasticdl2/elasticdl2_client/elasticdl2_client/python/client.py`.
- elasticdl3: It works. And the directory structure is simpler than `elasticdl3`. The path of `client.py` is `elasticdl2/elasticdl2_client/python/client.py`. 
- elasticdl4: It's similar with `elasticdl3`. The difference is that we put the common modules into elasticdl4_client.

Summary:
- If there are lots of common modules, it's worthy to make a separate package for elasticdl_common. `elasticdl3` is the choice.
- If there are not so many common modules, `elasticdl4` is the choice. => **It's the current status of our repo. We will choose this solution.**


skydoorkai:
`elasticdl4` looks better.

brightcoder01:
Close this issue. We will choose the solution `elasticdl4` after discussion.

PR #2015, opened 2020-06-13
Pull base image quietly

wangkuiyi:
- Move the base image into `scripts/travis/build_images.sh` from `Dockerfile`, so we have a chance to pull the base image quietly in CI.

- The moving keeps the fact that there is only one place where the base image is specified -- preventing from code duplication.


terrytangyuan (dismissed):

QiJune (approved):
Excellent!

PR #2017, opened 2020-06-13
Merge Dockerfile.dev-allreduce into Dockerfile

wangkuiyi:
`Dockerfile.dev_allreduce` uses a stage in `Dockerfile` as the base image. It makes sense to merge them into one file.


QiJune (approved):
Excellent!

PR #2021, opened 2020-06-14
Create a script to launch an ElasticDL job to run integration tests on minikube.

workingloong:
Move shell commands to launch an ElasticDL job on minikube to a file. So, the developer can use the file to launch a job on their machines.


scripts/travis/start_elasticdl_job.sh
```diff
@@ -0,0 +1,56 @@
+JOB_TYPE=$1
+
+kubectl apply -f elasticdl/manifests/elasticdl-rbac.yaml
```
wangkuiyi: I'd thought that this line should be right after the starting of the minikube cluster.  We don't need to run it multiple times.

scripts/travis/start_elasticdl_job.sh
```diff
@@ -0,0 +1,56 @@
...
+        -e MAXCOMPUTE_AK=$ODPS_ACCESS_ID \
+        -e MAXCOMPUTE_SK=$ODPS_ACCESS_KEY \
+        -v /var/run/docker.sock:/var/run/docker.sock \
+        -v $HOME/.kube:/root/.kube \
+        -v /home/$USER/.minikube/:/home/$USER/.minikube/ \
+        -v $(pwd):/work \
+        -w /work elasticdl:ci \
+        bash -c "scripts/client_test.sh \
+        ${JOB_TYPE} ${PS_NUM} ${WORKER_NUM}"
+    if [[ "$JOB_TYPE" != "local" ]]; then
```
wangkuiyi, line 37: Do we still need type "local"? I'd thought the local mode was developed at the very beginning stage of prototyping and will no longer be used.  If so, how about we remove this test and even the local mode source code in the next PR?

scripts/travis/start_elasticdl_job.sh
```diff
@@ -0,0 +1,56 @@
...
+        python3 scripts/validate_job_status.py \
+        ${JOB_TYPE} ${PS_NUM} ${WORKER_NUM}
+    fi
+    if [[ "$JOB_TYPE" == "odps" ]]; then
```
wangkuiyi, line 41: In the next PR, we should move this code snippet into a new script `scripts/travis/cleanup_odps_table.sh`.

scripts/travis/start_elasticdl_job.sh
```diff
@@ -0,0 +1,56 @@
...
+
+if [[ "$JOB_TYPE" == "odps" ]] && \
+{ [[ "$ODPS_ACCESS_ID" == "" ]] || \
+[[ "$ODPS_ACCESS_KEY" == "" ]]; }; then
+    echo "Skipping ODPS related integration tests since \
+    either ODPS_ACCESS_ID or ODPS_ACCESS_KEY is not set"
+else
+    echo "Running ElasticDL job: ${JOB_TYPE}"
+    if [[ "$JOB_TYPE" == "odps" ]]; then
+        export MAXCOMPUTE_TABLE="odps_integration_build_"`
```
wangkuiyi, line 11: In the next PR, it seems that we should move this code snippet into a new shell script file `scripts/travis/create_odps_table.sh`.

scripts/travis/start_elasticdl_job.sh
```diff
@@ -0,0 +1,56 @@
+JOB_TYPE=$1
```
wangkuiyi, line 1: The word "start" means to start the job but not waiting for the completion. However, this script does wait for the completion of the job.  So it should be renamed `run_job.sh`.

wangkuiyi:
After the merging of this PR, we should consider the emerging of `run_elasticdl_job.sh` and [`scripts/client_test.sh`](https://github.com/sql-machine-learning/elasticdl/blob/develop/scripts/client_test.sh).

scripts/travis/start_elasticdl_job.sh
workingloong: Yeah!

scripts/travis/start_elasticdl_job.sh
workingloong: We developed "local" to test the training loop using the `dataset_fn` defined in the model file. But it is not user-friendly because users should execute it using ElasticDL package and an ElasticDL client command. There may be a better solution to do it. If users can use the `dataset_fn` to get a `tf.data.Dataset`,  they can use `model.fit(dataset)` to test the training loop like a Kera-native program locally.

So, I will remove it and source codes in the next PR. Then I will design the proposed solution to get a `tf.dataDataset` using `dataset_fn` defined in the model file.

wangkuiyi (dismissed):

workingloong:
The PR cannot trigger the Travis CI, so I close it.

workingloong:
> After the merging of this PR, we should consider the emerging of `run_elasticdl_job.sh` and [`scripts/client_test.sh`](https://github.com/sql-machine-learning/elasticdl/blob/develop/scripts/client_test.sh).

Yes. We should write a tutorial for developers to run unit and integration tests using those script files.

PR #2023, opened 2020-06-15
mount elasticdl source file when running unit tests and building elasticdl Python wheel package

### LINKED ISSUE #2020: Don't copy source code into Docker images at build time
Instead, we should bind mount source tree to containers using `docker run -v $TRAVIS_BUILD_DIR:/work -w /work`, so we write the output `.whl` files back to the host via `/work/buiild/*.whl` as in https://github.com/sql-machine-learning/elasticdl/issues/1982.

QiJune:
Fix #2020 

First, we build `elasticdl:dev` image.

Second, we run`build_and_test.sh` with elasticdl source file mount in the container.

```
docker run --rm -v $PWD:/elasticdl -w /elasticdl elasticdl:dev bash -c "K8S_TESTS=False scripts/build_and_test.sh"
```

It will run go unit tests and python tests, then it generates the elasticdl Python wheel package in the build directory.

Third, we copy elasticdl Python wheel package in `elasticdl:ci` Dockerfile, and run integration tests.


wangkuiyi (approved):
Great work!

scripts/build_and_test.sh
```diff
@@ -0,0 +1,46 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -e
+
+# Build elasticdl.org/elasticdl Go package
+go env -w GOPROXY=https://goproxy.cn,https://developer.aliyun.com/mirror/goproxy,direct
+cd /elasticdl && \
```
wangkuiyi, line 19: This line assumes that 

1. the source tree is bound mount into the dev container, and
1. to `/elasticdl` in the container.

These assumptions prevent this script from being used in local development.

Let us try to fix this in a future PR.

scripts/build_and_test.sh
```diff
@@ -0,0 +1,46 @@
...
+make -f elasticdl/Makefile && \
+cp -r /elasticdl/elasticdl/go/ /root/elasticdl && \
```
wangkuiyi, line 21: The directory `/root/elasticdl` doesn't exist on the host.  This line prevents this script from being used in on-host development. Let us see if we can fix it in a future PR.

PR #2028, opened 2020-06-15
Add markdownlint in pre-commit

### LINKED ISSUE #2000: Add markdownlint in pre-commit and only check markdown files in the pull request during CI.
Using markdownlint to check the existed markdown files, quite a few of md files need be updated.
To avoid the large PR, we only check the markdown files in the pull request.

brightcoder01:
fix #2000


scripts/pre-commit.sh
```diff
@@ -26,4 +26,6 @@ GO_SOURCE_FILES=$(find elasticdl/pkg -name '*.go' -print0 | tr '\0' ' ')
 
 YAML_SOURCE_FILES=$(find .travis.yml -name '*.yml' -print0 | tr '\0' ' ')
 
-pre-commit run --files "$PYTHON_SOURCE_FILES $GO_SOURCE_FILES"
+MARKDOWN_SOURCE_FILES=$(find elasticdl -name '*.md' -print0 | tr '\0' ' ')
```
wangkuiyi: This skips over the /docs directory. Not right.

If you are afraid of the messy output from currently non-compliant Markdown files, the trick you need is

https://github.com/sql-machine-learning/sqlflow/blob/f566a9362f52d1ce66d3354526f4e118db1f3564/scripts/pre-commit/shellcheck.sh#L15

brightcoder01:
At the present, pre-commit.sh execute the following command.
```BASH
pre-commit run --files "$PYTHON_SOURCE_FILES $GO_SOURCE_FILES"
```
The command will ignore all the files.

**Our target is to check the changed files in this pull request.** Just tried the following command from the [doc](https://pre-commit.com/#usage-in-continuous-integration).
```BASH
pre-commit run --from-ref origin.HEAD --to-ref HEAD
```

we will get the following error message:
```
An unexpected error has occurred: CalledProcessError: command: ('/usr/bin/git', 'diff', '--name-only', '--no-ext-diff', '-z', 'origin.HEAD...HEAD')
return code: 128
expected return code: 0
stdout: (none)
stderr:
    fatal: ambiguous argument 'origin.HEAD...HEAD': unknown revision or path not in the working tree.
    Use '--' to separate paths from revisions, like this:
```
    'git <command> [<revision>...] -- [<file>...]'

brightcoder01:
Use `CHANGED_FILES=$(git diff --cached --name-only --diff-filter=ACMR)` will get a empty list of changed files in Travis CI.

scripts/pre-commit.sh
brightcoder01: In order to run pre-commit only on the changed files in this pull request, update the pre-commit.sh as the following, and it works.

```BASH
PROJECT_URL=https://github.com/sql-machine-learning/elasticdl.git
REMOTE=$(git remote -v | grep $PROJECT_URL | cut -f1 | head -1)
pre-commit run --from-ref "$REMOTE/develop" --to-ref HEAD
```

brightcoder01: After we update the format of all the markdown files, we can update the pre-commit command to check all the files in this repo.

.pre-commit-config.yaml
```diff
@@ -60,3 +60,11 @@ repos:
     rev: v1.23.0
     hooks:
       - id: yamllint
+  - repo: https://github.com/igorshubovych/markdownlint-cli
```
workingloong, line 63: Why do you not use the repo https://github.com/markdownlint/markdownlint? It is more popular than `igorshubovych/markdownlint-cli` with more stars.

brightcoder01: I have tried `https://github.com/markdownlint/markdownlint` but it requires Ruby. It's not worthy to install Ruby in our dev image just for pre-commit. So I choose markdownlint-cli.

wangkuiyi (approved):

PR #2031, opened 2020-06-15
Remove the integration test for "local"

workingloong:
We developed "local" to test the training loop using the dataset_fn defined in the model file. But it is not user-friendly because users should execute it using ElasticDL package and an ElasticDL client command. There may be a better solution to do it. If users can use the `dataset_fn` to get a `tf.data.Dataset`, they can use `model.fit(dataset)` to test the training loop like a Kera-native program locally.

I will remove the source codes about local execution in the next PR. Then I will design the proposed solution to get a tf.dataDataset using dataset_fn defined in the model file.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -70,7 +70,6 @@ jobs:
               train
               evaluate
               predict
-              local
           )
           for JOB_TYPE in "${JOB_TYPES[@]}"; do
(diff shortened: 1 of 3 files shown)


wangkuiyi (approved):

skydoorkai (approved):
Agree that if the purpose of `local` run is for users to develop models in the local environment, we can use Keras directly to make it more simple.

PR #2039, opened 2020-06-16
Install shellcheck in Dev image.

brightcoder01:
Currently shellcheck binary is missing in elasticdl:dev. It will cause pre-commit run with shellcheck hook fails.


skydoorkai (approved):
lgtm

PR #2040, opened 2020-06-16
fix ci

### LINKED ISSUE #2038: ODPS integration test failed but CI still returns success
Here is a log for reference: https://travis-ci.com/github/sql-machine-learning/elasticdl/builds/171459744

QiJune:
Fix #2038 

The `elasticdl.proto` imports `tensor.proto` of TensorFlow repo. As a result, when we build our PS Go package, it will pull the TensorFlow Go package. We do not want to import the large TensorFlow Go package.

So, we create a local simple Go package.

```bash
mkdir -p ${GOPATH}/pkg/mod
protoc -I${TF_PATH} ${TF_PATH}/tensorflow/core/framework/tensor.proto --go_out=${GOPATH}/pkg/mod
cd "$GOPATH"/pkg/mod/github.com/tensorflow
go mod init github.com/tensorflow
```

Then, our Go PS package depends on the local simple go module,  We replace the go.mod of elasticdl PS, to avoid pulling the large TensorFlow Go package.

```
go mod edit -replace github.com/tensorflow="${GOPATH}"/pkg/mod/github.com/tensorflow
```

Considering the complexity, we build the Go package inside Docker totally. Cache Go in Travis will be done in the next PR when we find an elegant solution.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,7 @@ cache:
     - $HOME/.keras/datasets
     # https://pre-commit.com/#travis-ci-example
     - $HOME/.cache
-    # Cache $hOME/go, which bound mount to /root/go of dev container.
+    # Cache $HOME/go, which bound mount to /root/go of dev container.
     - $HOME/go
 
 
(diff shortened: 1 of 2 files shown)


skydoorkai (approved):
lgtm

.travis.yml
```diff
@@ -47,12 +47,12 @@ jobs:
         # Run a minikube cluster
         - bash scripts/travis/start_minikube.sh
         # Run unit tests not related to ODPS
+        # TODO(qijun) cache Go
         - docker run --rm -it --net=host
           -v $HOME/.cache:/root/.cache
           -v $HOME/.keras/datasets:/root/.keras/datasets
           -v $HOME/.kube:/root/.kube
           -v $HOME/.minikube:/home/$USER/.minikube
-          -v $HOME/go:/root/go
```
wangkuiyi, line 55: Should we still try to cache the go get results?

PR #2044, opened 2020-06-17
Move elasticdl_preprocessing folder out of the elasticdl main package in the wheel deployment.

brightcoder01:
Currently we are building two pip packages from elasticdl repo: `elasticdl` and `elasticdl_preprocessing`. While we pip install elasticdl, it also extracts an `elasticdl_preprocessing` folder. It will overwrite the files from `elasticdl_preprocessing` package.

In order to fix this, we don't include `elasticdl_preprocessing` folder in elasticdl package.

### DIFF
--- a/setup_preprocessing.py
+++ b/setup_preprocessing.py
@@ -34,6 +34,8 @@ setup(
     install_requires=required_deps,
     extras_require=extras,
     python_requires=">=3.5",
-    packages=find_packages(include=["*elasticdl_preprocessing*"],),
+    packages=find_packages(
+        include=["elasticdl_preprocessing*"], exclude=["*test*"]
+    ),
     package_data={"": ["requirements.txt"]},
 )
--- a/elasticdl/docker/Dockerfile.ci
+++ b/elasticdl/docker/Dockerfile.ci
@@ -33,6 +33,12 @@ COPY model_zoo /model_zoo
 RUN python -m pip install --quiet -r /model_zoo/requirements.txt \
         --extra-index-url=${EXTRA_PYPI_INDEX}
 
+# Install elasticdl_preprocessing package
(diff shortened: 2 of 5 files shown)


setup_preprocessing.py
```diff
@@ -34,6 +34,8 @@
     install_requires=required_deps,
     extras_require=extras,
     python_requires=">=3.5",
-    packages=find_packages(include=["*elasticdl_preprocessing*"],),
+    packages=find_packages(
+        include=["elasticdl_preprocessing*"], exclude=["*test*"]
```
workingloong, line 38: Why do we need to set `exclude` when we have set `include` ?

setup_preprocessing.py
```diff
@@ -34,6 +34,8 @@
...
+    ),
     package_data={"": ["requirements.txt"]},
```
workingloong, line 40: elasticdl_preprocessing/requirements.txt

setup_preprocessing.py
brightcoder01: Because elasticdl_preprocessing also contains test package.

setup_preprocessing.py
brightcoder01: After verification, if update to `elasticdl_preprocessing/requirements.txt`, this file won't be packaged in the wheel.

workingloong: IMO, we should package the `requirements.txt` into the wheel.

workingloong (approved):

setup_preprocessing.py
brightcoder01: > IMO, we should package the `requirements.txt` into the wheel.

Yes, in the current implementation, it's packaged.

wangkuiyi (approved):
LGTM!

Issue #1918, opened 2020-04-20
Decouple elasticdl-client from elasticdl main package.


brightcoder01:
ElasticDL main package has the requirement of tensorflow 2.x, but elasticdl client only depends on kubernetes api.
We can decouple elasticdl-client from elasticdl main package.

brightcoder01:
The target is: user don't need `kubectl` to manage ElasticDL jobs. What users need is only `elasticdl-client`. 

Our client need cover the following features: 

- Build image: Make `elasticdl:latest` or `elasticdl:run` a base image. User build the image for ElasticDL training with model definition and dependencies using elasticdl-client.
- Submit job: Submit elasticdl jobs to minikube/Google Cloud/ASI using elasticdl-client. We can use elasticdl-client in our MacBook / Container in a Kubernetes cluster.
- List ElasticDL job detail: `elasticdl get job` can print the details of a specific elasticdl training/prediction/evaluation jobs.
- Monitor job / Wait for job done: `elasticdl monitor <job-name>`, exit until the job success or fails. Print some key progress logs on the console.

brightcoder01:
Implement ElasticDL Client using Python or Go?
Make verification using sample code.

brightcoder01:
### User Story
1. User develops model in local IDE or web IDE. The directory structure is as follows:

```
my_model_repo
- layers
- models
- utils
requirements.txt
```

2. `cd ${model_repo_root_path}` && `elasticdl_cli zoo init`, generate a Dockerfile just like follows:

```Dockerfile
FROM elastcdl:latest
COPY . /models
RUN pip install -r /models/requirements.txt
```

3. `elasticdl_cli zoo build --image_name=reg.docker.alibaba-inc.com/bright/elasticdl-wnd:1.0`: Execute docker build using the Dockerfile generated from step 2;
4. `elasticdl_cli zoo push --image_name=reg.docker.alibaba-inc.com/bright/elasticdl-wnd:1.0`: Push the docker image into the remote repository;
5. `elasticdl_cli train/predict/evaluate --image_name=reg.docker.alibaba-inc.com/bright/elasticdl-dien:1.0 --model_def=models.my_model_repo.models.wnd.custom_model ...`: Submit a model training/prediction/evalution job.

brightcoder01:
The issue is resolved by the associated PR.

PR #2082, opened 2020-06-28
ElasticDL Client: Implement "zoo init | build | push" command.

brightcoder01:
Relate #1918

### DIFF
--- a/elasticdl_client/api.py
+++ b/elasticdl_client/api.py
@@ -11,12 +11,24 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import os
+
 import docker
 from jinja2 import Template
 
 
 def zoo_init(args):
     print("Create the Dockerfile for the model zoo.")
+
+    cluster_spec_path = args.cluster_spec
+    cluster_spec_name = None
+    if cluster_spec_path:
+        if not os.path.exists(cluster_spec_path):
+            raise RuntimeError(
+                "The cluster spec {} doesn't exist".format(cluster_spec_path)
+            )
+        cluster_spec_name = os.path.dirname(cluster_spec_path)
+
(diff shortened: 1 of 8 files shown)


elasticdl_client/api.py
```diff
@@ -0,0 +1,116 @@
...
+{% endif %}
+"""
+    template = Template(tmpl_str)
+    docker_file_content = template.render(
+        BASE_IMAGE=args.base_image,
+        EXTRA_PYPI_INDEX=args.extra_pypi_index,
+        CLUSTER_SPEC_NAME=cluster_spec_name,
+    )
+
+    with open("./Dockerfile", mode="w+") as f:
```
skydoorkai, line 62: nitpicking: `mode="w" ` is enough.

elasticdl_client/common/args.py
```diff
@@ -0,0 +1,78 @@
...
+        type=str,
+        help="The file that contains user-defined cluster specification,"
+        "the file path can be accessed by ElasticDL client.",
+        default="",
+    )
+
+
+def add_zoo_build_arguments(parser):
+    parser.add_argument(
+        "path", type=str, help="The path where the build context locate."
```
skydoorkai, line 39: "The path where the build context locate." -> 
The path where the build context locates.

wangkuiyi (approved):
LGTM.

We could fix comments in a future PR.

elasticdl_client/common/args.py
brightcoder01: Will resolve in the next PR.

elasticdl_client/api.py
brightcoder01: Will resolve in the next PR.

PR #2089, opened 2020-06-29
Add the implementation of train | evaluate | predict command in elasticdl_client.

brightcoder01:
- Duplicate the necessary modules & methods from main package to elasticdl_client package for train | evaluate | predict command.
- In the original implementation of train | evaluate | predict in main package, we will build a new image for the job running. In the new implementation, we will submit the job using the image already built using `elasticdl zoo build` command instead of building a new one.
- In the next PR, we will remove the modules from main package and import from elasticdl_client package directly.


QiJune (approved):
Excellent!

PR #2092, opened 2020-06-30
Use the original argument name image_name instead of image for train | predict | evaluate.

brightcoder01:
Relate #1918 

Currently the argument parser in main package for master, ps, worker still use image_name. To make sure that the function won't be broken, we keep use the original argument name.

We can rename the argument name after we remove the duplicate modules in main package and import the module in elasticdl_client.


workingloong (approved):
LGTM

PR #2099, opened 2020-07-01
Rename the console entry of elasticdl_client package to elasticdl

brightcoder01:


wangkuiyi (dismissed):
Does this mean that we can use `elastcidl` as the client command-line tool name after the merge of this PR?

skydoorkai:
> Does this mean that we can use `elastcidl` as the client command-line tool name after the merge of this PR?

Yes.

skydoorkai (approved):

PR #2101, opened 2020-07-01
Enable go ps by default.

brightcoder01:


wangkuiyi (approved):

elasticdl_client/common/args.py
```diff
@@ -337,7 +337,7 @@ def add_common_params(parser):
     add_bool_param(
         parser=parser,
         name="--use_go_ps",
-        default=False,
+        default=True,
```
wangkuiyi, line 340: Do we have a CI test that can cover this change?

brightcoder01: Yes, the integration test will use the default setting for use_go_ps. So it will cover this change.
Maybe we can add a test to cover the python ps specifically.

PR #2107, opened 2020-07-02
Master does not delete and relaunch OOM workers.

### LINKED ISSUE #2043: Master occasionally does not clear and relaunch the OOM pod.
Now，master will clear the OOM worker pod and relaunch a new worker pod. However, we can see that the master occasionally does not clear and relaunch the OOM pod.

image

There are two problems:
1. Should master relaunch the OOM worker pod? 
Master will relaunch the OOM worker pod frequently if the configured memory for the worker is not enough. Then, the training is very slow and the job cannot exit. So, the user does not know why it is very slow. The proposed solution is that master does not relaunch the OOM worker pod and the job exits with an OOM error if all workers exit with OOM. Then, the user can restart the job with more memory for workers.

1. We may need a robust solution to relaunch pods.
Except for watching events, master can check the worker status periodically and relaunch the pod.

workingloong:
Fix #2043


elasticdl/python/master/master.py
```diff
@@ -32,6 +32,7 @@
     DistributionStrategy,
     InstanceManagerStatus,
     JobType,
+    PodStatus,
```
skydoorkai: Not needed?

workingloong: Yeah

skydoorkai (approved):

PR #2125, opened 2020-07-03
Fix SameFileError exception.

brightcoder01:
- Fix the SameFileError exception while using shuilts.copy2
- rename `add_zoo_init_arguments` to `add_zoo_init_params`, make the naming convention to be aligned in api.py


skydoorkai (approved):
lgtm

PR #2129, opened 2020-07-04
Add README.md for elasticdl_client folder.

brightcoder01:


elasticdl_client/README.md
```diff
@@ -0,0 +1,8 @@
+# ElasticDL Client
+
+This is the client command-line tool for ElasticDL. We can use it to build
```
skydoorkai: This is more straightforward:

`ElasticDL client is a command-line tool. Users can use it to prepare and submit ElasticDL jobs.`

brightcoder01: Good suggestion. Make some update based on this comment.

elasticdl_client/README.md
```diff
@@ -0,0 +1,8 @@
+# ElasticDL Client
+
+ElasticDL client is a command-line tool. Users can use it to prepare
+Docker images for a distributed ElasticDL job and submit a train, evaluate
```
workingloong: "Docker images" -> "the Docker image" 
Because a job only needs an image.

brightcoder01: Good catch, updated.

wangkuiyi (approved):

PR #2131, opened 2020-07-04
update experiment 1 of benchmark

QiJune:
I make some experiments at google cloud. The cluster has 3 nodes, and each node has 8 CPUs(7.91 CPUs actually. If we use 7 CPUs already, we could not request for 1 CPU more, since there is only 0.91 CPU left).

There are two jobs, each job requires 13 CPUs.

In experiment1, we first launch the first job. After 30 seconds, we launch the second job.

Please review the result in the [pdf doc](https://github.com/QiJune/elasticdl/blob/doc/update_benchmark/docs/benchmark/report_cn.pdf).  The total time of elastic scheduling is 26% less than gang scheduling. (Before, the value is 15%)


docs/benchmark/report_cn.md
```diff
@@ -69,14 +69,14 @@ scheduling，则两个作业并发执行，虽然后启动的作业拿不到期
 ![overlap jobs](./data/1.pdf)
 
 上图对应的实验里，我们用 gang scheduling
-的方式提交了两个训练作业，每个作业都需要 175 个 CPU。而集群总 CPU 数是
-320，不足同时运行两个作业，所以依次运行它们。可以看到第一个作业在 650
-秒时结束。随后集群花了一点时间调度，然后开始运行第二个作业，直到 1300 秒时结束。
+的方式提交了两个训练作业，每个作业都需要 13 个 CPU。而集群总 CPU 数是
```
wangkuiyi: 建议加几个字：而 Google Cloud 上租用的试验集群总 CPU 数是 24 —— 否则读者奇怪什么集群 CPU 这么少。

wangkuiyi (dismissed):
Approved, with a minor comment.

Thanks for the great and prompt work!

docs/benchmark/report_cn.md
QiJune: Done, 已修复

wangkuiyi (approved):

PR #2136, opened 2020-07-08
Add default value for the arguments to simplify the commands in tutorial.

brightcoder01:


workingloong (dismissed):
Excellent!

elasticdl_client/common/args.py
```diff
@@ -209,7 +209,7 @@ def add_common_params(parser):
         "default to master_resource_request",
     )
     parser.add_argument(
-        "--num_workers", type=int, help="Number of workers", default=0
+        "--num_workers", type=int, help="Number of workers", default=1
```
skydoorkai: default=2 to make it distributed?

elasticdl_client/common/args.py
```diff
@@ -309,7 +309,7 @@ def add_common_params(parser):
         "--num_minibatches_per_task",
         type=int,
         help="The number of minibatches per task",
-        required=True,
+        default=2,
```
skydoorkai: Is `2` too small for perf? 8?

elasticdl_client/common/args.py
brightcoder01: Good idea.

elasticdl_client/common/args.py
brightcoder01: Good idea.

skydoorkai (approved):

PR #2137, opened 2020-07-08
Add elasticdl_ps file path into PATH environment variable before executing the binary in the PS pod.

brightcoder01:
The logic of `Add elasticdl_ps file path into PATH environment variable` is in `/root/.bashrc`. And the script snippet for this logic is generated in the Dockerfile. Before executing `elasticdl_ps` binary in PS pod, we need execute `source /root/.bashrc` to update the PATH environment variable.


workingloong (approved):
LGTM

PR #2172, opened 2020-07-14
Set model zoo path in client init

### LINKED ISSUE #2171: Set the model zoo path in the `elasticdl zoo init`.

workingloong:
Fix #2171


QiJune (approved):
Excellent!

PR #2173, opened 2020-07-14
Fix allreduce args and workdir

### LINKED ISSUE #2169: Allreduce worker should not require ps_addrs
main.py: error: argument --ps_addrs: expected one argument

### LINKED ISSUE #2170: Allreduce Dockerfile should not set WORKDIR as `/root`

skydoorkai:
1. If using allreduce, master should not set worker args `--ps_addrs`.
1. Make sure to pass extra args (which includes `collective_communicator_service_name` for allreduce) to worker.
1. Do not change WORKDIR to `/root` for allreduce to be consistent with `ps`.

fix #2170 
fix #2169


workingloong (approved):
LGTM

PR #2177, opened 2020-07-14
Install local wheel packages in the image for development.

workingloong:


elasticdl_client/api.py
```diff
@@ -64,12 +64,19 @@ def init_zoo(args):
 {% if CLUSTER_SPEC_NAME  %}\
 COPY ./{{ CLUSTER_SPEC_NAME }} /cluster_spec/{{ CLUSTER_SPEC_NAME }}\
 {% endif %}
+
+{% if USE_SOURCE_REPO %}\
+COPY ./elasticdl /elasticdl
```
brightcoder01: Why do we copy the code into the docker image?
I think we can mount the code into the image for testing instead.

wangkuiyi: Yes, I have the same question?

workingloong: Using Mininkube, we can mount the code into containers for testing. However, we cannot mount modified codes into containers if the developer wants to run integration tests on On-prem Cluster or cloud. It is troublesome to place the modified codes on the remote storage of the cluster each time.

Yancey0623: Maybe L68-L72 would break the tutorial because the user is not at the root of this project.
> cd model_zoo
elasticdl zoo init
elasticdl zoo build --image=elasticdl:mnist .

c.f. https://github.com/sql-machine-learning/elasticdl/blob/develop/docs/tutorials/elasticdl_local.md#build-the-docker-image

How about rebuilding the base image if develops want to modify the source code of this project.

The argument `use_source_repo` is optional. We don't copy the source codes into the image with the default value. At the root of the project, the user can initialize an image by
```python
elasticdl zoo init --model_zoo=model_zoo
elasticdl zoo build --image=elasticdl:mnist .
```
I have added the `model_zoo` argument into `elasticdl zoo init` in #2172 .

QiJune: @workingloong For develops, we could directly use `elasticdl:ci` image, it copies the latest wheel package.

1. modify codes
2. mount the codes, use `elasticdl:dev` to run unit tests and generate three wheel package
3. build `elasticdl:ci` image, and submit distributed jobs to clusters based on it

workingloong: > @workingloong For develops, we could directly use `elasticdl:ci` image, it copies the latest wheel package.
> 
> 1. modify codes
> 2. mount the codes, use `elasticdl:dev` to run unit tests and generate three wheel package
> 3. build `elasticdl:ci` image, and submit distributed jobs to clusters based on it

In develop branch, we will generate wheel packages with version "develop" and we cannot use pip to install them.
For example, `pip install build/elasticdl_preprocessing-develop-py3-none-any.whl` will raise the following error.
image

If we rename the package to `elasticdl-0.2.0rc2-py3-none-any.whl`, `pip install` works.

Yancey0623: The word `-develop-` comes from the `version` field of `setup.py`, instead of using `develop`, we can pass the next version with `.dev` suffix like `0.2.0-rc2.dev`.

https://github.com/sql-machine-learning/elasticdl/blob/cdc76280320b4bb85a0d0c4131383205456afc3c/setup_preprocessing.py#L23-L26

workingloong: We can pip install wheel packages with word `-develop-` in Linux but not in Mac.
image

Yancey0623: `develop` is not a valid version, maybe we should modify it to  `0.2.0.rc2.dev`, c.f. https://www.python.org/dev/peps/pep-0440/#public-version-identifiers

elasticdl_client/api.py
```diff
@@ -52,10 +52,16 @@ def init_zoo(args):
     tmpl_str = """\
 FROM {{ BASE_IMAGE }} as base
 
+{% if USE_LOCAL_PKG %}\
+COPY build/*.whl /
```
brightcoder01: It will bring the limit that the local pkg path should be under the path `./build`.
How about using the argument name `local_pkg_dir` instead of a bool switch `use_local_pkg`

skydoorkai (approved):

PR #2179, opened 2020-07-15
add PSClient in worker

QiJune:
We want to support both TensorFlow and PyTorch.  Both TensorFlow worker and PyTorch worker will pull parameters from PS and push gradients to PS. We should implement a class called `PSClient`, which provides common methods for both TensorFlow worker and PyTorch worker.

It's the first PR for migrating methods to `PSClient`. I will move more PS related methods in next PRs.

### DIFF
--- a/elasticdl/python/worker/ps_client.py
+++ b/elasticdl/python/worker/ps_client.py
@@ -1,0 +1,62 @@
+# Copyright 2020 The ElasticDL Authors. All rights reserved.
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


skydoorkai:
We need to put all PS related calls to PSClient.
So this is the first PR to implement PSClient?

elasticdl/python/worker/ps_client.py
```diff
@@ -0,0 +1,62 @@
...
+        self._ps_num = len(self._ps_stubs)
+
+    def pull_embedding_vectors(self, layer_name, embedding_ids):
+        """
+        Pulls and returns embedding vectors ordered by the embedding ids.
+        Args:
+            layer_name: layer name
+            embedding_ids: a list of ids
+        Return:
+            embedding_vectors: a numpy ndarray
```
workingloong: a 2-D numpy array?

QiJune:
@skydoorkai  Yes, it's the first PR. I will move more PS related methods in next PRs.

elasticdl/python/worker/ps_client.py
QiJune: Done.

skydoorkai (approved):

PR #2181, opened 2020-07-15
clean checking variable created logic of worker

QiJune:
We support two styles of TensorFlow Keras APIs: subModel style API, and functional style API. Functional API creates variables before running a model, while subModel API does not. We handle these two cases with a lot of `if-else` logic.

This PR aims to clean the code. We merge functional API logic into submodel API logic. So, we will try to run `_run_model_call_before_training` method always. It does not bring any big overhead but simplify our code greatly.

### DIFF
--- a/elasticdl/python/worker/worker.py
+++ b/elasticdl/python/worker/worker.py
@@ -135,6 +135,7 @@ class Worker(object):
         self._init_from_args(args)
         self._timing = Timing(args.log_level.upper() == "DEBUG", self.logger)
         self._log_loss_count = 0
+        self._var_created = False
 
     def _init_from_args(self, args):
         """
@@ -229,15 +230,6 @@ class Worker(object):
         self._model = model_inst
         self._train_eagerly = False
         self._init_embeddings()
-        self._var_created = self._model.built
-        self._non_embed_vars = {}
-        if self._var_created:
-            for var in get_non_embedding_trainable_vars(
-                self._model, self._embedding_layers
-            ):
-                self._non_embed_vars[var.name] = var
-            if self._use_multi_ps:
-                self.init_ps_var_partition()
 
     def _init_embedding_layer(self):
         """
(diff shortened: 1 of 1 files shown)


elasticdl/python/worker/worker.py
```diff
@@ -679,12 +669,10 @@ def _run_model_call_before_training(self, features):
...
-        if not self._var_created:
-            if self._use_multi_ps:
-                self.init_ps_var_partition()
-            else:
-                self.report_variable()
-            self._var_created = True
+        self._var_created = True
+
+        if self._use_multi_ps:
+            self.init_ps_var_partition()
```
brightcoder01, line 675: Is it heavy to call `self.init_ps_var_partition` multiple times?

QiJune: It will only be called for once. 

`init_ps_var_partition` is called inside `_run_model_call_before_training`. The `_run_model_call_before_training` method is only called once.

elasticdl/python/worker/worker.py
```diff
@@ -679,12 +669,10 @@ def _run_model_call_before_training(self, features):
         ):
             self._non_embed_vars[var.name] = var
 
-        if not self._var_created:
-            if self._use_multi_ps:
-                self.init_ps_var_partition()
-            else:
-                self.report_variable()
-            self._var_created = True
+        self._var_created = True
```
workingloong, line 672: It seems that `self._var_created` is always `True`. How about moving it between L866 and L867?
https://github.com/sql-machine-learning/elasticdl/blob/2b0078dae70aaf6e827f3bbcf6dd2579c517adea/elasticdl/python/worker/worker.py#L866-L867

skydoorkai (approved):
lgtm

elasticdl/python/worker/worker.py
QiJune: At first, `self._var_created`  is `False`. Then, `_process_minibatch` will call `_run_model_call_before_training`, and set `self._var_created` to `True`.

So, the next time, `_run_model_call_before_training` will be skipped.

PR #2182, opened 2020-07-15
Support reliable ftlib consesus service connection

### LINKED ISSUE #2174: Support reliable worker pod detection for allreduce
In worker.py, it sleeps for 20 seconds, hoping that is enough for FTLib consensus service to detect worker pod.
This is not reliable and fails often.

```
    if args.distribution_strategy == DistributionStrategy.ALLREDUCE:
        logger.info(
            "Wait for %s seconds for FTLib consensus service to "
            "detect the worker pod" % str(_ALLREDUCE_STRATEGY_WARM_UP_SECS)
        )
        time.sleep(_ALLREDUCE_STRATEGY_WARM_UP_SECS)
```

skydoorkai:
Use retry instead of sleep 20 seconds.

fix #2174

### DIFF
--- a/elasticdl/python/collective_ops/communicator.py
+++ b/elasticdl/python/collective_ops/communicator.py
@@ -12,6 +12,7 @@
 # limitations under the License.
 
 import socket
+import time
 
 from elasticdl.python.common.constants import CollectiveCommunicatorStatus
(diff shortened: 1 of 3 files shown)


workingloong (approved):
LGTM

PR #2183, opened 2020-07-15
BUGFIX: elasticdl_ps is not found if the base image is tensorflow/tensorflow.

brightcoder01:
In the previous implementation, we append the installation path of `elasticdl_ps` binary into the PATH environment variable in the /root/.bash. The command for PS pod is `bash -c "source /root/.bashrc && elasticdl_ps ..."` to make sure `elasticdl_ps` can be found.

But the problem is that if using `tensorflow/tensorflow:2.1.0-py3` as the image, the `.bashrc` contains the following command at the head.

```SHELL
[ -z "$PS1" ] && return
```
If not running interactively, it doesn't do anything.

In order to fix this, we put the command of adding the `elasticdl_ps` path into envrionment variable into a new separate file `/root/.bashrc_elasticdl`, and then execute `source /root/.bashrc_elasticdl` in PS pod instead.

### DIFF
--- a/elasticdl/docker/Dockerfile.ci
+++ b/elasticdl/docker/Dockerfile.ci
@@ -31,7 +31,11 @@ RUN python -m pip install --quiet /elasticdl-develop-py3-none-any.whl \
     && rm /elasticdl-develop-py3-none-any.whl
 
 # Add elasticdl_ps to system path
-ENV PATH /usr/local/lib/python3.6/dist-packages/elasticdl/go/bin:$PATH
+RUN /bin/bash -c \
+        'PYTHON_PKG_PATH=$(pip3 show elasticdl | grep "Location:" | cut -d " " -f2); \
+        echo "PATH=${PYTHON_PKG_PATH}/elasticdl/go/bin:$PATH" >> \
+            /root/.bashrc_elasticdl; \
+        echo ". /root/.bashrc_elasticdl" >> /root/.bashrc'
 
 # Generate mnist checkpoint for evaluation and prediction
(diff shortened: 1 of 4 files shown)


workingloong (dismissed):
LGTM

workingloong (approved):

elasticdl/docker/Dockerfile.ci
```diff
@@ -31,7 +31,11 @@ RUN python -m pip install --quiet /elasticdl-develop-py3-none-any.whl \
     && rm /elasticdl-develop-py3-none-any.whl
 
 # Add elasticdl_ps to system path
-ENV PATH /usr/local/lib/python3.6/dist-packages/elasticdl/go/bin:$PATH
+RUN /bin/bash -c \
+        'PYTHON_PKG_PATH=$(pip3 show elasticdl | grep "Location:" | cut -d " " -f2); \
+        echo "PATH=${PYTHON_PKG_PATH}/elasticdl/go/bin:$PATH" >> \
+            /root/.bashrc_elasticdl; \
+        echo ". /root/.bashrc_elasticdl" >> /root/.bashrc'
```
skydoorkai, line 38: We are using `.bashrc_elasticdl` now, do we still need to append this PATH to `.bashrc`?

brightcoder01: In order to make sure that while running `docker run --rm -it <image_name> bash`, the `elasticd_ps` binary path will be appended to PATH environment variable. `source /root/.bashrc` is automatically executed when we runs the command above.

PR #2199, opened 2020-07-21
Move models of cifar10 to a folder.

workingloong:
We can separate models by datasets.


skydoorkai:
`mnist` also has two model definitions. We can also move them into a directory.

QiJune (approved):
Excellent!

workingloong:
> `mnist` also has two model definitions. We can also move them into a directory.

I will do it in the next PR.

PR #2202, opened 2020-07-21
migrate pull_dense_parameters method

QiJune:
Please review #2198  first.


elasticdl/python/worker/worker.py
```diff
@@ -304,40 +304,32 @@ def _update_local_model(self):
...
+                    parameters, ps_id, self._model_versions_from_ps[ps_id]
+                )
                 req = elasticdl_pb2.PullDenseParametersRequest()
                 req.version = self._model_versions_from_ps[ps_id]
-                var_future = stub.pull_dense_parameters.future(req)
-                variable_future_and_id_pairs.append((var_future, ps_id))
-
-            for var_future, ps_id in variable_future_and_id_pairs:
-                res = var_future.result()
+                res = self._ps_stubs[ps_id].pull_dense_parameters(req)
```
skydoorkai: `res` is not assigned to self._non_embed_vars.

Also, put the initialization logic into ps_client may make the code cleaner.

QiJune: 1. worker tries to pull dense parameters from the PS, and gets none
2. worker pushes local dense parameters to PS
3. worker pulls dense parameters from PS again, to make sure the dense parameters are initialized on the PS.

So, we need not assign the `res` to `self._non_embed_vars`. They have the same value.

QiJune: `PSClient` provides several basic client APIs for the PS, like `pull_dense_parameters`/`push_dense_parameters`. Initialization logic builds on top of these APIs. 
Right now, `self._non_embed_vars` are maintained by the worker, it's hard to move initialization logic into `PSClient`.

skydoorkai: Multiple workers may push dense parameters to PS and only one of them is accepted. Thus, workers need to assign the res to self._non_embed_vars as they may be different.

QiJune: Done, fix it.

elasticdl/python/worker/worker.py
```diff
@@ -304,39 +304,30 @@ def _update_local_model(self):
...
+            for ps_id in uninit_ps:
+                # push variable to ps for initialization
+                parameters = [
+                    Tensor(name, self._non_embed_vars[name].numpy(), None)
+                    for name in self._ps_client.ps_to_parameter[ps_id]
+                ]
+                self._ps_client.push_dense_parameters(
+                    parameters, ps_id, self._model_versions_from_ps[ps_id]
+                )
+                dense_params, uninit = self._ps_client.pull_dense_parameters(
```
skydoorkai: `dense_params` will be overridden by the last PS in uninit_ps and params from other PSs are lost here.

QiJune: Done. Fix it.

skydoorkai (dismissed):

elasticdl/python/worker/worker.py
```diff
@@ -314,39 +314,32 @@ def get_model(self):
...
-                        Tensor(name, self._non_embed_vars[name].numpy(), None)
-                        for name in self._ps_client.ps_to_parameter[ps_id]
-                    ]
-                    self._ps_client.push_dense_parameters(
-                        parameters, ps_id, self._model_versions_from_ps[ps_id]
+            dense_params, uninit_ps = self._ps_client.pull_dense_parameters(
+                [i for i in range(self._ps_num)], self._model_versions_from_ps
+            )
+
+            for ps_id in uninit_ps:
```
brightcoder01, line 326: How about adding more comments for these two steps
1) First pull dense parameters from all the PS node, maybe one or more ps is uninitialized
2) Then we will push the dense parameter from the worker pod to these Pservers to initialize its partition of parameters.

QiJune: Good idea! Add it.

brightcoder01: How about make this procedure three steps:
1) pull_dense_parameter from all the ps
2) Initialize the uninit ps if there is any -> wrap this process in a separate member function
3) pull_dense_parameters from all the ps again and then assign _non_embed_vars

QiJune: @brightcoder01 After the PS is initialized, we need not call `push_dense_parameters`. So, it's a waste of time to call `pull_dense_parameters` twice.

brightcoder01: The second pull_dense_parameters is only called when there is any uninitialize PS node. Just like initialize the singleton if the instance is null.

```Python
dense_param, unit_ps = pull_dense_param

if uninit_ps:
   initialize_ps_nodes(uninit_ps)
   dense_param, unit_ps = pull_dense_params
   assert(len(unit_ps) == 0)

assign_nonembed_var(dense_param)
···

**Review by skydoorkai — APPROVED** [review]

File: elasticdl/python/worker/worker.py (hunk 3 above)
Comment (QiJune) [reply elasticdl/python/worker/worker.py:326]: @brightcoder01 I will continue polish codes in the next PR.

Comment (skydoorkai) [reply elasticdl/python/worker/worker.py:326]: Yes, the code can be revised to be more clear such as:

```
dense_param, unit_ps = pull_dense_param(all_ps)

if uninit_ps:
   initialize_ps_nodes(uninit_ps)
   more_dense_param, unit_ps = pull_dense_params(uninit_ps)
   assert(len(unit_ps) == 0
   dense_param.update(more_dense_param)

assign_nonembed_var(dense_param)

```

Issue #2192, opened 2020-07-16
FTlib consensus is unreliable and slow, we should use master to do the consensus.


skydoorkai:

skydoorkai:
FTlib has two main components: consensus and communicator (init process group, broadcast, allreduce, barrier, etc).
In order to use the master for consensus implementation, we need to write similar two components in ElasticDL.

skydoorkai:
Tested with CIFAR10 training with 3 workers, the consensus is quite slow(30s or more) and occasionally fails.

The command to submit the job:
```
elasticdl train \
        --image_name=${your_image_hub}/elasticdl:allreduce \
 		  --model_zoo=/model_zoo \
 		  --model_def=cifar10_functional_api.cifar10_functional_api.custom_model \
 		  --training_data=/testdata/elasticdl/cifar10/train \
 		  --validation_data=/testdata/elasticdl/cifar10/test \
 		  --num_epochs=1 \
 		  --master_resource_request="cpu=1,memory=4096Mi,ephemeral-storage=10240Mi" \
 		  --master_resource_limit="cpu=1,memory=4096Mi,ephemeral-storage=10240Mi" \
 		  --worker_resource_request="cpu=1,memory=4096Mi,ephemeral-storage=10240Mi" \
 		  --worker_resource_limit="cpu=1,memory=4096Mi,ephemeral-storage=10240Mi" \
 		  --minibatch_size=64 \
 		  --num_minibatches_per_task=8 \
 		  --num_workers=3 \
 		  --job_name=edl-allreduce-test \
 		  --image_pull_policy=Always \
 		  --distribution_strategy=AllreduceStrategy \
                  --volume='claim_name=your_data_volume,mount_path=/testdata' 
```

Occasional failure:

```
 3742 Traceback (most recent call last):
 3743   File "/usr/lib/python3.6/runpy.py", line 193, in _run_module_as_main
 3744     "__main__", mod_spec)
 3745   File "/usr/lib/python3.6/runpy.py", line 85, in _run_code
 3746     exec(code, run_globals)
 3747   File "/elasticdl/python/worker/main.py", line 89, in <module>
 3748     main()
 3749   File "/elasticdl/python/worker/main.py", line 83, in main
 3750     set_parallelism=True,
 3751   File "/elasticdl/python/worker/worker.py", line 135, in __init__
 3752     self._init_from_args(args)
 3753   File "/elasticdl/python/worker/worker.py", line 174, in _init_from_args
[...]

PR #2204, opened 2020-07-21
Add allreduce timing and retry when join connection fails

skydoorkai:
1. Add timing for all reduce
1. Retry when consensus group join fails.

### DIFF
--- a/elasticdl/python/collective_ops/communicator.py
+++ b/elasticdl/python/collective_ops/communicator.py
@@ -77,11 +77,30 @@ class CollectiveCommunicator(object):
                     ),
                 },
             )
+            connection_try_num = 0
             while peer_list and not self._ftlib.consensus_joined():
                 logger.warning("Retry building consensus...")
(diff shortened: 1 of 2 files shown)


elasticdl/python/collective_ops/communicator.py
```diff
@@ -77,11 +77,30 @@ def __init__(self, service_name=None):
...
+                        > _FTLIB_CONSENSUS_CONNECTION_TIMEOUT_SECS
+                    ):
+                        logger.error(
+                            "Cannot join FTLib consensus service in %s "
+                            "seconds",
+                            str(_FTLIB_CONSENSUS_CONNECTION_TIMEOUT_SECS),
+                        )
+                        self._ftlib = None
+                        return
+                    logger.info(
```
workingloong: `logger.warning` maybe better?

workingloong:
We can attach the issue #2192 here.

workingloong (approved):
LGTM

PR #2206, opened 2020-07-22
clean use_multi_ps flag in worker

QiJune:
`use_multi_ps` flag is duplicated with the parameter server strategy flag. This PR removes it.


skydoorkai (approved):
lgtm

PR #2208, opened 2020-07-22
clean members of TaskDataService class

QiJune:
`TaskDataService` is actually a data generator. It gets tasks from the master and yield records. It should not depend on the `Worker` class.


workingloong (approved):
Excellent !

PR #2209, opened 2020-07-22
migrate push_embedding_table_infos method

QiJune:


elasticdl/python/worker/worker.py
```diff
@@ -341,46 +340,44 @@ def get_model(self):
                     raise RuntimeError(
                         "PS pod %d cannot be initialized" % ps_id
                     )
-                for k, v in ps_params.items():
-                    self._non_embed_vars[k].assign(v)
+                dense_params.update(ps_params)
```
skydoorkai, line 343: We can move out the pull_dense_parameters  from the for loop, and use one call instead:

```
ps_params, uninit = self._ps_client.pull_dense_parameters(
     uninit_ps, self._model_versions_from_ps)
dense_params.update(ps_params)
```

QiJune: Good idea, fix it.

skydoorkai (approved):

PR #2210, opened 2020-07-22
Train the model locally if there is only one worker using AllReduce strategy.

workingloong:


elasticdl/python/common/args.py
```diff
@@ -200,6 +200,9 @@ def parse_worker_args(worker_args=None):
     parser.add_argument(
         "--worker_id", help="ID unique to the worker", type=int, required=True
     )
+    parser.add_argument(
+        "--num_workers", help="The number of workers", type=int, required=True
```
terrytangyuan: This new required arg will break compatibility. Is this intended?

workingloong: Yes. If `num_workers=1`, we cannot use `CollectiveCommunicator` which cannot work if only a worker is configured. We can train the model locally without `CollectiveCommunicator` with a worker.

terrytangyuan: Yes but you are adding a new required argument here in common args, which will potentially break existing usage.

skydoorkai: Workers are launched by the master. 
The master will always have`--num_workers` argument which will be passed to workers.

Thus, nothing will be broken.

skydoorkai: But need to add the argument in the unittest.

terrytangyuan: Ok

terrytangyuan (dismissed):

skydoorkai (approved):

PR #2211, opened 2020-07-23
clean member of worker class

QiJune:
Please review #2209 first.


elasticdl/python/worker/worker.py
```diff
@@ -341,46 +332,44 @@ def get_model(self):
...
                 # tf.keras.initializers. Keep aligned between these two.
-                embedding_info.initializer = "uniform"
-                # set to float32
-                embedding_info.dtype = dtype_numpy_to_tensor(
-                    np.dtype("float32")
+                infos.append(
+                    EmbeddingTableInfo(
+                        column.embedding_weight_name,
+                        column.dimension,
+                        "uniform",
```
workingloong: Can we wrap the string "uniform" into a constant object?

QiJune: Good idea. Fix it.

skydoorkai (approved):

PR #2231, opened 2020-08-06
Polish AllReduce benchmark report

workingloong:


wangkuiyi (dismissed):

wangkuiyi (dismissed):

elasticdl/python/tests/test_utils.py
```diff
@@ -275,6 +275,10 @@ def create_pserver(
         pserver = ParameterServer(args)
         pserver.prepare()
         pservers.append(pserver)
+
+    for channel in channels:
+        grpc.channel_ready_future(channel).result()
```
brightcoder01: If we don't need the result object of the future, we can use future.wait().
Or futures.wait(future_list)

QiJune (approved):
Excellent!

PR #2234, opened 2020-08-10
Check GRPC ready before executing unit tests

workingloong:


elasticdl/python/tests/test_utils.py
```diff
@@ -275,6 +275,10 @@ def create_pserver(
         pserver = ParameterServer(args)
         pserver.prepare()
         pservers.append(pserver)
+
+    for channel in channels:
+        grpc.channel_ready_future(channel).result()
```
brightcoder01, line 280: We can try futures.wait(future_list)

workingloong: There is no `wait` in `_ChannelReadyFuture` of GRPC.
image

And we cannot use `_ChannelReadyFuture`  in ``futures.wait()` for the error:
image

skydoorkai (approved):

PR #2235, opened 2020-08-10
Design to implement elastic AllReduce with Horovod

workingloong:


docs/designs/allreduce_based_horovod.md
```diff
@@ -0,0 +1,117 @@
+# Design for Elastic AllReduce-based Training Support Based on Horovod
```
QiJune: The document name should be horovod based allreduce.

docs/designs/allreduce_based_horovod.md
```diff
@@ -0,0 +1,117 @@
...
+
+This document describes the design for supporting AllReduce-based distributed
+training based on Horovod in ElasticDL.
```
QiJune: for suppporting AllReduce distribution strategy.

and please give a link to Horovod.

docs/designs/allreduce_based_horovod.md
```diff
@@ -0,0 +1,117 @@
...
+## Motivation
+
+We have developed elastic AllReduce based on FTlib in ElasticDL.
+From the [benchmark report](../benchmark/ftlib_benchmark.md), we can
+find that the performance of FTlib for ResNet50 is worse than Horovod.
+What's more, it is not stable to build consensus by gossip protocol in FTlib
+like the [issue](https://github.com/sql-machine-learning/elasticdl/issues/2192#issuecomment-664096185).
+
+FTlib uses Gloo to implement elastic AllReduce because the worker process
+can catch the exception from Gloo and not exit if the AllReduce operator
```
QiJune: Does AllReduce operator here means `Broadcast`/`AllReduce`/`AllGather`? It's different from Kubernetes operator.

docs/designs/allreduce_based_horovod.md
```diff
@@ -0,0 +1,117 @@
...
+fails. Horovod also can use Gloo as backend. What's more, there are many
+optimizations like [Tensor Fusion](https://horovod.readthedocs.io/en/latest/tensor-fusion_include.html)
+in Horovod to improve performance. So, the performance of Horovod for ResNet50 is
+better than FTlib because ResNet50 has may small tensors.
```
QiJune: There are many small parameters in the ResNet50 model. We have to launch an AllReduce operator to synchronize each parameter. It brings a lot of overhead. There are many optimizations like [Tensor Fusion] in Horovod to reduce the overhead. So, the performance of Horovod for ResNet50 is better than FTlib.

docs/designs/allreduce_based_horovod.md
```diff
@@ -0,0 +1,117 @@
...
+
+Horovod provides Python APIs for Gloo and we can use those APIs in ElasticDL
+to implement elastic AllReduce.
```
QiJune: We have to answer one question: ElastiDL uses Horovod to implement the AllReduce distribution strategy. Why should users use ElasticDL, not Horovod directly?

docs/designs/allreduce_based_horovod.md
skydoorkai: FTlib uses the gossip protocol to build consensus, which is not stable as described in this issue(link).

skydoorkai: if any collective communications fail.

docs/designs/allreduce_based_horovod.md
```diff
@@ -0,0 +1,117 @@
...
+
+## ElasticDL Re-initialize Horovod When the Number of Workers Changes
+
+When using Horovod with Gloo backend, we need to create a `RendezvousServer`
```
skydoorkai: When using Horovod with Gloo backend, we need to create a `RendezvousServer`, which  has a KVStore. Also we need to put a host plan into this KVStore. A host plan includes worker hosts and their assigned ranks,  which are required by Gloo. 
The master in ElasticDL is responsible to create the `RendezvousServer`.  To support for elastic, when the master detects worker number change, it will create a new host plan and put it in the KVStore.

docs/designs/allreduce_based_horovod.md
```diff
@@ -0,0 +1,117 @@
...
+Then, the worker can call `hvd.init` to initialize the Gloo context for
+AllReduce.
+
+When the master finds the number of workers changes, it can re-create a new
+`RendezvousServer` and notify workers to re-initialize Horovod.
+In the Kubernetes cluster, the number of workers may change for the
+following reasons:
+
+1. Some workers fail because of preemption.
+1. The master re-launches new workers.
```
skydoorkai: A worker pod status becomes running. 
(master re-launches new workers  != worker pod is up and running)

docs/designs/allreduce_based_horovod.md
workingloong: I have answered the question in the next PR. You can review it.

skydoorkai (approved):

PR #2236, opened 2020-08-12
Install Horovod and dependencies in the dev image

### LINKED ISSUE #2226: Add Horovod package into dev Dockerfile

workingloong:
Fix #2226


elasticdl/docker/Dockerfile
```diff
@@ -72,3 +90,10 @@ RUN cd /root && git clone --depth=1 https://github.com/caicloud/ftlib.git
 RUN cd /root/ftlib && python -m pip install --quiet -r requirements.txt
 RUN cd /root/ftlib/ftlib/consensus/gossip && bash ./gen_shared_lib.sh
 RUN cp -r /root/ftlib/ftlib /usr/local/lib/python3.6/dist-packages/ftlib
+
+ENV HOROVOD_PATH /tmp/horovod
+RUN cd /tmp \
+        && git clone --recursive https://github.com/horovod/horovod.git
+
+RUN cd ${HOROVOD_PATH} && HOROVOD_WITH_TENSORFLOW=1 HOROVOD_WITH_PYTORCH=1  \
```
brightcoder01: Why do we clone horovod and run `python setup.py install` instead of `pip install horovod`?

elasticdl/docker/Dockerfile
```diff
@@ -72,3 +90,10 @@ RUN cd /root && git clone --depth=1 https://github.com/caicloud/ftlib.git
 RUN cd /root/ftlib && python -m pip install --quiet -r requirements.txt
 RUN cd /root/ftlib/ftlib/consensus/gossip && bash ./gen_shared_lib.sh
 RUN cp -r /root/ftlib/ftlib /usr/local/lib/python3.6/dist-packages/ftlib
+
+ENV HOROVOD_PATH /tmp/horovod
+RUN cd /tmp \
+        && git clone --recursive https://github.com/horovod/horovod.git
```
skydoorkai: We should fix with a version/commit of horovod.

elasticdl/docker/Dockerfile
workingloong: There is no elastic module in the latest version `0.19.5` of Horovod. So, we need to clone Horovod and use source codes to install it.

skydoorkai (approved):

Issue #2228, opened 2020-08-06
Master detects worker pod change and create/re-create RendezvousServer.


skydoorkai:

workingloong:
We need not to re-create a RendezvousServer and only need to initialize the `host_alloc_plan`.
```python
# Assign host string as worker domain names or IPs.
hosts_str = "172.18.0.22:1,172.18.0.23:1,172.18.0.24:1"
# Set number of workers
num_proc = 3

host_alloc_plan = get_host_plan(hosts_str, num_proc)
rendezvous = RendezvousServer(verbose=True)
global_rendezv_port = rendezvous.start()
rendezvous.init(host_alloc_plan)
```

PR #2240, opened 2020-08-13
Get alive worker service address

workingloong:
The master can detect worker pods change by the service addresses of alive workers.
#2228

### DIFF
--- a/elasticdl/python/master/k8s_instance_manager.py
+++ b/elasticdl/python/master/k8s_instance_manager.py
@@ -21,6 +21,8 @@ from elasticdl.python.common.constants import PodStatus
 from elasticdl.python.common.log_utils import default_logger as logger
 from elasticdl_client.common.constants import BashCommandTemplate
 
+_SERVICE_ADDR_SEP = ","
+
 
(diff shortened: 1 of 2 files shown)


elasticdl/python/master/k8s_instance_manager.py
```diff
@@ -377,22 +356,32 @@ def _event_cb(self, event):
...
-            with self._lock:
-                self._update_worker_addr(worker_id, new_worker_id)
         elif relaunch_ps:
             logger.info("Relaunching ps.")
             # Note: the ID and service address for relaunched parameter
             # server are intentionally left unchanged to support fault
             # tolerance.
             self._start_ps(ps_id)
 
+        self._worker_addrs = self._get_alive_worker_service__addr()
```
terrytangyuan: Redundant underscore here? Has the new code been tested?

workingloong: The unit test has failed for the reason and I will fix it.

skydoorkai (approved):

PR #2257, opened 2020-08-20
Broadcast model variables from the worker with rank=0

workingloong:


elasticdl/python/tests/allreduce_trainer_test.py
```diff
@@ -40,7 +40,7 @@ def test_training_process_with_fault_tolerance(self):
         version, _ = self._trainer.training_process_with_fault_tolerance(
             features, labels
         )
-        self.assertEqual(version, 1)
+        self.assertEqual(version, 2)
```
skydoorkai, line 46: Add a comment explaining why it is 2 not 1.

elasticdl/python/worker/allreduce_trainer.py
```diff
@@ -61,8 +64,14 @@ def _training_process(self, features, labels):
         return loss
 
     def training_process_with_fault_tolerance(self, features, labels):
+        if not self._var_created:
+            self._run_model_call_locally(features, labels)
+
         for _ in range(DEFAULT_MAX_ALLREDUCE_RETRY_NUM + 1):
             try:
+                if self._need_broadcast:
+                    self._broadcast_model()
```
brightcoder01, line 73: Do the workers other than rank 0 need broadcast too?

workingloong: Yes,  other workers need to execute `broadcast` to receive data from the source worker.

skydoorkai (approved):

PR #2264, opened 2020-08-23
 Don't initialize rendezvous for the new hosts if the last rendezvous has not completed

### LINKED ISSUE #2259: Master locks the rendezvous server when the worker initializes Horovod.

workingloong:
Fix #2259

### DIFF
--- a/elasticdl/python/master/rendezvous_server.py
+++ b/elasticdl/python/master/rendezvous_server.py
@@ -11,6 +11,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from elasticdl.python.common.log_utils import default_logger as logger
+
 try:
     from horovod.runner.common.util.hosts import (
         get_host_assignments,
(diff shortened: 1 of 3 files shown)


elasticdl/python/master/rendezvous_server.py
```diff
@@ -68,9 +75,21 @@ def get_rendezvous_port(self):
...
+            self._init_rendezvous_server()
+
         # -1 if host not in worker_hosts list.
         if host not in self._worker_hosts:
             return -1
+
+        if not self._rendezvous_completed:
+            self._ready_worker_hosts.add(host)
+            # If all active workers in the rendezvous are ready,
+            # the server can start to set hosts for the next rendezvous
```
skydoorkai: After a worker gets its rank, it needs to call `hvd.init()`. If the master set a new host plan before worker calls  `hvd.init()`, there may be crash/hang.

So the master needs to wait a few seconds after the last worker gets the rank to start the next round of rendezvous.

skydoorkai (dismissed):

skydoorkai (approved):

PR #2265, opened 2020-08-24
Adjust learning rate according to the world size.

workingloong:


elasticdl/python/worker/allreduce_trainer.py
```diff
@@ -41,6 +41,7 @@ def __init__(self, master_client, master_addr, model, loss_fn, optimizer):
         self._model = model
         self._loss = loss_fn
         self._optimizer = optimizer
+        self._learning_rate = optimizer.lr.numpy()
```
skydoorkai: learning rate may be a callable.
So, also use a callable to substitute it. 

Reference:
https://github.com/sql-machine-learning/elasticdl/blob/5fe8f4bc01cb0358926066b8d9dd014692e05f26/elasticdl/python/ps/learning_rate_modulator.py#L71

skydoorkai (approved):

PR #2275, opened 2020-08-31
Remove worker service, and use pod ip instead.

skydoorkai:


elasticdl/python/master/k8s_instance_manager.py
```diff
@@ -101,7 +101,7 @@ def __init__(
 
         # Protects followed variables, which are accessed from event_cb.
         self._lock = threading.Lock()
-        # worker id to (pod name, phase) mapping
+        # worker id to (pod name, ip, phase) mapping
```
workingloong, line 104: We can get pod by pod name and get IP by `pod.status.IP`. Should we need to set IP here?

elasticdl/python/master/k8s_instance_manager.py
```diff
@@ -359,30 +369,34 @@ def _event_cb(self, event):
 
     def get_alive_workers(self):
         alive_workers = []
-        for pod_name, phase in self._worker_pods_phase.values():
+        for pod_name, _, phase in self._worker_pods_ip_phase.values():
```
workingloong, line 372: We can get pod IP by 
```python
pod = k8s_client.get_pod(pod_name)
ip = pod.status.pod_ip
```
https://github.com/kubernetes-client/python/blob/master/kubernetes/docs/V1PodStatus.md

skydoorkai: After we record pod_ip, we don't need to call k8s API again.

workingloong (approved):
LGTM

PR #2283, opened 2020-09-07
Remove unused unit test of local executor

workingloong:


QiJune:
@workingloong Should we also remove the `local_executor.py` ?

skydoorkai (approved):

workingloong:
> @workingloong Should we also remove the `local_executor.py` ?

Yes, I will remove it in the next PR.

PR #2286, opened 2020-09-08
Set learning rate 0.01 for the mnist model.

workingloong:
lr=0.1 may be big for AllReduce and the training cannot converge.

### DIFF
--- a/model_zoo/mnist/mnist_functional_api.py
+++ b/model_zoo/mnist/mnist_functional_api.py
@@ -63,7 +63,7 @@ def loss(labels, predictions):
     )
 
 
-def optimizer(lr=0.1):
+def optimizer(lr=0.01):
     return tf.optimizers.SGD(lr)
 
 


brightcoder01 (approved):

model_zoo/mnist/mnist_functional_api.py
```diff
@@ -63,7 +63,7 @@ def loss(labels, predictions):
     )
 
 
-def optimizer(lr=0.1):
```
brightcoder01, line 66: How about the subclass model?

skydoorkai (approved):

model_zoo/mnist/mnist_functional_api.py
workingloong: The learning rate of the subclass model is 0.01.
https://github.com/sql-machine-learning/elasticdl/blob/2225a079778313c1b8173821357c6e8d787658d4/model_zoo/mnist/mnist_subclass.py#L60

PR #2289, opened 2020-09-08
Remove TensorFlow source codes after building go modules

### LINKED ISSUE #2243: It is redundant to git clone TensorFlow repo.
We only need those proto files to build elasticdl.proto, so we can only download those files by `wget`. Because TensorFlow repo is huge which makes the image huge.

https://github.com/sql-machine-learning/elasticdl/blob/5fac16d29854001f8f68d4a0d514f4f57b95b8eb/elasticdl/docker/Dockerfile#L57-L60

workingloong:
Fix #2243


egolearner:
It's necessary to run git clone and rm in the same RUN command to keep image small.

```
ENV TF_PATH /tmp/tensorflow
RUN cd /tmp \
        && git clone --depth=1 --branch v2.2.0-rc0 \
        https://github.com/tensorflow/tensorflow && mv xxx.proto some_dir && rm -rf /tmp/tensorflow/xxx
```

PR #2292, opened 2020-09-14
Fix allreduce when one worker starts with size=1

### LINKED ISSUE #2291: AllReduce may hang when one worker starts firsts
This is because in Horovod, when `hvd.size() == 1`, it won't use allreduce op.

https://github.com/horovod/horovod/blob/e4554de96100f5a0e8686cd41cf99a6fe8a71e62/horovod/tensorflow/__init__.py#L159

skydoorkai:
1. Set env properly so that allreduce calls allreduce op even if size=1
1. Put rendezvous_server alive_worker process in lock.

Fix #2291


workingloong (dismissed):
LGTM

workingloong (approved):

PR #2296, opened 2020-09-16
Fix learning rate scheduler for the CNN model of cifar10

workingloong:


model_zoo/cifar10/cifar10_functional_api.py
```diff
@@ -138,7 +138,7 @@ def _schedule(model_version):
         else:
             return 0.001
 
-    LearningRateScheduler(_schedule)
+    return [LearningRateScheduler(_schedule)]
```
skydoorkai, line 141: any other similar bugs?

workingloong: Yeah, and I have fixed it.

skydoorkai (approved):

PR #2297, opened 2020-09-16
Check whether new workers join the job and re-initialize Horovod if True

workingloong:


skydoorkai (dismissed):

elasticdl/python/worker/allreduce_trainer.py
```diff
@@ -159,6 +174,7 @@ def _run_model_call_locally(self, features, labels):
         self._optimizer.apply_gradients(
             zip(grads, self._model.trainable_variables)
         )
+        K.set_value(self._optimizer.iterations, 0)
```
skydoorkai, line 178: Add a todo to handle the case that the model is initialized from a checkpoint.

skydoorkai (approved):

PR #2299, opened 2020-09-18
Scale learning rate according to the world size

workingloong:


elasticdl/python/worker/allreduce_trainer.py
```diff
@@ -132,10 +120,18 @@ def init_horovod_if_needed(self):
...
-            self._world_size = hvd.size()
+            self._scale_learning_rate()
             self._rendezvous_id = rank_response.rendezvous_id
             self._need_broadcast = True
 
+    def _scale_learning_rate(self):
+        """Scale the learning rete according to the world size"""
+        if self._last_world_size is not None:
+            self._optimizer.lr = (
+                self._optimizer.lr * hvd.size() / self._last_world_size
```
skydoorkai, line 131: 1. What if original lr is callable?

2. In initialization, `self._last_world_size` is none. Then it is set to the 1st `hvd.size()`. Thus, the learning rate depends on the 1st hvd.size(), which is not correct.

workingloong:
Users can adjust the learning rate by calling `hvd.size()`. So we don't need to adjust learning rate.

PR #2311, opened 2020-09-28
Set worker IP into HOROVOD_HOSTNAME

workingloong:


elasticdl/python/worker/allreduce_trainer.py
```diff
@@ -13,6 +13,7 @@
 
 import os
 import time
+import socket
```
skydoorkai, line 15: Change import order for isort?

elasticdl/python/worker/allreduce_trainer.py
```diff
@@ -151,7 +152,8 @@ def _set_horovod_env(self):
             os.environ[HorovodEnv.RENDEZVOUS_ADDR] = self._rendezvous_addr
         os.environ[HorovodEnv.CONTROLLER] = "gloo"
         os.environ[HorovodEnv.CPU_OPERATIONS] = "gloo"
-        os.environ[HorovodEnv.HOSTNAME] = "master"
+        domain_ip = socket.gethostbyname(socket.gethostname())
+        os.environ[HorovodEnv.HOSTNAME] = domain_ip
```
brightcoder01, line 156: Is it required by the Horovod API?

workingloong: Yes. The RendezvousServer in Horovod initializes using worker IPs. The worker needs to send its IP to RendezvousServer in the master to create the communication ring. Horovod will get the IP from the environment, so we need to set IP into the environment.

skydoorkai (approved):

PR #2312, opened 2020-09-28
Do not scale learning rate of optimizer.

workingloong:
User can scale the learning rate of the optimizer by defining callbacks in the model definition like:
```python
def callbacks():
    def _schedule(model_version):
        LR = 0.001
        if model_version > 10000:
            lr = LR * 0.01
        elif model_version > 6000:
            lr = LR * 0.1
        elif model_version > 4000:
            lr = LR * 0.2
        elif model_version > 2000:
            lr = LR * 0.5
        else:
            lr = LR
        return lr * hvd.size()

    return [LearningRateScheduler(_schedule)]
```

### DIFF
--- a/.isort.cfg
+++ b/.isort.cfg
@@ -1,5 +1,5 @@
 [settings]
 multi_line_output=3
 line_length=79
-known_third_party = PIL,deepctr,docker,google,grpc,jinja2,kubernetes,matplotlib,numpy,odps,pandas,recordio,requests,setuptools,sklearn,tensorflow,yaml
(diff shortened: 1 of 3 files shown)


brightcoder01:
Please add some comments why we don not scale the LR for the optimizer.

skydoorkai (dismissed):

workingloong:
> Please add some comments why we don not scale the LR for the optimizer.

I have explained in the message of the PR. Do you mean we should add the comment in the code?

skydoorkai (approved):

PR #2316, opened 2020-10-12
Print loss log only when minibatch training is accepted

skydoorkai:
This will avoid unnecessary loss logs when AllReduce fails due to worker pod number change during training.


workingloong (approved):
LGTM

PR #2317, opened 2020-10-12
Rename method name to better describe its functionality

skydoorkai:
Rename method  `_process_minibatch_and_report` to _`safe__process_minibatch` as the method does not report but only process minibatch with exception catching.


workingloong (approved):
Excellent！

PR #2319, opened 2020-10-13
Add a example to train DeepCTR with the criteo dataset

### LINKED ISSUE #2318: Add a tutorial to use ElasticDL to train DeepCTR models on GCloud and Aliyun.
[DeepCTR](https://github.com/shenweichen/DeepCTR) is a popular repo for CTR prediction. DeepCTR uses Keras API to define models and users can only train those models locally now. ElasticDL provides distributed training for Keras models on Kubernetes. So, we can write a tutorial to tell users how to use ElasticDL to train DeepCTR models on the Kubernetes cluster of Google Cloud and Aliyun.

workingloong:
Fix #2318


wangkuiyi (approved):
Just want to make sure that we will have a Markdown document explaining how to use this example.

PR #2320, opened 2020-10-13
Rename dataset_fn to feed

### LINKED ISSUE #2085: Rename dataset_fn to feed

workingloong:
fix #2085

### DIFF
--- a/docs/blogs/ata_deepctr_modeling.md
+++ b/docs/blogs/ata_deepctr_modeling.md
@@ -124,7 +124,7 @@ def eval_metrics_fn():
         )
     }
 
-def dataset_fn(dataset, mode, metadata):
+def feed(dataset, mode, metadata):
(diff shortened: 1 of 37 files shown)


terrytangyuan (commented):
What's the reason for this change?

brightcoder01 (dismissed):

workingloong:
> What's the reason for this change?



> What's the reason for this change?
`feed` describes that users can define the function to transform raw data into tensors for the model. `dataset_fn` is only to transform `tf.Data.Dataset` and the function name may not be suitable for `Dataloader` in PyTorch.

terrytangyuan:
`feed` by itself sounds very strange to me though. What about some alternatives like `transform_fn`, `preprocess_fn`, or `data_loader`?

skydoorkai (approved):

PR #2321, opened 2020-10-14
Add cluster_spec_json

### LINKED ISSUE #2313: Add 'cluster_spec_json' argument to pass cluster-spec information as json string
Some on-premise Kubernetes cluster may require users to add some specific specs to pod and/or service, such as label, annotation, toleration, affinity, environment variables, etc. 

Currently we have `cluster_spec` argument, and users can write Python codes to add these add these specific specs. Users need to save Python  codes in a file and either put this file in the image or in a shared storage.

Below is an example

```python
from kubernetes import client

class MyCluster:
    def __init__(self):
        self._cluster = "my_cluster"
        self._pool = "elasticdl"
        self._app_name = "elasticdl"

    def with_pod(self, pod):
        self._with_labels(pod)
        self._with_tolerations(pod)
        self._with_annotations(pod)
        return pod

    def with_service(self, service):
        # Use ClusterIP
        service.spec.type = "ClusterIP"
        service.spec.cluster_ip = "None"
        return service

    def _with_labels(self, pod):
        pod.metadata.labels["my_cluster/app-name"] = "elasticdl"
        pod.metadata.labels["my_cluster/deploy-unit"] = "elasticdl"

    def _with_tolerations(self, pod):
        tolerations = [
            client.V1Toleration(
                effect="NoSchedule",
                key="my_cluster.com/app-logic-pool",
                operator="Equal",
                value="myvalue",
            ),
        ]
        pod.spec.tolerations = tolerations

    def _with_annotations(self, pod):
        pod.metadata.annotations[
            "tags.mycluster.com/log"
        ] = "enabled"

cluster = MyCluster()
```

When ElasticDL operator launches an ElasticDL master, it needs to tell the master how to set these specific specs when launching worker and ps pods. It is convenient for ElasticDL operator to use `cluster_spec` argument, as this requires to create a Python file store in shared storage. 

Thus, we add a new argument 'cluster_spec_json', which is a json string containing all these specific specs. Below is the corresponding json for the example above.

```
{
   "pod_spec": {
      "metadata": null,
      "annotations": {
         "tags.mycluster.com/log": "enabled"
      },
      "labels": {
         "my_cluster/app-name": "elasticdl",
         "my_cluster/deploy-unit": "elasticdl",
      },
       "tolerations": [
       {
         "effect": "NoSchedule",
         "key": "mycluster.com/app-logic-pool",
         "operator": "Equal",
         "value": "myvalue"
       }
   ]
   },
   "service_spec": {
      "spec": {
         "clusterIP": "None",
         "type": "ClusterIP"
      }
   }
}
```

The master can parse this json string and add corresponding specs to worker and ps pods and services.

skydoorkai:
Fix #2313

### DIFF
--- a/elasticdl_client/common/k8s_client.py
+++ b/elasticdl_client/common/k8s_client.py
@@ -87,8 +87,7 @@ def get_instance_from_value(type_name, value):
             return value
     # value must be a dict
     args = {}
-    # Note that swagger_types is renamed to openapi_types in kubernetes 11.x
-    for attr, attr_type in six.iteritems(cls.swagger_types):
+    for attr, attr_type in six.iteritems(cls.openapi_types):
         if cls.attribute_map[attr] in value:
             attr_value = get_instance_from_value(
                 attr_type, value[cls.attribute_map[attr]]
@@ -203,7 +202,7 @@ class ClusterSpec(object):
 
     def _patch_service_with_spec(self, service, spec):
         for attr, attr_type in six.iteritems(
-            client.V1ServiceSpec.swagger_types
+            client.V1ServiceSpec.openapi_types
         ):
             if client.V1ServiceSpec.attribute_map[attr] in spec:
                 attr_value = get_instance_from_value(
--- a/elasticdl_client/common/k8s_client.py
+++ b/elasticdl_client/common/k8s_client.py
@@ -11,17 +11,18 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import json
 import os
 import traceback
 
 import yaml
-from kubernetes import client, config
-from kubernetes.client import V1EnvVar, V1EnvVarSource, V1ObjectFieldSelector
-
+from elasticdl_client.common.constants import ClusterSpecConfig
 from elasticdl_client.common.k8s_resource import parse as parse_resource
 from elasticdl_client.common.k8s_volume import parse_volume_and_mount
 from elasticdl_client.common.log_utils import default_logger as logger
 from elasticdl_client.common.module_utils import load_module
+from kubernetes import client, config
+from kubernetes.client import V1EnvVar, V1EnvVarSource, V1ObjectFieldSelector
 
 ELASTICDL_APP_NAME = "elasticdl"
 ELASTICDL_JOB_KEY = "elasticdl-job-name"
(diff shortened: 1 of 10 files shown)


elasticdl_client/common/k8s_client.py
```diff
@@ -47,6 +49,71 @@ def append_pod_ip_to_env(env):
     return env
 
 
+class ClusterSpec(object):
+    def __init__(self, cluster_spec_json="", cluster_spec=""):
```
QiJune, line 107: Since `cluster_spec_json` contains all the information, it seems that there is no need to load `cluster_spec.py`.

The `with_service` and `with_pod` method could be defined in ElasticDL repo.

skydoorkai: This is for legacy support for cluster_spec.py support. 
Also, users may define a model in the repo and submit it to different clusters. No need to modify the repo if using command arguments or external files.

skydoorkai: We may remove cluster_spec.py later if all are switched to command arguments.

skydoorkai:
Verified on the internal k8s system.

We will remove the old argument `cluster_spec` later.

elasticdl_client/common/k8s_client.py
```diff
@@ -216,6 +378,7 @@ def _create_master_pod_obj(self, **kargs):
             volume=kargs["volume"],
             owner_pod=None,
             env=env,
+            pod_type="master",
```
brightcoder01: Can we add a enumeration class for the pod_type?

elasticdl_client/common/k8s_client.py
```diff
@@ -47,6 +50,168 @@ def append_pod_ip_to_env(env):
...
+    try:
+        cls = getattr(module, name)
+        return cls
+    except AttributeError:
+        return None
+
+
+def is_list_type(type_name):
+    if type_name.startswith("list["):
+        return True, type_name.split("[")[1].split("]")[0]
```
brightcoder01, line 69: From the method name `is_list_type `, I think this method only returns true | false instead of a tuple.

elasticdl_client/common/k8s_client.py
skydoorkai: Added.

elasticdl_client/common/k8s_client.py
skydoorkai: Renamed.

elasticdl_client/common/k8s_client.py
```diff
@@ -29,6 +33,21 @@
 ELASTICDL_REPLICA_INDEX_KEY = "elasticdl-replica-index"
 
 
+class PodType(Enum):
```
brightcoder01: Maybe we can declare the enumerated PodType string just like the following way in constants.py

```python
class DistributionStrategy(object):
    LOCAL = "Local"
    PARAMETER_SERVER = "ParameterServerStrategy"
    ALLREDUCE = "AllreduceStrategy"
```

skydoorkai: Done

elasticdl_client/common/k8s_client.py
```diff
@@ -47,6 +56,168 @@ def append_pod_ip_to_env(env):
...
+    #   type_name: a class name in client, or a basic type such as string,
+    #              int, float, etc.
+    #   value: a dict corresponding to the class, or a string/int/float
+    #          value if basic type.
+    # Return: a class instance if class, the basic type value otherwise.
+
+    cls = try_get_class(client, type_name)
+    if not cls:
+        # not a class
+        is_a_list, list_type = check_list_get_type(type_name)
```
QiJune, line 85: Why should we check the list type here? Could you please give an example?

elasticdl_client/common/k8s_client.py
```diff
@@ -47,6 +56,168 @@ def append_pod_ip_to_env(env):
...
+        cls = getattr(module, name)
+        return cls
+    except AttributeError:
+        return None
+
+
+def check_list_get_type(type_name):
+    if type_name.startswith("list["):
+        return True, type_name.split("[")[1].split("]")[0]
+    else:
```
QiJune, line 70: We could remove this `else`.

elasticdl_client/common/k8s_client.py
```diff
@@ -47,6 +56,168 @@ def append_pod_ip_to_env(env):
...
+        return False, None
+
+
+def get_instance_from_value(type_name, value):
```
QiJune, line 74: We may add more unit tests for this method later.

QiJune (approved):
LGTM

elasticdl_client/common/k8s_client.py
QiJune: A reference: https://github.com/kubernetes-client/python/blob/bc79976edfe01b028aef0b93bde54e002b113443/kubernetes/client/models/v1_node_selector.py#L36

elasticdl_client/common/k8s_client.py
```diff
@@ -47,6 +56,168 @@ def append_pod_ip_to_env(env):
...
 
+def try_get_class(module, name):
+    try:
+        cls = getattr(module, name)
+        return cls
+    except AttributeError:
+        return None
+
+
+def check_list_get_type(type_name):
```
brightcoder01, line 67: How about separate this method into two.
is_list_type
get_list_element_type

elasticdl_client/common/k8s_client.py
```diff
@@ -33,6 +36,12 @@ def get_master_pod_name(job_name):
     return "elasticdl-%s-master" % job_name
 
 
+class PodType(object):
```
brightcoder01, line 39: We can move this class into constants.py

PR #2328, opened 2020-10-19
Rename custom_model to forward

workingloong:


workingloong:
The name needs more consideration.

PR #2333, opened 2020-10-29
Support PyTorch to customize training loop.

workingloong:
The design for the custom training loop is in https://github.com/sql-machine-learning/elasticdl/blob/develop/docs/designs/custom_allreduce_training_loop_support.md


model_zoo/mnist/mnist_train_torch.py
```diff
@@ -0,0 +1,121 @@
...
+        "label": tf.io.FixedLenFeature([1], tf.int64),
+    }
+    r = tf.io.parse_single_example(record, feature_description)
+    features = {
+        "image": tf.math.divide(tf.cast(r["image"], tf.float32), 255.0)
+    }
+    return features["image"], tf.cast(r["label"], tf.int32)
+
+
+def eval_metrics_fn():
```
skydoorkai: pytorch does not support evaluation. Thus this is not needed at least for now.

elasticdl/python/tests/allreduce_trainer_test.py
```diff
@@ -100,17 +103,56 @@ def test_elastic_run(self):
...
-        controller.set_broadcast_optimizer(tf.optimizers.SGD(0.01))
+        controller = PyTorchAllReduceController(master_client, "")
+        model = TorchModel()
+        optimizer = optim.SGD(model.parameters(), lr=0.1)
+        hvd.init()
+        optimizer = hvd.DistributedOptimizer(optimizer)
+        controller.set_broadcast_model(model)
+        controller.set_broadcast_optimizer(optimizer)
         controller.broadcast()
+        controller.restore()
```
skydoorkai: Why put `restore` here, after `broadcast` before `train_one_batch_with_retries`?

skydoorkai (approved):

PR #2342, opened 2020-11-11
[WIP] Demo for Pytorch to keep batch size

workingloong:


workingloong:
Close the PR because we have sumitted the code in #2389

PR #2345, opened 2020-11-13
Master exits if all worker completed

### LINKED ISSUE #2262: Master should exit if all workers completed
image

workingloong:
Fix #2262

### DIFF
--- a/elasticdl/python/master/master.py
+++ b/elasticdl/python/master/master.py
@@ -236,6 +236,10 @@ class Master(object):
                             InstanceManagerStatus.FINISHED
                         )
                     break
+                if self.instance_manager.all_workers_failed:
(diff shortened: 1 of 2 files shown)


elasticdl/python/master/master.py
```diff
@@ -236,6 +236,11 @@ def run(self):
                             InstanceManagerStatus.FINISHED
                         )
                     break
+                if self.instance_manager.all_workers_completed:
```
skydoorkai: What will happen if half workers completed, the other half failed?

workingloong: We can fix the logic. The worker exits if completed or failed and the master exits if all workers exit.

skydoorkai (approved):

PR #2346, opened 2020-11-13
Delete the WAIT task in the generator of the training dataset.

workingloong:
The logic is complex to create the dataset for the training loop. We don't need waiting tasks in the generator of the dataset.

### DIFF
--- a/elasticdl/python/master/task_dispatcher.py
+++ b/elasticdl/python/master/task_dispatcher.py
@@ -365,7 +365,7 @@ class _TaskDispatcher(object):
 
     def finished(self):
         """Return if all tasks are done"""
-        return all([not self._todo, not self._eval_todo, not self._doing])
+        return all([not self._todo, not self._doing])
 
     def recover_tasks(self, worker_id):
         """Recover doing tasks for a dead worker"""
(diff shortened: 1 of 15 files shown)


elasticdl/python/master/task_dispatcher.py
```diff
@@ -365,7 +365,7 @@ def check_exceed_max_task_retries(self, task):
 
     def finished(self):
         """Return if all tasks are done"""
-        return all([not self._todo, not self._eval_todo, not self._doing])
```
brightcoder01, line 368: Will we have `_eval_todo` anymore?

elasticdl/python/worker/task_data_service.py
```diff
@@ -169,69 +171,24 @@ def get_dataset(self):
...
-                    logger.info("No tasks for now, maybe more later")
-                    # There are too many requests to get task from the master
-                    # if the worker does not sleep.
-                    time.sleep(5)
-                else:
-                    logger.info("No more task, stopping")
                 break
             with self._lock:
                 if task.type == elasticdl_pb2.TRAIN_END_CALLBACK:
                     self._pending_train_end_callback_task = task
```
brightcoder01, line 190: Will we have more than one TRAIN_END_CALLBACK tasks?
If we have break here, will we miss the other train end callback tasks?

workingloong: We will not have other tasks after the training.

skydoorkai (commented):
gltm as long as training tests are passed.

skydoorkai (approved):

brightcoder01 (approved):

PR #2347, opened 2020-11-13
The worker only create a dataset for an evaluation job

workingloong:
Now, the worker needs to create a dataset for each evaluation task. It will create much overhead and evenly result in OOM. The worker only creates a dataset that can query evaluation tasks.

### DIFF
--- a/elasticdl/python/worker/worker.py
+++ b/elasticdl/python/worker/worker.py
@@ -262,51 +262,33 @@ class Worker(object):
         self._timing.end_record_time("batch_process")
         return min_model_version
 
-    def _process_eval_task(self, task):
+    def _process_evaluation_dataset(self, dataset):
         """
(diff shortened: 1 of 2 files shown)


elasticdl/python/worker/worker.py
```diff
@@ -262,51 +262,33 @@ def _process_minibatch(
         self._timing.end_record_time("batch_process")
         return min_model_version
 
-    def _process_eval_task(self, task):
+    def _process_evaluation_dataset(self, dataset):
```
skydoorkai: To better describe this function:
`_process_evaluation_dataset` -> `_process_evaluation_if_exist`.

`evaluation_executed` -> `evaluation_exist`.

skydoorkai (approved):
LGTM as long as it has been tested.

PR #2348, opened 2020-11-16
[WIP][Discussion] Ensure the global batch size is not changed even the horovod size changes.

brightcoder01:
Please use the following command the do the verification.

```SHELL
horovodrun -np 3 -H localhost:3 python model_zoo/mnist/mnist_train_tfv1_fixed_global_size.py
```

The variable `backward_passes_per_step` of the first two workers is 3 and the value of the 3rd worker is 2.

### DIFF
--- a/model_zoo/mnist/customized_optimizers.py
+++ b/model_zoo/mnist/customized_optimizers.py
@@ -15,9 +15,9 @@ import horovod.tensorflow as hvd
 import tensorflow as tf
 
 
-class UpdateBackwardNumberByHorovodSizeHook(tf.train.SessionRunHook):
+class AdjustBackwardPassesPerStepHook(tf.train.SessionRunHook):
     def __init__(self, backward_passes_per_step_var, global_batch_count_per_step):
-        
+
         self._step = 0
         self._value_placeholder = tf.placeholder(
             backward_passes_per_step_var.dtype, []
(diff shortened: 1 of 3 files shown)


model_zoo/mnist/customized_optimizers.py
```diff
@@ -0,0 +1,376 @@
+import horovod.tensorflow as hvd
+import tensorflow as tf 
+
+from tensorflow.python.training import optimizer
+
+
+class UpdateVariableBySizeHook(tf.train.SessionRunHook):
+    def __init__(self, updatable_tensor, total_count):
```
workingloong: Add a docstring to describe the `total_count` and `updatable_tensor`.

model_zoo/mnist/customized_optimizers.py
```diff
@@ -0,0 +1,376 @@
...
+            name = "Distributed{}".format(type(optimizer).__name__)
+        super(ElasticDistributedOptimizer, self).__init__(name=name, use_locking=use_locking)
+
+        self._optimizer = optimizer
+        self._allreduce_grads = hvd._make_allreduce_grads_fn(
+            name, device_dense, device_sparse, compression, sparse_as_dense, op,
+            gradient_predivide_factor)
+
+        self._agg_helper = None
+        if backward_passes_per_step > 1:
```
workingloong: In the elastic training, the `backward_passes_per_step` need to change. So, should we use the `agg_helper` defaultly?

brightcoder01: Good idea, updated.

model_zoo/mnist/customized_optimizers.py
brightcoder01: Good idea. Docstring is added and do some renaming for readability.

brightcoder01:
The official solution is submitted in PR #2409 2409

PR #2351, opened 2020-11-17
Move the ps client initialization into a separate method in the main.py of worker.

brightcoder01:


elasticdl/python/worker/main.py
```diff
@@ -25,6 +25,46 @@
 CONNECT_PS_TIMEOUT = 300
 
 
+def build_ps_client(ps_addrs, logger):
```
workingloong: Should we move the function into ps_client.py?

workingloong: Add a docstring to describe the ps_addr.

brightcoder01: Good idea, updated.

skydoorkai (approved):

PR #2353, opened 2020-11-18
Polish code for master and common package.

brightcoder01:
- Do some function/variable renaming for readability.
- Separate some long function into several short ones.


skydoorkai (approved):
gltm

PR #2357, opened 2020-11-18
Use max_step argument to replace MaxStepStopping callback

workingloong:


elasticdl/python/master/task_dispatcher.py
```diff
@@ -102,18 +101,21 @@ def __init__(
             records_per_task: The number of records per task.
             num_epochs: The total number of epochs for the tasks where
                 an epoch is a complete iteration over the shards.
-            callbacks_list: The Keras CallbacksList object to contain all
-                callback instances.
+            batch_size:
```
skydoorkai: Also add brief descriptions for these two arguments.

elasticdl/python/master/task_dispatcher.py
```diff
@@ -346,12 +354,20 @@ def report(self, request, success):
...
                 if task in self._task_retry_count:
                     del self._task_retry_count[task]
-                if self._callbacks_list.model.stop_training:
-                    # Clear todo list to stop training
-                    self._todo = []
 
         return (time.time() - start_time), task, worker_id
 
+    def _check_exceed_max_step(self, task):
+        if self._max_step <= 0:
```
skydoorkai: `if self._max_step <=  0 or task.type != elasticdl_pb2.TRAINING:`

so we can omit the second if later.

workingloong: `if self._max_step > 0 and task.type == elasticdl_pb2.TRAINING`

skydoorkai (approved):

PR #2359, opened 2020-11-19
Separate data shard service from task data service

workingloong:


elasticdl/python/worker/data_shard_service.py
```diff
@@ -0,0 +1,108 @@
...
+        self._reported_record_count = 0
+        self._current_task = None
+        self._pending_tasks = deque()
+
+    def get_current_task(self):
+        return self._current_task
+
+    def get_task(self, task_type=None):
+        task = self._mc.get_task(task_type)
+        if task.shard.name and task.type == elasticdl_pb2.TRAINING:
```
skydoorkai, line 39: We may not depend on `task.shard.name` to determine if there are new tasks or not.

workingloong: Yes, I add a todo to use task type STOP to determine whether there are new tasks or not. And I will commit the modification in the next PR.

skydoorkai (approved):

PR #2369, opened 2020-11-23
Polish k8s_instance_manager.

brightcoder01:
- Polish the current code of k8s_instance_manager at first. And then we will do some refactor in the next step.
- Do function renaming; add more descriptions; extract utility functions to make the function body short.

### DIFF
--- a/elasticdl/python/master/k8s_instance_manager.py
+++ b/elasticdl/python/master/k8s_instance_manager.py
@@ -29,7 +29,7 @@ _SERVICE_ADDR_SEP = ","
 def _get_addrs(num_addrs, addr_get_fn):
     """
     Get `num_addrs` addresses and the concatenate
-    them to a common separated string.
+    them to a comma separated string.
     """
     addrs = []
     for addr_id in range(num_addrs):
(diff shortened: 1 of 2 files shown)


workingloong (dismissed):
Excellent!

elasticdl/python/master/k8s_instance_manager.py
```diff
@@ -51,6 +58,17 @@ def _parse_worker_pod_priority(num_workers, worker_pod_priority):
     return res
 
 
+def _is_pod_oom_killed(evt_obj):
```
skydoorkai: This is not `_is_pod_oom_killed`, but `killed_but_not_oomed`.

brightcoder01: Good catch, the method name is updated.

elasticdl/python/master/k8s_instance_manager.py
```diff
@@ -26,6 +26,17 @@
 _SERVICE_ADDR_SEP = ","
 
 
+def _get_addrs(num_addrs, addr_get_fn):
+    """
+    Get `num_addrs` addresses and the concatenate
```
workingloong: and then ?

brightcoder01: Good catch, updated.

workingloong (approved):

skydoorkai (approved):

PR #2372, opened 2020-11-23
Move the code to check the timeout task into task manager

workingloong:


elasticdl/python/master/task_manager.py
```diff
@@ -448,5 +463,70 @@ def set_evaluation_service(self, evaluation_service):
...
+        self._task_completed_times[task_type].append(completed_time)
+
+    def set_task_timeout_callback(self, callback_fn):
+        self._task_timeout_callbacks.append(callback_fn)
+
+    def _invoke_task_timeout_callback(self, worker_id):
+        for callback_fn in self._task_timeout_callbacks:
+            callback_fn(worker_id)
+
+    def _check_and_reassign_timeout_tasks(self, default_period=30):
```
skydoorkai: When will `default_period` used when calling `_check_and_reassign_timeout_tasks`?

elasticdl/python/tests/task_manager_test.py
```diff
@@ -132,6 +134,36 @@ def test_invoke_train_end_callback(self):
...
 
+    def test_check_and_reassign_timeout_tasks(self):
+        task_manager = create_task_manager({"f1": (0, 10), "f2": (0, 10)}, {})
+        task_manager.create_tasks(elasticdl_pb2.TRAINING)
+        task_count = len(task_manager._todo)
+        task_start_time = time.time() - 1000
+        task_manager._worker_start_task_time[0] = task_start_time
+        task_manager._doing[0] = (0, task_manager._todo[0], task_start_time)
+        task_manager._todo.pop()
+
```
brightcoder01, line 145: We can use the following way in unit test to make sure the test always passes. Our goal is to check the the logic correctness of `_check_and_reassign_timeout_tasks`.
If we create a thread, there may be a rare case that the background thread is not scheduled.
```
time.sleep(1)
task_manager._check_and_reassign_timeout_tasks()
```

elasticdl/python/master/task_manager.py
```diff
@@ -448,5 +463,67 @@ def set_evaluation_service(self, evaluation_service):
...
+                        logger.info(
+                            "worker %d timeout, relaunch it" % worker_id
+                        )
+                        self.recover_tasks(worker_id)
+                        # TODO: save worker logs before remove it
+                        self._invoke_task_timeout_callback(worker_id)
+                        break
+            time.sleep(30)
+
+    def _get_average_task_completed_time(self):
```
brightcoder01, line 509: We can add a description that the return value is a dict containing the average time for TRAINING and EVALUATION tasks.

elasticdl/python/tests/task_manager_test.py
workingloong: But, there is an infinite loop in the `task_manager._check_and_reassign_timeout_tasks()`.  If we call `task_manager._check_and_reassign_timeout_tasks()`in the unittest and the unittest will not exit.

skydoorkai (approved):

PR #2377, opened 2020-11-24
Set worker_id and master_addr into the envs of worker pod.

workingloong:


elasticdl/python/worker/main.py
```diff
@@ -23,13 +26,13 @@
 def main():
     args = parse_worker_args()
     logger = log_utils.get_logger(__name__)
-    logger.info("Starting worker %d", args.worker_id)
-    if args.master_addr is None:
+    worker_id = int(os.getenv(WorkerEnv.WORKER_ID))
```
skydoorkai: Add a default for worker_id to avoid a crash.

skydoorkai (approved):

PR #2379, opened 2020-11-25
Refactor master servicer

workingloong:


skydoorkai (approved):
After we move the servicer out of ElasticdlJobService, we can pass `ElasticdlJobService ` instead of `evaluation_service` for MasterServer init.

We can do it later.

PR #2381, opened 2020-11-26
Update base image and add more logs in the building process.

brightcoder01:
At the present, the base image is `tensorflow/tensorflow:2.1.0-py3`, but the requirement of elasticdl is tensorflow.2.1.2, it will cost much time to download and reinstall tensorflow.2.1.2. It would be faster to update the base image.

### DIFF
--- a/scripts/travis/build_images.sh
+++ b/scripts/travis/build_images.sh
@@ -14,7 +14,7 @@
 
 # Pull base image quietly.  We might want to use
 # tensorflow/tensorflow:2.1.0-gpu-py3 for GPU-accelerated deep learning.
-BASE_IMAGE=tensorflow/tensorflow:2.1.0-py3
+BASE_IMAGE=tensorflow/tensorflow:2.1.2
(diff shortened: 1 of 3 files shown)


scripts/travis/build_images.sh
```diff
@@ -14,7 +14,7 @@
 
 # Pull base image quietly.  We might want to use
 # tensorflow/tensorflow:2.1.0-gpu-py3 for GPU-accelerated deep learning.
-BASE_IMAGE=tensorflow/tensorflow:2.1.0-py3
+BASE_IMAGE=tensorflow/tensorflow:2.1.2
```
workingloong, line 17: What is the difference between 2.1.2 and 2.1.0-py3?

skydoorkai (approved):

scripts/travis/build_images.sh
brightcoder01: The pre-installed tensorflow version are different.
CI image building will be much faster if we use 2.1.2 because elasticdl pkg requires 2.1.2

PR #2383, opened 2020-11-27
refine loop run

skydoorkai:


elasticdl/python/master/master.py
```diff
@@ -52,19 +52,30 @@ def run(self):
...
-                            InstanceManagerStatus.FINISHED
-                        )
-                    break
-                if self.pod_manager.all_workers_exited:
-                    raise Exception(
-                        "All workers exited but there also are",
-                        "unfinished tasks",
-                    )
-                time.sleep(30)
+            no_loop_run = not self.task_manager and not self.pod_manager
```
brightcoder01: I'm wondering whether the scenario exists that neither task_manager nor pod_manager is created.

skydoorkai: When users set arguments wrong.

brightcoder01: Maybe we can do some argument check and raise ValueError directly.

skydoorkai: Maybe we should assume that when it `run`, it should satisfy that at least one manager is running. 
Then we need to add an argument validation in `init` or `prepare`.

workingloong (dismissed):

elasticdl/python/master/master.py
```diff
@@ -128,3 +130,23 @@ def create_master_grpc_service(self, args):
             self.rendezvous_server,
             evaluation_service,
         )
+
+    def validate(self):
```
workingloong, line 134: Add an unit test for the method.

elasticdl/python/tests/master_test.py
```diff
@@ -65,3 +65,34 @@ def test_master_run_and_stop(self):
...
+                master.validate()
+
+            self.arguments["need_elasticdl_job_service"] = "False"
+            args = self._get_args()
+            args = parse_master_args(args)
+            master = Master(args)
+            master.validate()
+
+            self.arguments["task_fault_tolerance"] = "False"
+            self.arguments["task_fault_tolerance"] = "True"
```
brightcoder01, line 91: Why do wen set `task_fault_tolerance` twice?

skydoorkai: Copy&Paste bug. Need to delete the 1st one.

workingloong (approved):

PR #2389, opened 2020-11-30
Develop an optimizer to keep the batch size fixed.

### LINKED ISSUE #2388: Develop an torch.optimizer to support fixed batch size when the worker number changes.
```python
import os
import warnings
from contextlib import contextmanager

import torch
from horovod.torch.compression import Compression
from horovod.torch.mpi_ops import Average, allreduce_async_, size, synchronize


class _ElasticDistributedOptimizer(torch.optim.Optimizer):
    def __init__(
        self,
        params,
        named_parameters=None,
        compression=Compression.none,
        backward_passes_per_step=1,
        op=Average,
        gradient_predivide_factor=1.0,
        batch_num_per_step=None,
        fixed_batch_size=False,
    ):
        super(self.__class__, self).__init__(params)
       ...

        self._fixed_batch_size = fixed_batch_size
        self._batch_num_per_step = batch_num_per_step
        self._iter_step = 0

    def load_state_dict(self, *args, **kwargs):
        self._handles = {}
        self._synchronized = False
        self._should_synchronize = True
        for p in self._allreduce_delay:
            self._allreduce_delay[p] = self.backward_passes_per_step
        super(self.__class__, self).load_state_dict(*args, **kwargs)

    def _allreduce_grad_async(self, p):
        name = self._parameter_names.get(p)
        tensor = p.grad
        tensor_compressed, ctx = self._compression.compress(tensor)

        if self.op == Average:
            # Split average operation across pre/postscale factors
            # C++ backend will apply additional 1 / size() factor
            # to postscale_factor for op == Average.
            if self._fixed_batch_size:
                prescale_factor = 1.0 / (
                    self.gradient_predivide_factor * self._batch_num_per_step
                )
                postscale_factor = self.gradient_predivide_factor * size()
            else:
                prescale_factor = 1.0 / self.gradient_predivide_factor
                postscale_factor = self.gradient_predivide_factor * size()
        else:
            prescale_factor = 1.0
            postscale_factor = 1.0

        handle = allreduce_async_(
            tensor_compressed,
            name=name,
            op=self.op,
            prescale_factor=prescale_factor,
            postscale_factor=postscale_factor,
        )
        return handle, ctx

    def step(self, closure=None):
        self._iter_step += 1
        if (
            self._fixed_batch_size
            and self._iter_step % self.backward_passes_per_step != 0
        ):
            self._update_gradients = False
        else:
            self._update_gradients = True

        if not self._update_gradients:
            return
        return super(self.__class__, self).step(closure)

    def zero_grad(self):
        if not self._update_gradients:
            return
       return super(self.__class__, self).zero_grad()


def ElasticDistributedOptimizer(
    optimizer,
    named_parameters=None,
    compression=Compression.none,
    backward_passes_per_step=1,
    op=Average,
    gradient_predivide_factor=1.0,
    batch_num_per_step=None,
    fixed_batch_size=False,
):

    cls = type(
        optimizer.__class__.__name__,
        (optimizer.__class__,),
        dict(_ElasticDistributedOptimizer.__dict__),
    )
    return cls(
        optimizer.param_groups,
        named_parameters,
        compression,
        backward_passes_per_step,
        op,
        gradient_predivide_factor,
        batch_num_per_step,
        fixed_batch_size,
    )
```

## Why does Horovod use `type` to create a class base `optimizer.__class__`?
Each optimizer in Pytorch implements its `step` function. So, Horovod needs to create an optimizer with the `step` of the original optimizer.  `type` can create an optimizer class inherited`optimizer.__class__` with methods of `_ElasticDistributedOptimizer`. The new class will have the `step` and methods of `_ElasticDistributedOptimizer`. If we don't use `type` to create a new class, we need to write distributed optimizers inherited optimizers in Pytorch.

workingloong:
fix #2388


elasticdl/python/allreduce/torch_optimizer.py
```diff
@@ -0,0 +1,294 @@
...
+import os
+import warnings
+from contextlib import contextmanager
+
+import torch
+from horovod.torch.compression import Compression
+from horovod.torch.mpi_ops import Average, allreduce_async_, size, synchronize
+
+
+class _ElasticDistributedOptimizer(torch.optim.Optimizer):
```
brightcoder01, line 22: Please add the permalink of horovod code snippet which we refer to.

elasticdl/python/allreduce/pytorch_controller.py
```diff
@@ -75,4 +81,26 @@ def restore(self):
...
+        self.backward_passes_per_step = int(
+            self.batch_num_per_step / world_size
+        )
+        if rank < self.batch_num_per_step % world_size:
+            self.backward_passes_per_step += 1
+        if (
+            self.backward_passes_per_step
+            != self._optimizer.backward_passes_per_step
+        ):
+            self._optimizer.backward_passes_per_step = (
```
brightcoder01, line 149: Will the internal optimizer such as SGD/Adam have the property `backward_passes_per_step`?

workingloong: Yes. The class of optimizer is the subclass of `SGD/Adam` and has attributes of `_ElasticDistributedOptimizer`.  
```python
cls = type(
        optimizer.__class__.__name__,
        (optimizer.__class__,),
        dict(_ElasticDistributedOptimizer.__dict__),
    )
```
The class generated by the code will have attributes and methods of `_ElasticDistributedOptimizer`.

elasticdl/python/allreduce/torch_optimizer.py
workingloong: Good suggestion!

elasticdl/python/allreduce/pytorch_controller.py
brightcoder01: I see. I thought it's the internal native optimzer inside _ElasticDistributedOptimizer. My mistake.

elasticdl/python/allreduce/pytorch_controller.py
```diff
@@ -75,4 +81,26 @@ def restore(self):
         time.sleep(3)
         # Call `load_state_dict` to reset the state of Horovod optimizer
         self._optimizer.load_state_dict(self._optimizer.state_dict())
+        self._optimizer.zero_grad()
         self._rendezvous_manager.init_horovod_if_needed()
+
+    def reset_backward_passes_per_step(self):
```
brightcoder01, line 137: So this method plays the same role of AdjustBackwardPassesPerStepHook in TF1.x?

workingloong: Yes.

brightcoder01 (dismissed):

brightcoder01 (approved):

PR #2390, opened 2020-11-30
Support creating shards for Text files

workingloong:


elasticdl/python/data/reader/text_reader.py
```diff
@@ -0,0 +1,83 @@
...
+            start_ind += records_per_task
+        num_records_left = size % records_per_task
+        if num_records_left != 0:
+            shards[shard_name_prefix + str(num_shards)] = (
+                start_ind,
+                num_records_left,
+            )
+        return shards
+
+    def get_size(self):
```
skydoorkai, line 60: If there is a blank line at the end of the file,  this also counts as a line?

workingloong: Hmm. It is difficult to check whether each line in the file is valid. We should tell users the rules if they use this reader.

elasticdl/python/data/reader/text_reader.py
```diff
@@ -0,0 +1,83 @@
...
+    Metadata,
+)
+
+
+class TextDataReader(AbstractDataReader):
+    """This reader is used to create shards for a file and
+    read records from the shard.
+    """
+
+    def __init__(self, **kwargs):
```
brightcoder01: We can write it as `__init__(self, filename, records_per_task, **kwargs)` directly.

elasticdl/python/data/reader/text_reader.py
```diff
@@ -0,0 +1,83 @@
...
+            task.shard.start : task.shard.end
+        ]
+        if shuffle:
+            np.random.shuffle(records)
+
+        for record in records:
+            yield record
+
+    def create_shards(self):
+        shard_name_prefix = self._kwargs["filename"] + ":shard_"
```
brightcoder01: What if the csv file has headers?

workingloong: There are many limits if we use a CSV file to store the meta information of images. Users need to pay attention to the header to create dataset if the CSV file has a header line. I think we should give users some rules to create the CSV file.

brightcoder01 (approved):

PR #2391, opened 2020-11-30
Decouple PodManager with TaskManager and RendezvousServer using PodEventCallbacks.

brightcoder01:
- Implement two subclass of PodEventCallback: `TaskRescheduleCallback` and `RendezvousServiceRefreshCallback`
- Inside PodManager._event_cb, we will call PodEventCallback instead of calling TaskManager and RendezvousServer directly
- Remove task_manager and rendezvous_server from the constructor of PodManager.

### DIFF
--- a/elasticdl/python/master/pod_manager.py
+++ b/elasticdl/python/master/pod_manager.py
@@ -430,22 +430,6 @@ class PodManager(object):
             if pod_name in self._failed_pods:
                 return
 
-            # Notify each PodEventCallback that PodStarted is fired
-            if evt_type in ["ADDED", "MODIFIED"] and phase == "Running":
-                for callback in self._pod_event_callbacks:
-                    callback.on_pod_started(
-                        PodInfo(type=None, id=None, name=pod_name),
-                        ClusterContext(pod_manager=self),
-                    )
-
-            # Notify each PodEventCallback that PodSucceeded is fired
-            if evt_type == "MODIFIED" and phase == "Succeeded":
-                for callback in self._pod_event_callbacks:
-                    callback.on_pod_succeeded(
-                        PodInfo(type=None, id=None, name=pod_name),
-                        ClusterContext(pod_manager=self),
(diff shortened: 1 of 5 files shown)


elasticdl/python/master/pod_manager.py
```diff
@@ -430,15 +420,28 @@ def _event_cb(self, event):
             if pod_name in self._failed_pods:
                 return
 
+            # Notify each PodEventCallback that PodStarted is fired
+            if evt_type in ["ADDED", "MODIFIED"] and phase == "Running":
+                for callback in self._pod_event_callbacks:
+                    callback.on_pod_started(
+                        PodInfo(type=None, id=None, name=pod_name),
```
workingloong: What's the mean of `type=None`?

brightcoder01: The type can be either WORKER or PS. At the present, we don't care which one it is. So set it to None.

elasticdl/python/master/pod_manager.py
```diff
@@ -430,15 +420,28 @@ def _event_cb(self, event):
             if pod_name in self._failed_pods:
                 return
 
+            # Notify each PodEventCallback that PodStarted is fired
+            if evt_type in ["ADDED", "MODIFIED"] and phase == "Running":
```
skydoorkai: `ADDED` can be removed here, as a pod is always MODIFIED from pending to running. `ADDED` with running only happens without pod status change.
Since we will track pod status change later, this can leave to the next PR.

skydoorkai (dismissed):

elasticdl/python/master/pod_manager.py
brightcoder01: If the image is cached in the Node, we can see event `ADDED` with phase `RUNNING` directly.

skydoorkai (approved):

workingloong (approved):

PR #2386, opened 2020-11-27
Decouple the creation of PodManager from ElasticDLJobService

brightcoder01:
This is the first step of PodManager refactor. There will be three in total.

- Decouple PodManager creation from ElasticDLJobService.
- Decouple PodManager with TaskManager and RendezvousServer using PodEventCallbacks.
- More Polish on the __init__ parameter building and the composition among different components.

PR #2393, opened 2020-11-30
Support custom pod command

workingloong:


elasticdl/python/master/master.py
```diff
@@ -46,20 +46,7 @@ def prepare(self):
...
-                        self._args
-                    ),
-                    ps_command=command,
-                    ps_args=self.elasticdl_job_service.get_ps_args(self._args),
-                )
-            else:
-                # TODO: Get the Pod arguments from the input
-                # args directly
-                pass
+            self._set_pod_manager()
```
skydoorkai: `_set_pod_manager` name does not show what this function is doing.
`set_command_in_pod_manager` ?

elasticdl/python/master/master.py
```diff
@@ -76,6 +63,30 @@ def prepare(self):
...
+            self.pod_manager.set_up(
+                worker_command=command,
+                worker_args=self.elasticdl_job_service.get_worker_args(
+                    self._args
+                ),
+                ps_command=command,
+                ps_args=self.elasticdl_job_service.get_ps_args(self._args),
+            )
+        elif self._args.job_command:
+            self.pod_manager.set_up(
```
skydoorkai, line 78: Also, `set_up` can rename to `set_pod_command`.

elasticdl/python/master/master.py
workingloong: Done!

elasticdl/python/master/master.py
workingloong: You can see the discussion about the function name. https://github.com/sql-machine-learning/elasticdl/pull/2386#discussion_r531547517

elasticdl/python/master/master.py
```diff
@@ -76,6 +63,30 @@ def prepare(self):
...
+                worker_command=["/bin/bash"],
```
brightcoder01, line 79: I'm not sure that whether the customized job_command always starts with "/bin/bash -c"

workingloong: I think the job command should be a shell command and can use "/bin/bash -c" to execute the shell command.  What's more, the command in the container must be `["/bin/bash", "-c", "python -m xxxx"]` not `["/bin/bash -c python -m xxxx"]`. Because the first string the command list must be the path of an executable file.

skydoorkai (approved):

brightcoder01 (approved):

PR #2400, opened 2020-12-03
Set the priority of at least 1 worker to high if the priority is a fraction.

workingloong:
The master will exit if there is no worker. So we must guarantee at least 1 worker will not be preempted.

### DIFF
--- a/elasticdl/python/master/pod_manager.py
+++ b/elasticdl/python/master/pod_manager.py
@@ -52,6 +52,8 @@ def _get_addrs(num_addrs, addr_get_fn):
 
 
 def _is_float_str(str_number):
+    if not str_number:
+        return False
     try:
         float(str_number)
         return True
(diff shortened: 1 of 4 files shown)


brightcoder01 (dismissed):

elasticdl/python/master/pod_manager.py
```diff
@@ -51,28 +51,36 @@ def _get_addrs(num_addrs, addr_get_fn):
...
-                "low priority for all worker pods by default."
-            )
-            for i in range(num_workers):
-                res[i] = None
-    else:
+    if _is_float_str(worker_pod_priority):
+        fraction = float(worker_pod_priority)
+        high_count = int(num_workers * fraction)
+        # At least 1 worker has high priority
+        high_count = 1 if high_count < 1 else high_count
```
skydoorkai, line 70: The logic is reversed?

skydoorkai: if high_count < 1 and high_count  > 0

workingloong: `math.ceil()` is good.

skydoorkai (approved):

PR #2402, opened 2020-12-04
Record the start running time stamp in PodInfo and sort the live workers according to this property for allreduce.

brightcoder01:


skydoorkai (approved):
gltm

PR #2404, opened 2020-12-07
Set distributed strategy to AllReduce when PS number is 0

workingloong:


brightcoder01 (dismissed):

elasticdl/python/common/args.py
```diff
@@ -159,6 +159,9 @@ def _check_master_args_validity(args):
         logger.warning(
             "grads_to_wait is set to 1 while using asynchronous SGD."
         )
+    if args.num_ps_pods == 0:
+        logger.warning("AllReduce strategy does not need PS")
```
brightcoder01: if args.num_ps_pods == 0 && args.distribution_strategy != DistributionStrategy.ALLREDUCE:
    logger.warning("Set the distribution strategy to be AllReduce if ps number is 0.")

workingloong: Yeah, fixed it.

skydoorkai (approved):

PR #2407, opened 2020-12-08
Set the default type number of task to NONE

workingloong:
There is no more task if the worker gets a task with type = -1.

### DIFF
--- a/elasticdl/python/master/servicer.py
+++ b/elasticdl/python/master/servicer.py
@@ -111,7 +111,13 @@ class MasterServicer(elasticdl_pb2_grpc.MasterServicer):
             # we are trying to pop and invoke the callback.
             # Then the master tells the worker to wait
             # in case of new tasks later.
-            res.type = elasticdl_pb2.WAIT
+            if self._rendezvous_server:
+                # If there is no more task, master only send wait task to
+                # the last worker and other workers exit.
+                if len(self._instance_manager.get_alive_workers()) == 1:
+                    res.type = elasticdl_pb2.WAIT
+            else:
+                res.type = elasticdl_pb2.WAIT
         with self._lock:
             self._task_manager.reset_worker_start_task_time(request.worker_id)
         return res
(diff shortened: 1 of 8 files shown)


elasticdl/python/worker/task_data_service.py
```diff
@@ -114,12 +114,8 @@ def _gen(self):
         """
         while True:
             task = self._data_shard_service.get_task()
-            if not task.shard.name:
+            if task.type != elasticdl_pb2.TRAINING:
                 break
-            with self._lock:
-                if task.type == elasticdl_pb2.TRAIN_END_CALLBACK:
-                    self._pending_train_end_callback_task = task
-                    break
```
workingloong, line 122: Here, we don't need to check whether the task is `TRAIN_END_CALLBACK`. Because the task must be `WAIT` after Training if there are also training tasks is not completed. The task will be `TRAIN_END_CALLBACK` only after all training tasks are completed. 
https://github.com/sql-machine-learning/elasticdl/blob/d7d081b43ae05525d83d73aa6d430ececfcf5ba5/elasticdl/python/master/servicer.py#L105-L119

After the training loop is completed, the worker will get the `TRAIN_END_CALLBACK` task in codes:

https://github.com/sql-machine-learning/elasticdl/blob/d7d081b43ae05525d83d73aa6d430ececfcf5ba5/elasticdl/python/worker/task_data_service.py#L80-L94

elasticdl/python/master/servicer.py
```diff
@@ -85,6 +85,7 @@ def get_task(self, request, _):
         shard = elasticdl_pb2.Shard()
         res = elasticdl_pb2.Task(shard=shard)
         res.model_version = self._version
+        res.type = -1
```
brightcoder01: We can add `NONE` or `INVALID` in enum `TaskType` in the `elasticdl.proto`

brightcoder01 (dismissed):

brightcoder01 (approved):

PR #2408, opened 2020-12-09
Set pipefail for the worker command

workingloong:


elasticdl/python/master/pod_manager.py
```diff
@@ -352,16 +352,20 @@ def _process_worker(self):
...
+            job_command += " --ps_addrs {}".format(self._ps_addrs)
         if self._log_file_path:
-            bash_command += BashCommandTemplate.REDIRECTION.format(
+            job_command += BashCommandTemplate.REDIRECTION.format(
                 self._log_file_path
             )
         for extra_arg in self._worker_args[2:]:
-            bash_command += " {}".format(extra_arg)
-        worker_args = [self._worker_args[0], bash_command]
+            job_command += " {}".format(extra_arg)
```
brightcoder01: `job_command += " ".join(self._worker_args[2:])` is simpler.

elasticdl/python/master/pod_manager.py
```diff
@@ -352,16 +352,20 @@ def _process_worker(self):
 
     def _start_worker(self, worker_id):
         logger.info("Starting worker: %d" % worker_id)
-        bash_command = self._worker_args[1]
+        # self._worker_args has 2 strings. The first string is "-c" and
+        # the second string is the shell command to run, like
+        # ["-c", "python -m elasticdl.python.worker.main --minibatch_size 64"]
+        job_command = self._worker_args[1]
         if self._ps_addrs:
```
brightcoder01: We can extract L358 - L367 into a separate method: _complement_job_command in order to keep the function body short.

workingloong: Good suggestion.

brightcoder01 (approved):

PR #2409, opened 2020-12-09
Add the DistributedOptimizer for TF1.x for elastic training.

brightcoder01:
For the TF1.x model trained using AllReduce strategy with horovod, if we want to train them using elastic way. Use this DistributedOptimizer instead of the one from horovod.

The Optimizer is already verified using the following sample model which are updated based on the sample model from horovod repo.
```Python
import argparse
import errno
import os

import horovod.tensorflow as hvd
import numpy as np
import tensorflow as tf
from tensorflow import keras

from elasticdl.python.allreduce.tensorflow_optimizer import (
    AdjustBackwardPassesPerStepHook,
    DistributedOptimizer,
)

layers = tf.layers

tf.logging.set_verbosity(tf.logging.INFO)

# Training settings
parser = argparse.ArgumentParser(description="Tensorflow MNIST Example")
parser.add_argument(
    "--use-adasum",
    action="store_true",
    default=False,
    help="use adasum algorithm to do reduction",
)
parser.add_argument(
    "--gradient-predivide-factor",
    type=float,
    default=1.0,
    help="apply gradient predivide factor in optimizer (default: 1.0)",
)
args = parser.parse_args()


def conv_model(feature, target, mode):
    """2-layer convolution model."""
    # Convert the target to a one-hot tensor of shape (batch_size, 10) and
    # with a on-value of 1 for each one-hot vector of length 10.
    target = tf.one_hot(tf.cast(target, tf.int32), 10, 1, 0)

    # Reshape feature to 4d tensor with 2nd and 3rd dimensions being
    # image width and height final dimension being the number of color
    # channels.
    feature = tf.reshape(feature, [-1, 28, 28, 1])

    # First conv layer will compute 32 features for each 5x5 patch
    with tf.variable_scope("conv_layer1"):
        h_conv1 = layers.conv2d(
            feature,
            32,
            kernel_size=[5, 5],
            activation=tf.nn.relu,
            padding="SAME",
        )
        h_pool1 = tf.nn.max_pool(
            h_conv1, ksize=[1, 2, 2, 1], strides=[1, 2, 2, 1], padding="SAME"
        )

    # Second conv layer will compute 64 features for each 5x5 patch.
    with tf.variable_scope("conv_layer2"):
        h_conv2 = layers.conv2d(
            h_pool1,
            64,
            kernel_size=[5, 5],
            activation=tf.nn.relu,
            padding="SAME",
        )
        h_pool2 = tf.nn.max_pool(
            h_conv2, ksize=[1, 2, 2, 1], strides=[1, 2, 2, 1], padding="SAME"
        )
        # reshape tensor into a batch of vectors
        h_pool2_flat = tf.reshape(h_pool2, [-1, 7 * 7 * 64])

    # Densely connected layer with 1024 neurons.
    h_fc1 = layers.dropout(
        layers.dense(h_pool2_flat, 1024, activation=tf.nn.relu),
        rate=0.5,
        training=mode == tf.estimator.ModeKeys.TRAIN,
    )

    # Compute logits (1 per class) and compute loss.
    logits = layers.dense(h_fc1, 10, activation=None)
    loss = tf.losses.softmax_cross_entropy(target, logits)

    return tf.argmax(logits, 1), loss


def train_input_generator(x_train, y_train, batch_size=64):
    assert len(x_train) == len(y_train)
    while True:
        p = np.random.permutation(len(x_train))
        x_train, y_train = x_train[p], y_train[p]
        index = 0
        while index <= len(x_train) - batch_size:
            yield x_train[
                index : (index + batch_size)  # noqa: ignore=E203
            ], y_train[
                index : (index + batch_size)  # noqa: ignore=E203
            ],
            index += batch_size


def main(_):
    WORKER_NUM = 3

    # Horovod: initialize Horovod.
    hvd.init()

    # Keras automatically creates a cache directory in ~/.keras/datasets for
    # storing the downloaded MNIST data. This creates a race
    # condition among the workers that share the same filesystem. If the
    # directory already exists by the time this worker gets around to creating
    # it, ignore the resulting exception and continue.
    cache_dir = os.path.join(os.path.expanduser("~"), ".keras", "datasets")
    if not os.path.exists(cache_dir):
        try:
            os.mkdir(cache_dir)
        except OSError as e:
            if e.errno == errno.EEXIST and os.path.isdir(cache_dir):
                pass
            else:
                raise

    # Download and load MNIST dataset.
    (x_train, y_train), (x_test, y_test) = keras.datasets.mnist.load_data(
        "MNIST-data-%d" % hvd.rank()
    )

    # The shape of downloaded data is (-1, 28, 28), hence we need to reshape it
    # into (-1, 784) to feed into our network. Also, need to normalize the
    # features between 0 and 1.
    x_train = np.reshape(x_train, (-1, 784)) / 255.0
    x_test = np.reshape(x_test, (-1, 784)) / 255.0

    # Build model...
    with tf.name_scope("input"):
        image = tf.placeholder(tf.float32, [None, 784], name="image")
        label = tf.placeholder(tf.float32, [None], name="label")
    predict, loss = conv_model(image, label, tf.estimator.ModeKeys.TRAIN)

    lr_scaler = hvd.size()
    # By default, Adasum doesn't need scaling when increasing batch size.
    # If used with NCCL, scale lr by local_size
    if args.use_adasum:
        lr_scaler = hvd.local_size() if hvd.nccl_built() else 1

    # Horovod: adjust learning rate based on lr_scaler.
    opt = tf.train.AdamOptimizer(0.001 * lr_scaler)

    # Use the customized optimizer instead of the DistributedOptimizer
    # from horovod.
    opt = DistributedOptimizer(
        opt,
        op=hvd.Sum,
        backward_passes_per_step=1,
        gradient_predivide_factor=args.gradient_predivide_factor,
        global_batch_count_per_step=WORKER_NUM,
        fixed_global_batch_size=True,
    )

    global_step = tf.train.get_or_create_global_step()
    train_op = opt.minimize(loss, global_step=global_step)

    hooks = [
        # Horovod: BroadcastGlobalVariablesHook broadcasts initial variable
        # states from rank 0 to all other processes. This is necessary to
        # ensure consistent initialization of all workers when training is
        # started with random weights or restored from a checkpoint.
        hvd.BroadcastGlobalVariablesHook(0),
        # Horovod: adjust number of steps based on number of GPUs.
        tf.train.StopAtStepHook(last_step=20000 // hvd.size()),
        tf.train.LoggingTensorHook(
           tensors={"step": global_step, "loss": loss}, every_n_iter=10
        ),
        # Add the hook to update the backward_passes_per_step variable based on
        # the horovod size and the rank of this process.
        AdjustBackwardPassesPerStepHook(
            backward_passes_per_step_var=opt.backward_passes_per_step,
            global_batch_count_per_step=WORKER_NUM,
        ),
    ]

    # Horovod: pin GPU to be used to process local rank (one GPU per process)
    config = tf.ConfigProto()
    config.gpu_options.allow_growth = True
    config.gpu_options.visible_device_list = str(hvd.local_rank())

    # Horovod: save checkpoints only on worker 0 to prevent other workers from
    # corrupting them.
    checkpoint_dir = "./checkpoints" if hvd.rank() == 0 else None
    training_batch_generator = train_input_generator(
        x_train, y_train, batch_size=100
    )
    local_step = 1
    # The MonitoredTrainingSession takes care of session initialization,
    # restoring from a checkpoint, saving to a checkpoint, and closing when
    # done or an error occurs.
    with tf.train.MonitoredTrainingSession(
        checkpoint_dir=checkpoint_dir, hooks=hooks, config=config
    ) as mon_sess:
        while not mon_sess.should_stop():
            # Run a training step synchronously.
            image_, label_ = next(training_batch_generator)
            _, counter_value, backward_passes_per_step_value, global_step_value = mon_sess.run([train_op, opt.counter, opt.backward_passes_per_step, global_step], feed_dict={image: image_, label: label_})
            print("counter_value: {}, backward_passes_per_step_value: {}, global_step_value: {}, local_step: {}".format(counter_value, backward_passes_per_step_value, global_step_value, local_step))
            local_step += 1


if __name__ == "__main__":
    tf.app.run()
```

The sample model definition for elastic training will be submitted in the next PR.

### DIFF
--- a/elasticdl/python/allreduce/tensorflow_optimizer.py
+++ b/elasticdl/python/allreduce/tensorflow_optimizer.py
@@ -18,16 +18,13 @@ import tensorflow as tf
 from horovod.tensorflow import _LegacyOptimizer
 
 
-def complement_value_from_env_if_none(original_value, key, clz):
+def complement_value_from_env_if_none(
+    original_value, key, clz, default_value=None
+):
     if original_value is not None:
         return original_value
 
-    if key in os.environ:
-        return clz(os.environ.get(key))
-
-    raise ValueError(
-        "Cannot complement the value with key {} from environment".format(key)
-    )
+    return clz(os.environ.get(key, default_value))
 
 
 class AdjustBackwardPassesPerStepHook(tf.train.SessionRunHook):
@@ -36,9 +33,7 @@ class AdjustBackwardPassesPerStepHook(tf.train.SessionRunHook):
     the horovod size dynamically.
     """
 
-    def __init__(
-        self, backward_passes_per_step, global_batch_count_per_step=None
-    ):
+    def __init__(self, backward_passes_per_step, hvd_max_size=None):
         """
         Args:
             backward_passes_per_step: A tf.Variable which stands for
@@ -53,8 +48,11 @@ class AdjustBackwardPassesPerStepHook(tf.train.SessionRunHook):
             backward_passes_per_step, self._value_placeholder
         )
 
-        self._global_batch_count_per_step = complement_value_from_env_if_none(
-            global_batch_count_per_step, "global_batch_count_per_step", int
+        hvd_max_size = complement_value_from_env_if_none(
+            hvd_max_size, "WORKER_NUM", int, 1
+        )
+        self._global_batch_count_per_step = (
+            hvd_max_size * backward_passes_per_step
         )
 
     def before_run(self, run_context):
@@ -110,9 +108,11 @@ class LocalGradientAggregationHelper:
     ):
         self._allreduce_grads = allreduce_func
 
-        # backward_passes_per_step controls how often gradient updates are
-        # synchronized.
-        self.backward_passes_per_step = tf.Variable(
+        # backward_passes_per_step is the value set in the model definition,
+        self.backward_passes_per_step = backward_passes_per_step
+        # mutable_local_backward_passes_per_step controls how often gradient
+        # updates are synchronized for this process.
+        self.mutable_local_backward_passes_per_step = tf.Variable(
             initial_value=backward_passes_per_step,
             trainable=False,
             dtype=tf.int32,
@@ -127,8 +127,9 @@ class LocalGradientAggregationHelper:
         self.locally_aggregated_grads = []
 
         # Used to know when to allreduce and apply gradients. We allreduce
-        # when `self.counter` is equal to `self.backward_passes_per_step`. We
-        # apply gradients when `self.counter` is equal to 0.
+        # when `self.counter` is equal to
+        # `self.mutable_local_backward_passes_per_step`. We apply gradients
+        # when `self.counter` is equal to 0.
         self.counter = None
 
         self.sparse_as_dense = sparse_as_dense
@@ -262,31 +263,18 @@ class LocalGradientAggregationHelper:
                 )
 
             with tf.control_dependencies([reset_op]):
-                # *Elastic Update*: If ReduceOP is average,
+                # *ElasticDL Update*: If ReduceOP is average,
                 # multiply horovod_size / global_batch_count_per_step.
                 averaged_gradients = self.update_gradients_for_elastic_workers(
                     averaged_gradients
                 )
 
-                # Divide by backward_passes_per_step if
-                # average_aggregated_gradients is True.
-                with tf.control_dependencies(
-                    [g.op for g in averaged_gradients if g is not None]
-                ):
-                    gradient_divisor = (
-                        self.backward_passes_per_step
-                        if self.average_aggregated_gradients
-                        else 1
-                    )
-
-                    averaged_gradients = apply_op_to_not_none_tensors(
-                        tf.divide, averaged_gradients, gradient_divisor,
-                    )
-                    return averaged_gradients
+                return averaged_gradients
 
     # *ElasticDL Update*: Update the gradient using
     # global_batch_count_per_step if the ReduceOP is average.
     def update_gradients_for_elastic_workers(self, grads):
+        gradient_multiplier = 1
         if self.op == hvd.Average:
             valid_grads = get_not_none_from_list(grads)
             horovod_size = tf.cast(
@@ -296,13 +284,24 @@ class LocalGradientAggregationHelper:
                 dtype=valid_grads[0].dtype,
             )
             gradient_multiplier = (
-                horovod_size / self.global_batch_count_per_step
+                (horovod_size / self.global_batch_count_per_step)
+                if self.average_aggregated_gradients
+                else (
+                    horovod_size
+                    * self.backward_passes_per_step
+                    / self.global_batch_count_per_step
+                )
             )
-            return apply_op_to_not_none_tensors(
-                tf.multiply, grads, gradient_multiplier
+        elif self.op == hvd.Sum:
+            gradient_multiplier = (
+                1.0 / self.backward_passes_per_step
+                if self.average_aggregated_gradients
+                else 1
             )
 
-        return grads
+        return apply_op_to_not_none_tensors(
+            tf.multiply, grads, gradient_multiplier
+        )
 
     def compute_gradients(self, grads):
         """
@@ -336,7 +335,9 @@ class LocalGradientAggregationHelper:
             # equivalent to `backward_passes_per_step`. This the condition is
             # true, it also resets the counter back to 0.
             allreduced_grads = tf.cond(
-                tf.equal(self.counter, self.backward_passes_per_step),
+                tf.equal(
+                    self.counter, self.mutable_local_backward_passes_per_step
+                ),
                 lambda: self._allreduce_grads_helper(grads),
                 lambda: grads,
             )
@@ -524,6 +525,7 @@ def DistributedOptimizer(
     average_aggregated_gradients=False,
     num_groups=0,
     fixed_global_batch_size=False,
+    hvd_max_size=None,
     global_batch_count_per_step=None,
 ):
     """Construct a new DistributedOptimizer, which uses another optimizer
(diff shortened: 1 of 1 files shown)


elasticdl/python/allreduce/tensorflow_optimizer.py
```diff
@@ -0,0 +1,590 @@
...
+    raise ValueError("Cannot complement the value with key {} from environment".format(key))
+
+class AdjustBackwardPassesPerStepHook(tf.train.SessionRunHook):
+    """
+    Hooks that adjusts `backward_passer_per_step` according to
+    the horovod size dynamically.
+    """
+
+    def __init__(
+        self, backward_passes_per_step_var, global_batch_count_per_step=None
```
workingloong: `backward_passes_per_step_var` -> `backward_passes_per_step`?

elasticdl/python/allreduce/tensorflow_optimizer.py
```diff
@@ -0,0 +1,590 @@
...
+        self._value_placeholder = tf.placeholder(
+            backward_passes_per_step_var.dtype, []
+        )
+        self._update_op = tf.assign(
+            backward_passes_per_step_var, self._value_placeholder
+        )
+
+        self._global_batch_count_per_step = complement_value_from_env_if_none(
+            global_batch_count_per_step,
+            "global_batch_count_per_step",
```
workingloong: Is `global_batch_count_per_step` the `num_workers` in args?

elasticdl/python/allreduce/tensorflow_optimizer.py
```diff
@@ -0,0 +1,590 @@
...
+            int)
+
+    def before_run(self, run_context):
+        size = hvd.size()
+        rank = hvd.rank()
+        backward_passer_per_step_value = (
```
workingloong: `backward_passer_per_step_value` -> `backward_passes_per_step_value`

elasticdl/python/allreduce/tensorflow_optimizer.py
```diff
@@ -0,0 +1,590 @@
...
+
+            # *Elastic Update*: If ReduceOP is average,
+            # multiply horovod_size / global_batch_count_per_step
+            with tf.control_dependencies([reset_op]):
+                averaged_gradients = self.update_gradients_for_elastic_workers(averaged_gradients)
+
+                # Divide by backward_passes_per_step if
+                # average_aggregated_gradients is True.
+                with tf.control_dependencies([g.op for g in averaged_gradients if g is not None]):
+                # with tf.control_dependencies([reset_op]):
```
workingloong: Delete the commented code if it is unused.

elasticdl/python/allreduce/tensorflow_optimizer.py
```diff
@@ -0,0 +1,590 @@
...
+        with tf.control_dependencies([aggregation_read_ops]):
+            averaged_gradients = self._allreduce_grads(aggregated_grads)
+
+            # Reset counter.
+            with tf.control_dependencies([g.op for g in averaged_gradients if g is not None]):
+                reset_op = self.counter.assign(
+                    tf.constant(0), use_locking=True)
+
+            # *Elastic Update*: If ReduceOP is average,
+            # multiply horovod_size / global_batch_count_per_step
```
workingloong: It is better to move the comment to line 261 before `update_gradients_for_elastic_workers`

elasticdl/python/allreduce/tensorflow_optimizer.py
brightcoder01: We can rename it because we have already indicate that this parameter should be a tf.Variable.

elasticdl/python/allreduce/tensorflow_optimizer.py
brightcoder01: Good catch, updated.

elasticdl/python/allreduce/tensorflow_optimizer.py
brightcoder01: Good catch, updated.

elasticdl/python/allreduce/tensorflow_optimizer.py
brightcoder01: Not exactly the same.
If backward_passes_per_step is greater than 1 in the model definition, global_batch_count_per_step = num_workers * backward_passes_per_step.

elasticdl/python/allreduce/tensorflow_optimizer.py
```diff
@@ -0,0 +1,671 @@
...
+    device_sparse="",
+    compression=hvd.Compression.none,
+    sparse_as_dense=False,
+    backward_passes_per_step=1,
+    op=hvd.Average,
+    gradient_predivide_factor=1.0,
+    average_aggregated_gradients=False,
+    num_groups=0,
+    fixed_global_batch_size=False,
+    global_batch_count_per_step=None,
```
workingloong: Do users need to set the argument `global_batch_count_per_step` in their codes?  If not, do we need to expose the argument to users?

elasticdl/python/allreduce/tensorflow_optimizer.py
```diff
@@ -0,0 +1,671 @@
...
+                device_dense,
+                device_sparse,
+                compression,
+                backward_passes_per_step,
+            )
+
+        # If global_batch_count_per_step is None from the parameter,
+        # try to get the value from environment variable.
+        global_batch_count_per_step = complement_value_from_env_if_none(
+            global_batch_count_per_step, "global_batch_count_per_step", int
```
workingloong: Now,  we can set "WORKER_NUM" in env of workers. 
https://github.com/sql-machine-learning/elasticdl/blob/d7d081b43ae05525d83d73aa6d430ececfcf5ba5/elasticdl/python/master/pod_manager.py#L140-L142

Can we get the `global_batch_count_per_step` by following codes?
```python
worker_num = int(os.getenv("WORKER_NUM", 0))
global_batch_count_per_step = (
    backward_passes_per_step * worker_num
    if global_batch_count_per_step is None
    else global_batch_count_per_step
)
```

elasticdl/python/allreduce/tensorflow_optimizer.py
```diff
@@ -0,0 +1,671 @@
...
+    def get_slot_names(self, *args, **kwargs):
+        """Calls this same method on the underlying optimizer."""
+        return self._optimizer.get_slot_names(*args, **kwargs)
+
+    def variables(self, *args, **kwargs):
+        """Calls this same method on the underlying optimizer."""
+        return self._optimizer.variables(*args, **kwargs)
+
+
+def DistributedOptimizer(
```
workingloong, line 521: Rename `ElasticDistributedOptimizer`?

elasticdl/python/allreduce/tensorflow_optimizer.py
brightcoder01: Good idea, updated.

elasticdl/python/allreduce/tensorflow_optimizer.py
brightcoder01: Rename the argument to `hvd_max_size`. The value can be set by users or get from the environment if not set in the model definition.

elasticdl/python/allreduce/tensorflow_optimizer.py
brightcoder01: After discussion, we keep the name `DistributedOptimizer` just as the one in horovod package.

elasticdl/python/allreduce/tensorflow_optimizer.py
brightcoder01: Good idea, updated.

workingloong (approved):
LGTM

PR #2410, opened 2020-12-10
Support gpu with type in resource

workingloong:


elasticdl_client/common/k8s_resource.py
```diff
@@ -84,6 +84,9 @@ def parse(resource_str):
                     % k
                 )
             _valid_gpu_spec(v)
+        elif "nvidia.com" in k:
```
brightcoder01: k.startswith("nvidia.com") will be more precise

workingloong: Yeah！

brightcoder01 (approved):

PR #2414, opened 2020-12-10
Elastic AllReduce controller for TensorFlow 1.x

workingloong:


elasticdl/python/tests/allreduce_trainer_test.py
```diff
@@ -163,8 +163,8 @@ def test_elastic_run(self):
         controller.set_broadcast_model(model)
         controller.set_broadcast_optimizer(optimizer)
         controller.broadcast()
-        result = controller.train_one_batch_with_retries(self.train)
         controller.restore()
+        result = controller.train_one_batch_with_retries(self.train)
```
brightcoder01: Why do we move `train_one_batch_with_retries` after `restore`?

elasticdl/python/allreduce/tensorflow_controller.py
```diff
@@ -74,3 +75,44 @@ def train_one_batch_with_retries(self, func, *args, **kwargs):
...
+                return result
+            except UnknownError as e:
+                logger.warning(
+                    "Failed to perform allreduce operation on "
+                    "the gradients. Retrying..."
+                )
+                # Those error message show that the communication
+                # to merge gradient fails and we can rebuild the
+                # communication.
+                if (
```
brightcoder01, line 112: How about the `HorovodUpdateError` and `HorovodInternalError`?

elasticdl/python/allreduce/tensorflow_controller.py
```diff
@@ -74,3 +75,44 @@ def train_one_batch_with_retries(self, func, *args, **kwargs):
                     self._rendezvous_manager.init_horovod_if_needed()
         self.data_shard_service.report_batch_done()
         return result
+
+
+class TensorFlowV1AllReduceController(AllReduceController):
+    """The controller is responsible for elastic training of
+    TensorFlow eager execution using AllReduce.
```
brightcoder01: eager -> graph

elasticdl/python/tests/allreduce_trainer_test.py
workingloong: This is a mistake in the PR and I will remove the modification.

elasticdl/python/allreduce/tensorflow_controller.py
workingloong: `HorovodUpdateError` and `HorovodInternalError` are the errors in Horovod not TensorFlow. Horovod will convert the `UnknownError` in TensorFlow to those errors.
https://github.com/horovod/horovod/blob/e94d8ea0dff8c2b45698cfe4fabb2e6553d0b9a8/horovod/tensorflow/elastic.py#L53-L61

Here, we get the error from TensorFlow, not Horovod. So we can not use `HorovodUpdateError` and `HorovodInternalError`.

brightcoder01 (approved):

PR #2418, opened 2020-12-14
Set ENV PYTHONUNBUFFERE=0 in the dockerfile

workingloong:
Now, the log message will not be dumped in the log of pod if users use `print` in their codes.

### DIFF
--- a/elasticdl/docker/Dockerfile
+++ b/elasticdl/docker/Dockerfile
@@ -4,6 +4,10 @@ ARG BASE_IMAGE
 FROM ${BASE_IMAGE} as dev
 ARG EXTRA_PYPI_INDEX=https://pypi.org/simple
 
+# Allows for log messages by `print` in Python to be immediately dumped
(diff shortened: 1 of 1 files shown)


elasticdl/docker/Dockerfile
```diff
@@ -4,6 +4,10 @@ ARG BASE_IMAGE
 FROM ${BASE_IMAGE} as dev
 ARG EXTRA_PYPI_INDEX=https://pypi.org/simple
 
+# Allows for log messages by `print` in Python to be immediately dumped
+# to the stream instead of being buffered.
+ENV PYTHONUNBUFFERED 0
```
brightcoder01, line 9: What's the side effect if the message is buffered? Will we drop any message?

workingloong: We cannot view the log message of `print` by `kubectl logs` and we will miss the log message if the pod is deleted.

brightcoder01 (approved):

PR #2420, opened 2020-12-15
Master does not exit if there are no worker

workingloong:


elasticdl/python/master/pod_manager.py
```diff
@@ -503,6 +503,9 @@ def _event_cb(self, event):
 
     @property
     def all_workers_exited(self):
+        # At start, there maybe not workers.
```
brightcoder01: there may be no worker

elasticdl/python/master/pod_manager.py
```diff
@@ -503,6 +503,9 @@ def _event_cb(self, event):
 
     @property
     def all_workers_exited(self):
+        # At start, there may be no worker.
+        if not self._pod_info_cache[PodType.WORKER]:
```
brightcoder01: We can move these code snippets inside the lock.

skydoorkai (dismissed):

skydoorkai (approved):

Issue #2417, opened 2020-12-14
ElasticDL Package Refactor


brightcoder01:
# Goal
- The package is light-weight in the model development stage.
- The package can be composable with various deep learning frameworks including TF1.x / TF2.x / PyTorch.
- Optional: user can execute/debug the model code locally in the Macbook / PC.

# Development Stage
## Requirement
- Users develop the model using TF1.x / TF2.x / PyTorch.
- Users develop the model with horovod (AllReduce) or without horovod (PS).
- Users don't need install `kubernetes` package.

## Solutions
Release a separate wheel named `elasticdl_sdk`, it contains the required minimal modules for user model development.

Directory Structure
```TXT
-- elasticdl_sdk
---- tensorflow
---- pytorch
---- common
```
It contains the following:
Controller, DistributedOptimizer  => Separate implementations under `tensorflow/pytorch`
master_client, DataShardService (other DL frameworks for PS based training), RendevousManager => `common`

# Execution Stage
## Requirements
- Execute TF1.x / TF2.x / PyTorch model. (TF1.x and TF2.x cannot be installed in the container at the same time)
- Execute using AllReduce or PS strategy. => We can always install `horovod` in the container even though using the PS strategy.
- Current status: worker is tightly coupled with TF2.x.

## Solutions
- AllReduce(TF1.x + PyTorch): we don't need worker.py play the role of launcher. The training loop and data feed process both are defined in the model file, only need elasticdl_sdk
- TF1.x + PS: the same with above
- TF2.x: need worker.py play the role of launcher, need the complete set of elasticdl package. 

# Action Items
- Release a new package `elasticdl_sdk` (P0)
- Refactor master (P1), separate it into two types of components, and user can install either one or both: the core components which take the role of Pod management and task management. the components  aware with the model definition (requires tensorflow or pytorch)
- Refactor worker (P3): composable with various deep learning framework (TF1.x/TF2.x/PyTorch)

Question:
- For the scenario of AllReduce and TF1.x PS, the docker image of worker is different with the one of master. The worker image only need elasticdl_sdk and TF/PyTorch that user specifies. The master image need the complete set of elasticdl package and TF2.0 (because these two are tightly coupled).

# Reference
1. [horovod](https://github.com/horovod/horovod) can be composable with various deep learning frameworks.
Take `HOROVOD_WITH_TENSORFLOW=1 pip install horovod[tensorflow]` for example.
[...]

PR #2422, opened 2020-12-15
Initial version of elasticai_api package.

brightcoder01:
- `setup.py` and add it in the build script.
- Move `DistributedOptimizer` from elasticdl main package to elasticdl_sdk package.

Relate with issue #2417


setup_sdk.py
```diff
@@ -0,0 +1,36 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from setuptools import find_packages, setup
+
+tensorflow_require_list = ["tensorflow"]
+pytorch_require_list = ["torch"]
+
+setup(
+    name="elasticdl_sdk",
```
workingloong: We should rename it.

brightcoder01: After discussion, we will name it `elasticai_api`. The wheel name is `elasticai-api`.

setup_api.py
```diff
@@ -0,0 +1,36 @@
...
+setup(
+    name="elasticai-api",
+    version="0.2.0rc3.dev0",
+    description="The model development api for ElasticDL.",
+    long_description="This is the sdk for developing ElasticDL models."
+    " Model developers can make slight update using the api in it based on"
+    " their TensorFlow or Pytorch models. And then the model can be trained"
+    " in the elastic and fault-tolerant way.",
+    long_description_content_type="text/markdown",
+    author="Ant Financial",
```
skydoorkai: Ant Group?

setup_api.py
```diff
@@ -0,0 +1,36 @@
...
+
+tensorflow_require_list = ["tensorflow"]
+pytorch_require_list = ["torch"]
+
+setup(
+    name="elasticai-api",
+    version="0.2.0rc3.dev0",
+    description="The model development api for ElasticDL.",
+    long_description="This is the sdk for developing ElasticDL models."
+    " Model developers can make slight update using the api in it based on"
```
skydoorkai: Model developers can use these APIs to support elastic and fault-tolerant training for their TensorFlow or Pytorch models.

setup_api.py
brightcoder01: Good catch, updated.

setup_api.py
brightcoder01: Good idea, updated.

skydoorkai (approved):

PR #2424, opened 2020-12-15
Use the max completed time of task to check timeout tasks.

workingloong:
If the completed time of the task is very small, the timeout will happen frequently.

### DIFF
--- a/elasticdl/python/master/task_manager.py
+++ b/elasticdl/python/master/task_manager.py
@@ -14,7 +14,6 @@
 """TaskQueue Implementation"""
 
 import random
-import statistics
 import threading
 import time
 
@@ -497,7 +496,7 @@ class TaskManager(object):
         while True:
             doing_tasks = self._doing.copy()
             cur_time = time.time()
-            avg_time = self._get_average_task_completed_time()
+            max_time = self._get_max_task_completed_time()
             for _, (worker_id, task, start_time) in doing_tasks.items():
                 if task.type == elasticdl_pb2.TRAINING:
                     start_time = self._worker_start_task_time[worker_id]
@@ -507,7 +506,7 @@ class TaskManager(object):
                 ]:
                     task_elapsed_time = cur_time - start_time
                     if task_elapsed_time > max(
-                        _TASK_TIMEOUT_THRESHOLD, 3 * avg_time[task.type]
+                        _TASK_TIMEOUT_THRESHOLD, 3 * max_time[task.type]
                     ):
                         logger.info(
                             "worker %d timeout, relaunch it" % worker_id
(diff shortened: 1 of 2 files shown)


brightcoder01 (dismissed):

elasticdl/python/master/task_manager.py
```diff
@@ -514,7 +517,7 @@ def _check_and_reassign_timeout_tasks(self):
                         break
             time.sleep(30)
 
-    def _get_average_task_completed_time(self):
+    def _get_max_task_completed_time(self):
```
skydoorkai: Change the comments below from average to max.

elasticdl/python/master/task_manager.py
```diff
@@ -523,23 +526,23 @@ def _get_average_task_completed_time(self):
...
         if len(self._task_completed_times[elasticdl_pb2.TRAINING]) < 20:
-            average_task_completed_time[elasticdl_pb2.TRAINING] = 300
-        else:
-            average_task_completed_time[
+            max_task_completed_time[
                 elasticdl_pb2.TRAINING
-            ] = statistics.mean(
+            ] = _TASK_TIMEOUT_THRESHOLD
+        else:
+            max_task_completed_time[elasticdl_pb2.TRAINING] = max(
```
brightcoder01: self._task_completed_times[elasticdl_pb2.TRAINING] contains the completed time for all the Training tasks. The array will be large if there is a lot of data in the source table and the max function will be slower and slower. If we only care the max time, only update the max value for each completed tasks instead of recording the time of each task.

elasticdl/python/tests/task_manager_test.py
```diff
@@ -154,7 +154,7 @@ def test_check_and_reassign_timeout_tasks(self):
     def test_get_average_task_completed_time(self):
         task_manager = create_task_manager([("f1", 0, 10), ("f2", 0, 10)], [])
         average_task_completed_time = (
```
brightcoder01: average_task_completed_time -> max_task_completed_time.

brightcoder01 (dismissed):

elasticdl/python/master/task_manager.py
```diff
@@ -30,6 +29,7 @@
 from elasticdl.python.data.reader.data_reader_factory import create_data_reader
 
 _MAX_TASK_RETRIES = 3
+_TASK_TIMEOUT_THRESHOLD = 300
```
brightcoder01: _TASK_TIMEOUT_THRESHOLD -> _TASK_TIMEOUT_THRESHOLD_MS, we can tell the reader that the unit of the time is milliseconds.

workingloong: Good suggestion.

brightcoder01 (approved):

PR #2434, opened 2020-12-17
Add an argument to disable the thread to check timeout tasks.

workingloong:


elasticdl_client/common/args.py
```diff
@@ -177,6 +177,13 @@ def add_train_params(parser):
         help="If true, task manager supports fault tolerance, otherwise "
         "no fault tolerance.",
     )
+    add_bool_param(
+        parser=parser,
+        name="--task_timeout_tolerance",
```
brightcoder01: I prefer to name it `recycle_timeout_task`

skydoorkai: It does more than recover tasks. It also kills the corresponding worker pod.

`--worker_timeout_detection` ?

workingloong: `relaunch_timeout_worker` ?

brightcoder01 (approved):

PR #2443, opened 2020-12-21
Move grpc_utils.py and log_utils.py to util folder in elasticai_api.

brightcoder01:


workingloong:
Can elasticai_api and elasticdl have their `log_util` and `grpc_util` separately? It can decrease dependency of elasticdl.

brightcoder01:
> Can elasticai_api and elasticdl have their `log_util` and `grpc_util` separately? It can decrease dependency of elasticdl.

The dependency on `elasticai_api` is necessary for `elasticdl` because the proto messages and services about Dynamic Sharding is required by these two packages. I would prefer keeping only one copy of the common modules for these two packages in `elasticai_api` in order to avoid duplicated code.

workingloong (approved):

PR #2444, opened 2020-12-21
Rearrange worker pod priority

skydoorkai:


elasticdl/python/master/pod_manager.py
```diff
@@ -188,10 +201,11 @@ def __init__(
         **kwargs
     ):
         self._num_workers = num_workers
+        self._chief_worker_index = chief_worker_index
         self._worker_resource_request = worker_resource_request
         self._worker_resource_limit = worker_resource_limit
         self._worker_pod_priority = _parse_worker_pod_priority(
-            self._num_workers, worker_pod_priority
+            self._num_workers, worker_pod_priority, self._chief_worker_index
```
workingloong: Can we set the worker-0 as the chief worker? Because the worker has high priority is the `worker_pod_priority` is not `low`. Then, we don't need any code to set the chief worker. What's more, the chief worker is a role in TensorFlow PS and we only set the configuration of the chief worker into env for the worker-0.

workingloong:
It is better to create a PR from your forked repo, not the original repo.

elasticdl/python/master/pod_manager.py
skydoorkai: For estimator, the last worker is the chief worker for evaluation.

elasticdl_client/common/args.py
```diff
@@ -189,6 +189,12 @@ def add_train_params(parser):
         help="The command executed in the pod launched by the master",
         default="",
     )
+    parser.add_argument(
```
brightcoder01: Will we public this arguments to the user?

skydoorkai: No need. Deleted.

workingloong (approved):

PR #2446, opened 2020-12-22
Raise a runtime error if fail to perform allreduce operation

workingloong:


elasticai_api/common/base_controller.py
```diff
@@ -37,6 +37,7 @@
 DEFAULT_SECS_TO_CHECK_RENDEZVOUS = min(
     60, int(os.getenv(HorovodEnv.GLOO_TIMEOUT_SECONDS, 30))
 )
+RETRY_ALLREDUCE_INTERVALU_SECS = 30
```
skydoorkai: INTERVALU -> INTERVAL?

elasticai_api/pytorch/controller.py
```diff
@@ -131,11 +134,13 @@ def train_one_batch_with_retries(self, func, *args, **kwargs):
             except RuntimeError:
                 traceback.print_exc()
                 self.restore()
+        if not allreduce_success:
+            raise RuntimeError("Failed to perform allreduce.")
```
skydoorkai, line 138: So the worker pod will fail after this runtime error?

workingloong: Yes. The worker will fail if it fails to retry to perform the allreduce operation.

brightcoder01 (approved):

skydoorkai (approved):

PR #2447, opened 2020-12-22
The master only exit according to the status of workers

workingloong:


elasticdl/python/master/master.py
```diff
@@ -103,18 +103,14 @@ def run(self):
...
             while True:
-                if self.task_manager and self.task_manager.finished():
-                    if self.pod_manager:
-                        self.pod_manager.update_status(
-                            PodManagerStatus.FINISHED
-                        )
-                    break
                 if self.pod_manager and self.pod_manager.all_workers_exited:
-                    if self.task_manager:
+                    if self.task_manager and not self.task_manager.finished():
```
skydoorkai, line 107: If workers early stop, there will be tasks left. Thus, warning only.

workingloong: Done

skydoorkai (approved):

PR #2448, opened 2020-12-22
The worker sends the training parameters to the master.

workingloong:


elasticai_api/common/master_client.py
```diff
@@ -90,3 +90,14 @@ def get_comm_rank(self):
         req = elasticai_api_pb2.GetCommRankRequest()
         req.worker_id = self._worker_id
         return self._stub.get_comm_rank(req)
+
+    def report_training_params(
+        self, batch_size, num_epochs=None, dataset_size=None
```
skydoorkai, line 107: Why batch_size is required, others are optional?

skydoorkai: records_per_task also can be an optional argument.

workingloong: `batch_size` is required because the worker needs it to report the status of tasks. I don't think `records_per_task` should be in the arguments because it is difficult for users to understand `records_per_task`. Now, we use the default value of `num_minibatches_per_task` 50 to create shards.

skydoorkai (dismissed):

skydoorkai (dismissed):

skydoorkai (approved):

PR #2451, opened 2020-12-23
The worker sends the start and end message to the master.

### LINKED ISSUE #2445: The worker sends start and end message to the master for AllReduce.
Now, users can customize the training loop. Sometimes, users may do some preparation work before starting the training loop. For example, users may need to transform the total images and load the model from a checkpoint. After finishing the training loop, users may need to do post-training work. Those preparation and post-training work may have an influence on the training using AllReduce of other workers.

The master will add the living worker into the ring if the worker starts. However, the worker will not start the training loop immediately for the preparation work. Then, other workers in the ring must wait for the worker which may raise a timeout. 

The master will remove the living worker out of the ring if the worker exists. The worker will not exits if there is post-training work. Other workers may also raise a timeout if there are remaining shards to train.

We can develop a `start` and `end` in the `AllReduceController`. The controller should send a start message to the master before the training loop and send an end message to the master.

```python
class AllReduceController(object):
    def __enter__(self):
        send_start()
    def __exit__(self, exc_type, exc_val, exc_tb):
         send_end()
```

Then, users need to execute the training loop in `with allreduce_controller`:

```python

allreduce_controller = create_elastic_controller(
    batch_size=args.batch_size
)

with allreduce_controller:
    for batch_idx, (data, target) in enumerate(data_loader):
        target = target.type(torch.LongTensor)
    
        # Use the elastic function to wrap the training function with a batch.
        elastic_train_one_batch = allreduce_controller.elastic_run(
            train_one_batch
        )
        loss = elastic_train_one_batch(model, optimizer, data, target)
```

workingloong:
Fix #2445

### DIFF
--- a/elasticai_api/proto/elasticai_api.proto
+++ b/elasticai_api/proto/elasticai_api.proto
@@ -78,7 +78,7 @@ message GetCommRankResponse {
   int32 rendezvous_port = 4;
 }
 
-message ReportTrainingLoopStatusRequest{
+message ReportTrainingLoopStatusRequest {
   string worker_host = 1;
   int32 status = 2;
 }
--- a/elasticai_api/proto/elasticai_api.proto
+++ b/elasticai_api/proto/elasticai_api.proto
@@ -68,7 +68,7 @@ message ReportTaskResultRequest {
 }
 
 message GetCommRankRequest {
-  int32 worker_id = 1;
+  string worker_host = 1;
 }
 
 message GetCommRankResponse {
@@ -79,7 +79,7 @@ message GetCommRankResponse {
 }
 
 message ReportTrainingLoopStatusRequest{
-  int32 worker_id = 1;
+  string worker_host = 1;
   int32 status = 2;
 }
 
--- a/elasticai_api/proto/elasticai_api.proto
+++ b/elasticai_api/proto/elasticai_api.proto
@@ -78,9 +78,16 @@ message GetCommRankResponse {
   int32 rendezvous_port = 4;
 }
 
+message ReportTrainingLoopStatusRequest{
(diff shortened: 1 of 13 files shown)


elasticai_api/proto/elasticai_api.proto
```diff
@@ -78,9 +78,16 @@ message GetCommRankResponse {
   int32 rendezvous_port = 4;
 }
 
+message ReportTrainingLoopStatusRequest {
+  string worker_host = 1;
```
brightcoder01, line 88: What does worker_host mean? IP, service domain name or pod name?

elasticdl/python/master/servicer.py
```diff
@@ -166,12 +166,18 @@ def report_version(self, request, _):
...
         res.rank_id = self._rendezvous_server.get_worker_host_rank(worker_host)
         res.world_size = self._rendezvous_server.get_size()
         res.rendezvous_id = self._rendezvous_server.get_rendezvous_id()
         res.rendezvous_port = self._rendezvous_server.get_rendezvous_port()
         return res
+
+    def report_training_loop_status(self, request, _):
+        training_loop_status = request.status
+        if training_loop_status == TrainingLoopStatus.START:
+            self._rendezvous_server.add_worker(request.worker_host)
```
brightcoder01, line 189: Add the check 
```Python
if self._rendezvous_server:
    self._rendezvous_server.add_worker(request.worker_host)
```

elasticai_api/proto/elasticai_api.proto
workingloong: IP

skydoorkai (approved):

PR #2453, opened 2020-12-23
Add a lock to get task

workingloong:


elasticai_api/common/data_shard_service.py
```diff
@@ -34,12 +34,13 @@ def get_current_task(self):
...
-        task = self._mc.get_task(task_type)
-        if task.type == elasticai_api_pb2.TRAINING:
-            self._pending_tasks.append(task)
-            if len(self._pending_tasks) == 1:
-                self._current_task = task
-        return task
+        with self._lock:
+            task = self._mc.get_task(task_type)
+            if task.type == elasticai_api_pb2.TRAINING:
+                self._pending_tasks.append(task)
```
skydoorkai, line 40: The lock can be added before `self._pending_tasks.append(task)`

workingloong: Good suggestion and fixed it.

skydoorkai (approved):

PR #2460, opened 2020-12-25
Add more parameters in the factory method of data_shard_service.

brightcoder01:


workingloong:
Why do we need the argument?

skydoorkai (approved):

PR #2462, opened 2020-12-28
create worker service and set TF_CONFIG env when needed

skydoorkai:
1. Add need_tf_config arg. 
2. Create worker service
3. Set TF_CONFIG for ps/worker using the service domain name.
4. Patch worker service when the corresponding worker pod changes.

### DIFF
--- a/elasticdl/python/master/pod_manager.py
+++ b/elasticdl/python/master/pod_manager.py
@@ -293,9 +293,7 @@ class PodManager(object):
             original_index = self._worker_pod_priority_and_original_index[
                 worker_id
             ][1]
-            tf_config = self._k8s_client.get_tf_config_data(
-                self._num_workers, self._num_ps, PodType.WORKER, original_index
-            )
+            tf_config = self.get_tf_config_data(PodType.WORKER, original_index)
             envs.append(
                 V1EnvVar(name="TF_CONFIG", value=json.dumps(tf_config))
             )
--- a/elasticdl/python/master/pod_manager.py
+++ b/elasticdl/python/master/pod_manager.py
@@ -185,7 +185,6 @@ class PodManager(object):
         image_pull_policy=None,
         restart_policy="Never",
         envs=None,
-        need_tf_config=False,
         disable_relaunch=False,
         log_file_path=None,
         **kwargs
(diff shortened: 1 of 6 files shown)


elasticdl/python/master/pod_manager.py
```diff
@@ -162,6 +162,7 @@ def create_pod_manager(args):
             cluster_spec=cluster_spec,
             cluster_spec_json=args.cluster_spec_json,
             envs=env,
+            need_tf_config_env=args.need_tf_config_env,
```
workingloong: `args.need_tf_config_env` -> `args.need_tf_config` ?

elasticdl/python/master/pod_manager.py
```diff
@@ -184,14 +186,15 @@ def __init__(
...
+        need_tf_config=False,
         disable_relaunch=False,
         log_file_path=None,
         **kwargs
     ):
         self._num_workers = num_workers
         self._worker_resource_request = worker_resource_request
         self._worker_resource_limit = worker_resource_limit
-        self._worker_pod_priority = _parse_worker_pod_priority(
+        self._worker_pod_priority_and_ori_index = _parse_worker_pod_priority(
```
brightcoder01: _worker_pod_priority_and_ori_index => _worker_pod_priority_and_original_index

elasticdl/python/master/pod_manager.py
```diff
@@ -279,12 +283,29 @@ def _start_worker(self, worker_id):
         worker_args = [self._worker_args[0], job_command]
         envs = copy.deepcopy(self._envs)
         envs.append(V1EnvVar(name=WorkerEnv.WORKER_ID, value=str(worker_id)))
+        need_create_service = False
+        need_patch_service = False
+        ori_index = worker_id
```
brightcoder01: ori_index -> original_index

elasticdl/python/master/pod_manager.py
```diff
@@ -279,12 +286,29 @@ def _start_worker(self, worker_id):
...
```
brightcoder01: I think we can maintain a "Resource" to Pod Map
The Resource contains (priority, service_name ...) and is just like a slot / placeholder. And then we will assign this resource to each created pod. If one pod fails, then the resource is assigned to None and open to a incoming new created pod.
In this way, we can avoid the original_id/new_id logic and easy to maintain.

What's more, we can create the service at first and then always patch the service to the new launched worker no matter whether we are creating the worker at the start time or relaunch a new one for a failed worker.

brightcoder01 (approved):

PR #2463, opened 2020-12-28
Read indices from a shard 

### LINKED ISSUE #2459: Fail to fetch shard using multi-process in Python.
```python
class ImageDataset(IterableDataset):
    def __init__(self, data_shard_service, shuffle=False):
        self.data_shard_service = data_shard_service
        self._shuffle = shuffle

    def __iter__(self):
        while True:
            index = self.data_shard_service.fetch_shard()
            image, label = read_images(index)
            yield image, label

dataset = ImageDataset(
      allreduce_controller.data_shard_service, shuffle=True
  )

data_loader = DataLoader(
      dataset=dataset,
      batch_size=args.batch_size,
      num_workers=2
  )
```

image

workingloong:
Fix #2459

### DIFF
--- a/elasticai_api/common/data_shard_service.py
+++ b/elasticai_api/common/data_shard_service.py
@@ -11,7 +11,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import queue
 import random
 import threading
 import time
@@ -21,7 +20,6 @@ from multiprocessing import SimpleQueue
 from elasticai_api.common.constants import TaskExecCounterKey
 from elasticai_api.common.master_client import build_master_client
 from elasticai_api.proto import elasticai_api_pb2
-from elasticai_api.util.log_utils import default_logger as logger
 
 
 def build_data_shard_service(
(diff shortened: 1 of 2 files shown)


elasticai_api/common/data_shard_service.py
```diff
@@ -125,12 +130,45 @@ def report_batch_done(self, batch_size=None, err_msg=""):
             return True
         return False
 
-    def fetch_shard(self):
```
brightcoder01: TF 1.x custom training loop relies on this method. We can keep this method and also provide the implementation of multiple process.

workingloong: Yes, it is a mistake and I have restored it.

elasticai_api/common/data_shard_service.py
```diff
@@ -134,3 +137,51 @@ def fetch_shard(self):
...
+        shuffle=False,
+    ):
+        super(RecordIndexService, self).__init__(
+            master_client, batch_size, num_epochs, dataset_size, task_type
+        )
+        self._shuffle = shuffle
+        self._shard_queue = SimpleQueue()
+        threading.Thread(
+            target=self._get_shard_indices,
+            name="check_timeout_tasks",
```
skydoorkai: name not related?

elasticai_api/common/data_shard_service.py
```diff
@@ -134,3 +137,51 @@ def fetch_shard(self):
...
+            daemon=True,
+        ).start()
+
+    def _get_shard_indices(self):
+        while True:
+            if self._shard_queue.empty():
```
skydoorkai, line 165: Don't need to wait for empty, if the queue size is small, such as   `< batchsize`, get next shard.

```
 if self._shard_queue.qsize() < batchsize:

skydoorkai (approved):

PR #2465, opened 2020-12-31
Add elasticdl job arguments only when need_elasticdl_job_args=True

skydoorkai:


elasticdl/python/master/pod_manager.py
```diff
@@ -336,7 +339,7 @@ def _complement_job_command(self):
         # the second string is the shell command to run, like
         # ["-c", "python -m elasticdl.python.worker.main --minibatch_size 64"]
         job_command = self._worker_args[1]
-        if self._ps_addrs:
+        if self._ps_addrs and self._need_elasticdl_job_args:
```
brightcoder01, line 342: Will there be bug if we keep --ps_addrs arguments for custom_training_loop?

skydoorkai: For custom_training_loop, elasticdl worker pods are used. Thus there will be no bug as elasticdl worker can process this arg.

Also, custom_training_loop is used for _elastic_allreduce_train only for now. there will be no ps, so no ps_addrs.

brightcoder01 (approved):

PR #2468, opened 2021-01-05
Master raise a runtime error if all workers failed

workingloong:


elasticdl/python/master/pod_manager.py
```diff
@@ -596,6 +596,19 @@ def all_workers_exited(self):
 
         return all_exited
 
+    def all_workers_failed(self):
+        with self._lock:
+            all_failed = all(
```
brightcoder01, line 602: We need check `_not_created_worker_id` is empty here. The worker creation may fail and will be retried in the next periodical call.

workingloong: What does the master do If `_not_created_worker_id` is not empty and other workers failed?

brightcoder01 (dismissed):

brightcoder01 (approved):

PR #2469, opened 2021-01-06
Support shuffling the total dataset.

workingloong:


elasticai_api/proto/elasticai_api.proto
```diff
@@ -24,6 +24,7 @@ message Shard {
   // Starting and ending (non-inclusive) record number.
   int64 start = 2;
   int64 end = 3;
+  repeated int64 indices = 4;
```
brightcoder01, line 27: I think we can make the shard size small and then shuffle the shard list instead of shuffling each id and returning the id list.

workingloong: We usually store the images in the folder with the class name. For example:
```
|--mnist_png
    |--0
        |--001.png
        |--002.png
        |...
    |--1
       |--001.png
       |--002.png
```

Then, we can use `torchvision.datasets.ImageFolder("./mnist_png")` to load paths and classes of the images. 
image

If we create shards according to the `dataset.imgs`, the shard may only contains samples of a class. So we must shuffle all indices to create shards.

elasticai_api/common/data_shard_service.py
```diff
@@ -45,12 +44,14 @@ def __init__(
...
         num_epochs=None,
         dataset_size=None,
+        shuffle=False,
         task_type=elasticai_api_pb2.TRAINING,
     ):
         self._mc = master_client
         self._batch_size = batch_size
         self._num_epochs = num_epochs
         self._dataset_size = dataset_size
+        self._shuffle = shuffle
```
brightcoder01, line 54: I think we can name it element_wise_shuffle or per_sample_shuffle.
Because we will shuffle the shard list for training by default.

elasticdl/python/master/task_manager.py
```diff
@@ -293,6 +308,11 @@ def create_tasks(self, task_type, model_version=-1):
                     start_idx_this_task + self._records_per_task,
                     max_idx_this_shard,
                 )
+                task_record_indices = task_record_indices = (
```
brightcoder01: `task_record_indices = task_record_indices = xxx`.
Is it a typo?

workingloong: Yes, fix it!

elasticai_api/common/data_shard_service.py
workingloong: `shuffle` is a common word in PyTorch and TensorFlow. Here, I want to keep the same as these frameworks and not introduce a new word to users. I think we should use the argument `shuffle_shards` to shuffle the shard list for training.

brightcoder01: `shuffle_shards` is a good idea.

workingloong: Yeah, and we can add the argument `shuffle_shards` in the next PR.

brightcoder01 (approved):

PR #2473, opened 2021-01-08
Add num_minibatches_per_shard param in report_training_params

brightcoder01:
For the scenario of custom training loop, the training parameters are reported from the workers such as batch_size, num_epochs and so on. `num_minibatches_per_shard` cannot be configured and always use the default value `8`. In this PR, we add this parameters in data_shard_service and make it configurable to the model developer.

### DIFF
--- a/elasticai_api/common/data_shard_service.py
+++ b/elasticai_api/common/data_shard_service.py
@@ -76,7 +76,7 @@ class DataShardService(object):
                 dataset_size=self._dataset_size,
                 shuffle=self._shuffle,
                 shuffle_shards=self._shuffle_shards,
-                num_minibatches_per_shard=self._num_minibatches_per_shard
+                num_minibatches_per_shard=self._num_minibatches_per_shard,
             )
 
     def get_current_task(self):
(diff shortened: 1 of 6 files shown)


workingloong (dismissed):
LGTM

elasticai_api/common/data_shard_service.py
```diff
@@ -51,6 +51,7 @@ def __init__(
         shuffle=False,
         shuffle_shards=False,
         task_type=elasticai_api_pb2.TRAINING,
+        num_minibatches_per_shard=0,
```
workingloong, line 54: Is it better to set a positive integer for the default value?

brightcoder01: The default value 0 means that we will use the value set by master.

elasticdl/python/master/task_manager.py
```diff
@@ -222,43 +222,66 @@ def _set_completed_steps_by_checkpoint(self, checkpoint_dir_for_init):
...
-                )
-                self._dataset_size = (
-                    dataset_size if dataset_size > 0 else self._dataset_size
-                )
-                self._training_shards = self._create_shards_by_dataset_size(
-                    dataset_size
+            if self._training_shards:
+                logger.info(
+                    "The training shards have already been initialized."
+                    "Igore these training parameters."
```
skydoorkai: Igore -> Ignore

brightcoder01: Good catch, fixed.

skydoorkai (approved):

workingloong (approved):

PR #2497, opened 2021-02-02
Fix the counter if the worker retry to aggregate gradient in ElasticDL

workingloong:


elasticai_api/tensorflow/optimizer.py
```diff
@@ -331,14 +331,16 @@ def compute_gradients(self, grads):
             assert len(grads) == len(self.locally_aggregated_grads)
 
             # Allreduce locally aggregated gradients when the counter is
-            # equivalent to `backward_passes_per_step`. This the condition is
-            # true, it also resets the counter back to 0.
+            # not less than `backward_passes_per_step`. Because, the counter
```
skydoorkai: when the counter equals or exceeds `backward_passes_per_step`. The counter may exceed `backward_passes_per_step` because of retries in the fault-tolerant allreduce.

skydoorkai (approved):

PR #2499, opened 2021-02-02
Remove unnecessary Worker param

zuston:
when running test, it will throw warning log. like following. 

`[2021-02-02 18:02:49,547] [WARNING] [args.py:239:parse_worker_args] Unknown arguments: ['--worker_id', '0']`

Actually, worker.py don't use worker_id. it can be removed.

### DIFF
--- a/elasticdl/python/tests/worker_ps_interaction_test.py
+++ b/elasticdl/python/tests/worker_ps_interaction_test.py
@@ -156,8 +156,6 @@ class WorkerPSInteractionTest(unittest.TestCase):
         model_def = "mnist.mnist_functional_api.custom_model"
         self._create_pserver(model_def, 2)
         arguments = [
-            "--worker_id",
-            0,
             "--job_type",
             elasticai_api_pb2.TRAINING,
             "--minibatch_size",
--- a/elasticdl/python/tests/worker_ps_interaction_test.py
+++ b/elasticdl/python/tests/worker_ps_interaction_test.py
@@ -71,8 +71,6 @@ class WorkerPSInteractionTest(unittest.TestCase):
             tf.keras.backend.clear_session()
             tf.random.set_seed(22)
             arguments = [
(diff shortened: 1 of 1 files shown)


workingloong (dismissed):

zuston:
The PS written by Python has been deprecated, right? @workingloong

workingloong:
> The PS written by Python has been deprecated, right? @workingloong

Yes, we have developed the PS using Go. The performance of go-PS is much better than Python. Now, we only use the Python-PS to run unit tests of the worker.

zuston:
Thanks reply. @workingloong 
After reading design doc and example, I found that current **elasticdl** code is deeply bound to k8s. In fact, whether it is k8s or Yarn, the design of elasticdl is reusable. There are some plans to support different schedulers ?

workingloong:
> Thanks reply. @workingloong
> After reading design doc and example, I found that current **elasticdl** code is deeply bound to k8s. In fact, whether it is k8s or Yarn, the design of elasticdl is reusable. There are some plans to support different schedulers ?

No, there is no plan to support other schedulers.  ElasticDL needs to call the API of schedulers to monitor the worker status and relaunch the worker if it is preempted.  And k8s is becoming more and more popular in Machine Learning.

zuston:
> > Thanks reply. @workingloong
> 
> > After reading design doc and example, I found that current **elasticdl** code is deeply bound to k8s. In fact, whether it is k8s or Yarn, the design of elasticdl is reusable. There are some plans to support different schedulers ?
> 
> 
> 
> No, there is no plan to support other schedulers.  ElasticDL needs to call the API of schedulers to monitor the worker status and relaunch the worker if it is preempted.  And k8s is becoming more and more popular in Machine Learning.

Pity. I have to fork it to support Yarn. Thanks your great work on it~

workingloong (approved):
Perfect！

PR #2501, opened 2021-02-05
Remove the code to report the training data

workingloong:


elasticdl/python/master/task_manager.py
```diff
@@ -274,19 +270,9 @@ def set_training_params(
...
-            ):
-                # The training_data is a directory only using RecordIO.
-                self._create_training_tasks(training_data, None)
-            elif dataset_size > 0:
-                self._training_shards = self._create_shards_by_dataset_size(
-                    dataset_size
-                )
-            else:
-                logger.warning("No data to create shards")
+            self._training_shards = self._create_shards_by_dataset_size(
```
skydoorkai: Still need to check `dataset_size>0`, or we may get wrong training_shards.

skydoorkai (approved):

PR #2502, opened 2021-02-17
Develop a reader to read records from recordio files

workingloong:


elasticai_api/io/recordio_reader.py
```diff
@@ -0,0 +1,61 @@
...
+                    else:
+                        break
+
+    def _get_record_file(self, start, end):
+        target_files = []
+        for block in self._data_blocks:
+            if start < block.end:
+                if end < block.end:
+                    target_files.append(
+                        (block.name, start - block.start, end - start)
```
skydoorkai, line 58: what if  `start < block.start`?
Also if `end <  block.start`?

workingloong: The range of blocks is in increasing order. For example, blocks are [[0,100),[100, 200),[200,300)].  So, we can check which block the shard is in by the order. So, `start > blocks[i].start` if `start > blocks[i-1].end` and `start < blocks[i].end` .

workingloong: I have added a docstring to explain it.

skydoorkai (approved):

PR #2508, opened 2021-02-22
Master will not add any worker if the current rendezvous hosts become empty after starting training.

workingloong:


elasticai_api/common/base_controller.py
```diff
@@ -63,6 +63,8 @@ def init_horovod_if_needed(self):
                 time.sleep(RETRY_ALLREDUCE_INTERVAL_SECS)
             else:
                 break
+        if rank_response.rank_id < 0:
+            raise ValueError("Cannot get an invalid rank")
```
skydoorkai: It can get an invalid rank.  -> valid.

skydoorkai (approved):

PR #2518, opened 2021-04-14
Add an example of Tensorflow 1.x with pictures.

workingloong:


model_zoo/mnist/mnist_tfv1_img.py
```diff
@@ -0,0 +1,220 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+Download the mnist dataset from
+https://s3.amazonaws.com/fast-ai-imageclas/mnist_png.tgz
+and then untar it into ${data_store_dir}. On minikube, we can use the
```
brightcoder01, line 17: Extract the mnist dataset into raw png files?
How about adding the full command about the steps.

model_zoo/mnist/mnist_tfv1_img.py
```diff
@@ -0,0 +1,220 @@
...
+
+def arg_parser():
+    parser = argparse.ArgumentParser(description="Process training parameters")
+    parser.add_argument("--batch_size", type=int, default=64, required=False)
+    parser.add_argument("--num_epochs", type=int, default=1, required=False)
+    parser.add_argument(
+        "--learning_rate", type=float, default=0.1, required=False
+    )
+    parser.add_argument(
+        "--no-cuda",
```
brightcoder01: It seems that the argument is not used in the train loop.

model_zoo/mnist/mnist_tfv1_img.py
workingloong: The command to untar the mnist_png.tgz ?

brightcoder01: We can keep the current statement.

brightcoder01 (approved):

PR #2530, opened 2021-09-09
A PyTorch tutorial

### LINKED ISSUE #2514: Write a tutorial of elastic AllReduce with PyTorch

workingloong:
Fix #2514


Kelang-Tian:
Does the 'dataset' here refer to some data sets obtained by the worker through shards index instead of all training data sets? Because I think the traing dataset size is fixed.

> L47 in mnist_pytorch.py
    def __len__(self):
        if self.data_shard_service:
            # Set the maxsize because the size of dataset is not fixed
            # when using dynamic sharding
            return sys.maxsize
        else:
            return len(self._images)

Kelang-Tian:
Does each worker store a complete dataset locally， and the worker selects the required data from the local dataset through the dataset shard index which is get from data_shard_service? 
@workingloong 

https://github.com/workingloong/elasticdl/blob/2eab01011171dbe1eab3cfce1daa9327f875dac5/model_zoo/mnist/mnist_pytorch.py#L107

model_zoo/mnist/mnist_pytorch.py
```diff
@@ -131,15 +107,14 @@ def train(args):
     train_data = torchvision.datasets.ImageFolder(args.training_data)
```
Kelang-Tian, line 107: Does each worker store a complete dataset locally， and the worker selects the required data from the local dataset through the dataset shard index which is get from data_shard_service?

model_zoo/mnist/mnist_pytorch.py
```diff
@@ -131,15 +107,14 @@ def train(args):
...
         batch_size=args.batch_size,
         dataset_size=len(train_data.imgs),
         num_epochs=args.num_epochs,
         shuffle=True,
+        dataset_name="mnist_training_data",
     )
-    train_dataset = ElasticDataset(
-        train_data.imgs, allreduce_controller.data_shard_service
-    )
+    train_dataset = ElasticDataset(train_data.imgs, data_shard_service)
```
Kelang-Tian, line 117: Does the 'dataset' here refer to some data sets obtained by the worker through shards index instead of all training data sets? Because I think the traing dataset size is fixed.

L49:   # Set the maxsize because the size of dataset is not fixed
            # when using dynamic sharding

model_zoo/mnist/mnist_pytorch.py
workingloong: No, each worker can only need to access the dataset storage. For example, we can store the dataset in the remote distribution storage like glusterfs, NAS or HDFS which each worker can access. In the example, I store the dataset on each worker because the dataset is small.

model_zoo/mnist/mnist_pytorch.py
workingloong: Yes, the dataset reads images by sample indices in the shard.  The reason of `sys.maxsize` is that PyTorch `dataloader` will get the size N of the dataset and then only read N samples. After reading N samples, the data loader thinks an epoch is completed and will start a new epoch. If epochs are done, the data loader will finish. However, in elastic training, the epoch processing is across all workers, so the sample count of a worker is not fixed.

Kelang-Tian (approved):
LGTM

